PR #6, opened 2015-08-16
Breaking: CommonMark-compatible code fences (fixes #2)

### LINKED ISSUE #2: Does not unindent two-space indents
If this tool is going to be generalized for other projects to use for linting code blocks, it should account for two-space indents as well as tabs and four-space indents.  Currently, this test fails:

```js
it("should unindent 2 space-indented code fences", function() {
    var code = [
        "  ```js",
        "  var answer = 6 * 7;",
        "  ```"
    ].join("\n");
    var blocks = processor.preprocess(code);

    assert.equal(blocks[0], "var answer = 6 * 7;\n");
});
```

This seems due to the regex in: https://github.com/eslint/eslint-plugin-markdown/blob/45f03b57cb1cf41660c1c9c4ab714d41f77f3de8/lib/processor.js#L92

btmills:
http://spec.commonmark.org/0.21/#fenced-code-blocks

- Anything indented >4 spaces is considered an indented code block
- Code fences may be indented by 0 to 3 spaces
- Lines are unindented up to the level of the opening fence
- Fences may have backticks (`) or tildes (~)
- Fences may consist of >= 3 characters
- Closing fences must be at least as long as the opening fence
- Opening fence info strings may contain everything except backticks
- Whitespace around opening fence info strings is ignored
- Closing fences may be followed by whitespace but nothing else

@IanVS how's this?

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -49,11 +49,58 @@ function parse(text) {
     }
 
     /**
+     * Tries to match an opening code fence on the current line.
+     * @returns {object|boolean} Fence data or false if no match.
+     */
+    function matchOpeningCodeFence() {
+        var captures;
+
+        // Exit early if an opening fence is not expected.
+        if (mode !== modes.NORMAL) {
+            return false;
+        }
+
+        captures = match(/^( {0,3})(`{3,}|~{3,})\h*(?:js|javascript)\h*\r?\n/i);
+        if (captures) {
+            return {
+                indent: captures[1],
+                fence: captures[2]
+            };
+        }
+
+        return false;
+    }
+
+    /**
+     * Tries to match a closing code fence on the current line.
+     * @returns {string[]|boolean} Match result or false if no match.
+     */
+    function matchClosingCodeFence() {
+        var block, regexp;
+
+        // Exit early if a closing fence is not expected.
+        if (mode !== modes.CODE) {
+            return false;
+        }
+
+        block = result[result.length - 1];
+        regexp = [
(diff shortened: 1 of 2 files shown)


IanVS:
Looks awesome.  One question about indented code blocks, though.  You listed:

> Anything indented >4 spaces is considered an indented code block

According to http://spec.commonmark.org/0.21/#indented-code-blocks [emphasis mine]:

> An indented code block is composed of one or more indented chunks separated by blank lines. An indented chunk is a sequence of non-blank lines, each indented **four or more** spaces.

But, it seems you're not starting code blocks based on indentation, and in fact you're testing that you're not: https://github.com/eslint/eslint-plugin-markdown/commit/731829b2dbab875e4c17355f35250e80b8a0e906#diff-3afb72c6840bab858402afbe15e6315fL40.  Or am I missing something?

btmills:
Good catch, that's a typo. Should be: Anything indented **>=** 4 spaces is considered an indented code block. So a line beginning with >= 4 spaces followed by 3 backticks would be 3 backticks inside a code block, not the start of a code fence. And since the plugin only cares about code fences marked "js" or "javascript", it can skip checking any line beginning with >= 4 spaces for the start of a code fence. Hopefully the description matches the code and the spec now.

IanVS:
OK, makes sense.

btmills:
#5 was merged, and this is now rebased on top of it. :ship:

IanVS:
:clap:

PR #18, opened 2015-09-17
Update: Optimize expected error checking (fixes #16)

### LINKED ISSUE #16: Optimize expected error checking
After merging #14, it occurred to me that there may be a couple optimizations possible. I'll look into it and see what I come up with.

btmills:
I realized we already had the text of the current line, so there was no need to regenerate it with the `split()`. Now the error checking only runs when a line of code is found, and it only checks mode once.

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -30,9 +30,7 @@ function parse(text) {
         column = 0,
         mode = modes.NORMAL,
         next,
-        expErr,
-        expErrQty,
-        currentLineText;
+        err;
 
     var EXP_ERR_BLOCK_REGEX = /\/\*\s*error\s+(.+?)\s*\*\//g,
         EXP_ERR_LINE_REGEX = /\/\/\s*error\s+(.+)/g,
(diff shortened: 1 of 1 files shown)


btmills:
@IanVS any input?

lib/processor.js
```diff
@@ -170,23 +166,26 @@ function parse(text) {
 
         // Line content
         } else {
-            next = match(/[^\r\n]*/);
+            next = match(/^[^\r\n]*/);
```
IanVS: Will this only grab one line now? Before it was returning everything to end of the file.

If it is different from the `next` as used above, maybe it deserves a different name? `nextLine` maybe?

btmills: Adding the anchor makes no difference to the logic of the program. It was already only reading to the end of the line (until it saw `\r` or `\n`). Combined with the earlier check for `/^/r?/n/`, this will only be hit when the line does not begin with a newline. That also means it will always match. Adding the anchor would be an optimization if it ever didn't match, but now it shouldn't change anything. It does, however, make it consistent with the other (non-error) regexes, which have anchors.

IanVS: Oh, ok.  It was `text` that was the whole text up to the end of file.  Not sure why I didn't think to use `next`.

IanVS:
LGTM

PR #22, opened 2015-09-20
Update: Support jsx code fences (fixes #19)

### LINKED ISSUE #19: Allow `jsx` info strings
The code blocks in [jsx-quotes](http://eslint.org/docs/rules/jsx-quotes) would be better expressed as `jsx` rather than `js`

`js`:
```js
<a b='c' />
```

`jsx`:
```jsx
<a b='c' />
```

But this would require supporting `jsx` info strings in this plugin.  Ideally, would the pre-processor also be able to add the `jsx` ecmaFeature? Otherwise, linting of these will need to wait until https://github.com/eslint/eslint/issues/3698.

btmills:


IanVS:
I was planning to change the info strings to `jsx` in the `jsx-quotes` rule docs, which for now would work because they won't be linted.  But if this lands, linting will break, until there's a way to set `ecmaFeatures` from within a file / code block.  Maybe we can hold off on this one for a bit?

btmills:
:+1: Ping back here when you're ready for this.

btmills:
@IanVS where are we on this?

IanVS:
It was decided not to lint the rule docs, so feel free to merge if this is useful.

PR #24, opened 2015-10-12
Breaking: Remove optional error suppression (fixes #23)

### LINKED ISSUE #23: Remove "expected error" undocumented feature.
After the conversation starting with https://github.com/eslint/eslint-plugin-markdown/pull/13#issuecomment-141810700, it seems best to remove the functionality of ignoring error messages, as it is not generally useful to consuming projects and could cause needless confusion. 

This would essentially mean reverting #5, #13, #14 and #18.

btmills:
/cc @IanVS


IanVS:
LGTM

PR #36, opened 2015-12-22
Update: Add ESLint integration tests (fixes #33)

### LINKED ISSUE #33: Add ESLint integration tests
These should use the [CLIEngine API](http://eslint.org/docs/developer-guide/nodejs-api#cliengine) to exercise the plugin.

btmills:


btmills:
@ilyavolodin would you mind taking a glance at [tests/plugin.js](https://github.com/eslint/eslint-plugin-markdown/pull/36/files#diff-d35f7840503664a126d94bc389c9c52c) in this PR and letting me know if that's the right way to test the plugin's integration with ESLint?

ilyavolodin:
@btmills That looks correct to me. If you are asking about `executeOnText` vs. `executeOnFile` then it doesn't matter, since both of them call into `processText` which is a function responsible for loading processors. :+1:

btmills:
@ilyavolodin :cool: thanks!

PR #39, opened 2016-01-16
Update: Use remark for parsing (fixes #38)

### LINKED ISSUE #38: Consider using Remark for parsing
Remark is a dedicated library for parsing Markdown files to an AST. It would probably be better to use a dedicated parser instead of rolling out our own.

btmills:
/cc @BYK

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -9,179 +9,51 @@
 "use strict";
 
 var assign = require("object-assign");
+var remark = require("remark");
+
+var SUPPORTED_SYNTAXES = ["js", "javascript", "node", "jsx"];
 
 var blocks = [];
 
 /**
- * Parses Markdown text for fenced code blocks containing JS.
- * @param {string} text Markdown text.
- * @returns {Block[]} List of parsed code blocks.
+ * Performs a depth-first traversal of the Markdown AST.
+ * @param {ASTNode} node A Markdown AST node.
+ * @param {object} callbacks A map of node types to callbacks.
+ * @returns {void}
  */
-function parse(text) {
-    var modes = {
-        NORMAL: "NORMAL",
-        CODE: "CODE"
-    };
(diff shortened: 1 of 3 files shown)


lib/processor.js
```diff
@@ -9,179 +9,51 @@
...
-        }
-
-        captures = match(/^( {0,3})(`{3,}|~{3,})\h*(?:js|javascript|node|jsx)\h*\r?\n/i);
-        if (captures) {
-            return {
-                indent: captures[1],
-                fence: captures[2]
-            };
-        }
+function traverse(node, callbacks) {
```
BYK: Hah, may be we should start a project called `marktraverse` for this stuff :)

lib/processor.js
```diff
@@ -9,179 +9,51 @@
...
  */
 function preprocess(text) {
-    blocks = parse(text);
+    var ast = remark.parse(text);
+
+    blocks = [];
+    traverse(ast, {
+        "code": function(node) {
+            if (node.lang && SUPPORTED_SYNTAXES.indexOf(node.lang.toLowerCase()) >= 0) {
+                blocks.push(node);
```
BYK: Why not `blocks.push(node.value)` directly if you're not using any other property of the node afterwards?

BYK:
Nice. I think the build failure is not directly related.

lib/processor.js
btmills: I thought about shaving that yak just for fun, but it was late, and I didn't want to write all the tests I'd have to write for a new module :)

btmills:
> I think the build failure is not directly related.

Yeah, that had to do with `eslint-config-eslint` and `eslint@2.0.0-beta.1` as a `peerDependency`. There's another PR that attempted to solve that, but I'll have to try something else today.

btmills:
Build failures were solved in #41.

PR #41, opened 2016-01-16
Upgrade: Dependencies (fixes #40)

### LINKED ISSUE #40: Upgrade eslint-config-eslint
Need the changes from eslint/eslint#4910.

btmills:
This will pull in the changes to `eslint-config-eslint` from eslint/eslint#4910, which requires `eslint@2`. Updated `object-assign` as well to clean up `npm outdated` output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,16 +26,13 @@
     "test": "eslint --ext .js --ext .md . && mocha tests"
   },
   "main": "index.js",
-  "peerDependencies": {
-    "eslint": ">=0.16.0"
-  },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


btmills:
Looks like `npm@2` doesn't like the peer dependency on `eslint@2.0.0-beta.1`. Barring some creative solution, I'm going to hold off until ESLint 2 is released.

btmills:
Turns out the solution was as simple as dropping the peer dependency. eslint-plugin-react doesn't specify a peer dependency and specifies the 2.0 beta in `devDependencies`, which is exactly what this does.

PR #49, opened 2016-04-15
Fix: Exclude eol-last from output (fixes #48)

### LINKED ISSUE #48: Impossible to satisfy `eol-last` rule
With the upgrade to `remark`, it is not impossible to enforce `eol-last` because all trailing whitespace on code blocks are removed.

May be a good idea to silence this completely internally if possible. Or at least say something about this on docs.

btmills:
/cc @BYK for review

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -12,6 +12,9 @@ var assign = require("object-assign");
 var remark = require("remark");
 
 var SUPPORTED_SYNTAXES = ["js", "javascript", "node", "jsx"];
+var UNSATISFIABLE_RULES = [
+    "eol-last" // The Markdown parser strips trailing newlines in code fences
+];
 
 var blocks = [];
 
@@ -58,6 +61,15 @@ function preprocess(text) {
 }
 
 /**
+ * Excludes unsatisfiable rules from the list of messages.
+ * @param {Message} message A message from the linter.
+ * @returns {boolean} True if the message should be included in output.
+ */
+function excludeUnsatisfiableRules(message) {
+    return UNSATISFIABLE_RULES.indexOf(message.ruleId) < 0;
+}
+
+/**
  * Transforms generated messages for output.
  * @param {Array<Message[]>} messages An array containing one array of messages
  *     for each code block returned from `preprocess`.
@@ -65,7 +77,7 @@ function preprocess(text) {
  */
 function postprocess(messages) {
     return [].concat.apply([], messages.map(function(group, i) {
-        return group.map(function(message) {
+        return group.filter(excludeUnsatisfiableRules).map(function(message) {
             return assign({}, message, {
                 line: message.line + blocks[i].position.start.line,
                 column: message.column + blocks[i].position.indent[message.line - 1] - 1
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ Blocks that don't specify either `js`, `javascript`, `jsx`, or `node` syntax are
     print("This doesn't get linted either.")
     ```
 
+## Unsatisfiable Rules
+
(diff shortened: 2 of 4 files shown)


lib/processor.js
```diff
@@ -12,6 +12,9 @@ var assign = require("object-assign");
 var remark = require("remark");
 
 var SUPPORTED_SYNTAXES = ["js", "javascript", "node", "jsx"];
+var UNSATISFIABLE_RULES = [
```
BYK: May wanna make this configurable at a future version.

lib/processor.js
```diff
@@ -58,14 +61,23 @@ function preprocess(text) {
 }
 
 /**
+ * Excludes unsatisfiable rules from the list of messages.
+ * @param {Message} message A message from the linter.
+ * @returns {boolean} True if the message should be included in output.
+ */
+function excludeUnsatisfiableRules(message) {
+    return UNSATISFIABLE_RULES.indexOf(message.ruleId) < 0;
```
BYK: If you are okay with dropping support for Node < 4 , we can use a `Set` instead of an array and have quicker (and simpler) lookups.

BYK:
LGTM

Only concern is a bout tests. It's not clear to me from the PR that this new feature is properly covered by tests.

lib/processor.js
btmills: That's a good point. I don't believe we have a way to support plugin-specific configuration. Is that correct, @ilyavolodin? If other core rules comes up, we can add them to this list. Hopefully by the time we get requests to exclude rules in third-party plugins, we have glob configs that will allow users to exclude those rules on their own.

lib/processor.js
btmills: This is definitely something we can do as soon as core drops support for Node < 4.

btmills:
@byk Previously, all of the plugin tests would blow up if the exclusion wasn't working because I enabled `eol-last` in the integration tests. However, you make a good point that it wasn't clear that's what was happening, so I added an additional postprocess test to make it explicit what's supposed to happen. What do you think now?

lib/processor.js
ilyavolodin: As in allow rules inside plugins to specify if they should apply to markdown or not? Or configure all of the rules inside plugin in a specific way? If former - then no.

btmills: A way to apply config to a postprocess step in a plugin. Something like:

```json
{
    "pluginConfig": {
        "markdown": {
            "unsatisfiableRules": ["eol-last", "react/no-multi-comp"]
        }
    }
}
```

I checked on the plugin documentation and didn't see anything like that.

ilyavolodin: Hmm... You could use `settings`. Although I'm having trouble finding it in the documentation for some reason.
Ah, found it (yay, search!), it's strange that it's not mentioned in working with plugins, but it is mentioned in the configuring eslint: http://eslint.org/docs/user-guide/configuring#adding-shared-settings

btmills: `settings` only gets passed to rules though. If necessary, we could:

1. Modify core to pass `settings` to processors.
2. Hack it in by offering a `markdown/unsatisfiable-rules` meta-rule that would remember its config.
3. Shave an entire herd of yacks and get glob config in core.

ilyavolodin: I vote for shaving yacks! You get nice sweaters that way too!

BYK: Let's do both? :D

lib/processor.js
BYK: Do we wanna follow the core conventions for the plugin? My feeling is plugins should not be bound by what core supports or not.

BYK:
@btmills - looks great, prefrect!

lib/processor.js
btmills: Since this is under the ESLint organization, I think it's in our best interest to maintain the same compatibility targets unless there's a compelling reason not to.

PR #55, opened 2016-10-24
Update: Support configuration comments (fixes #46, fixes #54)

### LINKED ISSUE #46: New: ability to only lint for certain parts of the code block
Ref https://github.com/airbnb/javascript/issues/685

Not sure how this would work but an example would be in airbnb's style guide.

> 3.1 Use the literal syntax for object creation.
> eslint rules: no-new-object.

```js
// bad
const item = new Object();

// good
const item = {};

```

So in this case you would want to check the `// good` part is linted correctly while the `// bad` snippet is incorrect.

### LINKED ISSUE #54: Support configuring rules in comments
@ljharb [requested](https://github.com/eslint/eslint-plugin-markdown/issues/46#issuecomment-231518548) the ability to configure rules inside comments.

Once we figure out the syntax for disabling ESLint on specific blocks via comments in #46, we can look at figuring out how to do this.

btmills:
Initially, #46 was about adding custom syntax and logic into the processor to support configuration comments. The plugin would permit disabling ESLint or specific rules for entire code blocks, and all of that would be done in the plugin. Eventually, #54 would support configuring rule options.

I realized while working on #46 that custom syntax is unnecessary and overly complex when ESLint already has built-in support for configuration comments! With this change, when the plugin sees an HTML comment preceding a code block, it converts the comment's body into a JS block comment at the beginning of the source code. If it's an ESLint config comment, ESLint will pick up on it.

This example enables the browser environment, disables the no-alert rule, and configures the quotes rule to prefer single quotes:

    
    
    

    ```js
    alert('Hello, world!');
    ```

Internally, the generated code that gets passed to ESLint looks like this:

```js
/* eslint-env browser */
/* eslint-disable no-alert */
/* eslint quotes: ["error", "single"] */
alert('Hello, world!');
```

### DIFF
--- a/tests/fixtures/long.md
+++ b/tests/fixtures/long.md
@@ -17,16 +17,33 @@ function foo() {
 }
 ```
 
+<!-- eslint-env node -->
+<!-- eslint-disable eol-last, quotes -->
+
 ```node
 console.log(process.version);
 ```
 
 How about some JSX?
 
+<!--
+    eslint quotes: [
+        "error",
+        "single"
+    ]
+-->
+<!--eslint-disable no-console-->
+
 ```jsx
 console.log("Error!");
 ```
 
     I may be a code block, but don't lint me!
+
+<!-- eslint-disable -->
+
+```js
+!@#$%^&*()
+```
 
 The end.
--- a/README.md
+++ b/README.md
@@ -71,6 +71,39 @@ Blocks that don't specify either `js`, `javascript`, `jsx`, or `node` syntax are
     print("This doesn't get linted either.")
     ```
 
+## Configuration Comments
+
+The processor will convert HTML comments immediately preceding a code block into JavaScript block comments and insert them at the beginning of the source code that it passes to ESLint. This permits configuring ESLint via configuration comments while keeping the configuration comments themselves hidden when the markdown is rendered. Comment bodies are passed through unmodified, so the plugin supports any [configuration comments](http://eslint.org/docs/user-guide/configuring) supported by ESLint itself.
+
+This example enables the `browser` environment, disables the `no-alert` rule, and configures the `quotes` rule to prefer single quotes:
+
+    <!-- eslint-env browser -->
+    <!-- eslint-disable no-alert -->
+    <!-- eslint quotes: ["error", "single"] -->
+
+    ```js
+    alert('Hello, world!');
+    ```
+
+Each code block in a file is linted separately, so configuration comments apply only to the code block that immediately follows.
+
+    Assuming `no-alert` is enabled in `.eslintrc`, the first code block will have no error from `no-alert`:
+
+    <!-- eslint-env browser -->
+    <!-- eslint-disable no-alert -->
+
+    ```js
+    alert("Hello, world!");
+    ```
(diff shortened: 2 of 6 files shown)


wooorm:
Woah cool! 👍

tests/fixtures/long.md
```diff
@@ -17,16 +17,33 @@ function foo() {
...
+-->
+<!--eslint-disable no-console-->
+
 ```jsx
 console.log("Error!");
 ```
 
     I may be a code block, but don't lint me!
 
+<!-- eslint-disable -->
```
ljharb: what about `eslint-disable-line` and `eslint-disable-next-line`?

btmills: Since ESLint only supports those as line comments, those wouldn't have any effect inserted at the top of the code block. There's nothing preventing using them inside though. Is there a need to be able to disable individual lines from outside the code block?

ljharb: i guess if `eslint-disable` only affects the next code block - as opposed to the entire file it appears in (like in JS files), then there's no need.

btmills: Correct. I should document that.

    $ cat pull55.md
    
    
    ```js
    alert('Hello, world!')
    ```
    
    ```js
    alert('Hello, world!')
    ```
    $ node_modules/.bin/eslint pull55.md
    
    /Users/brandon/code/eslint/eslint-plugin-markdown/pull55.md
      8:1   error  Unexpected alert              no-alert
      8:1   error  'alert' is not defined        no-undef
      8:7   error  Strings must use doublequote  quotes
      8:23  error  Missing semicolon             semi
    
    ✖ 4 problems (4 errors, 0 warnings)

btmills: Just added a bit about that to the docs.

btmills:
@ljharb thanks for your patience! I'm not giving up on asserting expected failures, but I do want to get this out.

If I do a release with this and as #56 as `v1.0.0-beta.3`, can you test it out, make sure it works the way you need it to, then let me know and I'll release `v1.0.0` final?

ljharb:
Sure!

btmills:
@ljharb I just published `eslint-plugin-markdown@v1.0.0-beta.3` - let me know how it works!

ljharb:
@btmills thanks, looking good so far!

(side note, if you could help me figure out how to configure this in sublime, that'd be awesome)

ljharb:
@btmills it does kind of seem like some `<!-- eslint-disable foo -->` directives might be disabling all linting of that code block.

btmills:
@ljharb I opened #57 for editor integrations and posted some of my research thus far.

Do you have an example I can look at where it's disabling too much?

ljharb:
@btmills i haven't gotten the tests to pass yet (because linking circular deps in a monorepo setup is hard) but if you check out https://github.com/airbnb/javascript/compare/master...ljharb:lint_markdown locally and manually set up the expected dependencies, and run `npm run lint` in the root of the repo, you should see the linter passing - but there's still a number of un-disabled rule errors in some of the blocks that have comments on them.

PR #62, opened 2017-04-26
Update remark

wooorm:
Hi folks! 👋

I’ve updated remark! This does come with a change in HTML comments, so I had to use a loop around the `getComments` functionality.

I also removed `remark` directly. Rather, I added `unified` and `remark-parse`, which is a bit smaller, meaning faster installs.

Let me know if I should fix-up some code-style!

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -6,7 +6,6 @@
 "use strict";
 
 var assign = require("object-assign");
-var parse5 = require("parse5");
 var unified = require("unified");
 var remarkParse = require("remark-parse");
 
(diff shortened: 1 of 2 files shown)


btmills:
@wooorm thanks for this, and thanks for your work on `remark`!

Previously `getComments` would parse the `html` AST node that was passed to it. Now it looks like `remark` appears to split comments into distinct nodes - is that correct? If so, can I remove the `parse5` step and just strip the `<!--` and `-->` from the node's `value`?

wooorm:
That’s correct, the following markdown:

```markdown
<!--some eslint comment-->
<!--one more eslint comment-->
~~~js
code();
~~~
```

...is turned into:

```json
{
  "type": "root",
  "children": [
    {
      "type": "html",
      "value": "<!--some eslint comment-->"
    },
    {
      "type": "html",
      "value": "<!--one more eslint comment-->"
    },
    {
      "type": "code",
      "lang": "js",
      "value": "code();"
    }
  ]
}
```

I’ll update the PR somewhere today!

wooorm:
Done! Let me know if there’s some code-style I need to fix!

platinumazure:
Do we want to support the line comments as well? (e.g., `// eslint-disable-next-line someRule`)

Probably a better question for @btmills, anyway.

wooorm:
Awesome! 👍

PR #63, opened 2017-04-29
Build: Use eslint-release

btmills:
Adapted [Makefile.js](https://github.com/eslint/typescript-eslint-parser/blob/65c2e0ae7b04c67d9bca17fb5a883cbe8a0deeaa/Makefile.js) from [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser) and added [eslint-release](https://github.com/eslint/eslint-release) for automated releases.


Makefile.js
```diff
@@ -0,0 +1,106 @@
...
+"use strict";
+
+/* eslint no-console: 0*/
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+require("shelljs/make");
+
+var nodeCLI = require("shelljs-nodecli");
```
ilyavolodin: Do you need nodecli? `shelljs/make` already exposes `exec` command as is.

Makefile.js
```diff
@@ -0,0 +1,106 @@
...
+// Helpers
+//------------------------------------------------------------------------------
+
+/**
+ * Generates a function that matches files with a particular extension.
+ * @param {string} extension The file extension (i.e. "js")
+ * @returns {Function} The function to pass into a filter method.
+ * @private
+ */
+function fileType(extension) {
```
ilyavolodin: I know you just copied it from ESLint makefile, but you can use `ls('lib/**/*.js')` instead (I'm pretty sure you can, at least)

not-an-aardvark:
What's the advantage of moving test/lint scripts to the Makefile rather than keeping them as npm scripts? At first glance it seems like unnecessary abstraction, but there is probably a reason that I'm missing.

Makefile.js
btmills: Looks like I was able to remove it!

Makefile.js
btmills: Indeed, that did work :+1:

btmills:
> @not-an-aardvark What's the advantage of moving test/lint scripts to the Makefile rather than keeping them as npm scripts?

I'm going to go with "consistency with other eslint org projects", but I'm not attached to that argument. If the team would prefer npm scripts, I'm happy to do that too.

not-an-aardvark:
My two cents is that using npm scripts might be better because additional flags can be provided, e.g. `npm run lint -- --fix` will pass the `--fix` flag to eslint. I don't feel strongly about this though, so I'm fine with whatever.

btmills:
@not-an-aardvark just pushed up a commit to use npm scripts instead

PR #64, opened 2017-04-29
Chore: Increase code coverage

btmills:
🚧  ~~This depends on #63, which should be merged first, and then I'll update the base branch for this PR back to `master`.~~ 🚧  *Edit: Done*

### DIFF
--- a/tests/lib/processor.js
+++ b/tests/lib/processor.js
@@ -355,6 +355,40 @@ describe("processor", function() {
             ].join("\n"));
         });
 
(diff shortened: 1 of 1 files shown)


wooorm:
Whoops, thanks for fixing!

btmills:
@wooorm these tests were missing before you made any changes, so you're completely off the hook 😅  I had started working on #63, which added code coverage reports, last weekend, which is when I discovered these missing cases that were in the code all along.

wooorm:
Ah, good! 😌

PR #67, opened 2017-04-30
Build: Explicitly specify package.json files

btmills:
`eslint-release` expects `pkg.files` to be an array of relative paths without any globs.

`lib/.eslintrc.yml` shouldn't be included in the bundle, so it's necessary to specify individual paths instead of just `lib`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,9 @@
   },
   "main": "index.js",
   "files": [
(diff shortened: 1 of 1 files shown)


btmills:
Agreed, working on an issue for that now.

PR #88, opened 2018-03-31
Fix: Emit correct endLine numbers

### LINKED ISSUE #87: Does not emit correct endLine numbers
When the `message.line` numbers are updated by `processor`, the message's `endLine` number is not updated (if it exists).

This breaks highlighting in IDE plugins such as `vscode-eslint`:

screen shot 2018-03-31 at 6 30 43 pm

(Edit: new screenshot)

plmrry:
Fixes #87

This adds each blocks `start.line` to each message's `endLine`, if it exists on the message.

~~This PR also updates the `eslint` used in tests to version 4. The older version of eslint did not emit `endLine` numbers.~~

### DIFF
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -60,10 +60,33 @@ describe("plugin", function() {
         var report = cli.executeOnText(code, "test.md");
         assert.equal(report.results[0].messages[0].message, "'baz' is not defined.");
         assert.equal(report.results[0].messages[0].line, 5);
-        assert.equal(report.results[0].messages[0].endLine, 5);
+        assert.equal(report.results[0].messages[0].endLine, 4);
         assert.equal(report.results[0].messages[1].message, "'blah' is not defined.");
         assert.equal(report.results[0].messages[1].line, 8);
-        assert.equal(report.results[0].messages[1].endLine, 8);
+        assert.equal(report.results[0].messages[1].endLine, 7);
+    });
+
+    it("should emit correct line numbers with leading comments", function() {
+        var code = [
+            "# Hello, world!",
+            "",
+            "<!-- eslint-disable quotes -->",
+            "",
+            "```js",
+            "var bar = baz",
+            "",
+            "var str = 'single quotes'",
+            "",
+            "var foo = blah",
+            "```"
+        ].join("\n");
+        var report = cli.executeOnText(code, "test.md");
+        assert.equal(report.results[0].messages[0].message, "'baz' is not defined.");
+        assert.equal(report.results[0].messages[0].line, 6);
+        assert.equal(report.results[0].messages[0].endLine, 6);
+        assert.equal(report.results[0].messages[1].message, "'blah' is not defined.");
+        assert.equal(report.results[0].messages[1].line, 10);
+        assert.equal(report.results[0].messages[1].endLine, 10);
     });
 
     it("should run on .mkdn files", function() {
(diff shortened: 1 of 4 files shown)


plmrry:
Hmm, further testing shows something is weird with the way comment blocks are being used in the line number calculations...

plmrry:
Fixed! I wasn't actually adding the `leadingCommentLines` in my calcuation

plmrry:
It seems Node < 4.0 doesn't like eslint 4.0 😕

To get the build passing, I added a test to `processor.js` and skipped my tests in `plugin.js` (which require eslint 4.0 to pass, since before eslint 4.0, `endLine` was not a part of the `message` object)

Aladdin-ADD:
yes, eslint v4 does not support Node < 4. and the up-coming eslint v5 will drop supporting Node < 6.

So is there a plan to drop the older Node?

plmrry:
@Aladdin-ADD I think we would just need to update [.travis.yml](https://github.com/eslint/eslint-plugin-markdown/blob/master/.travis.yml) to not build against those old Node versions?

Aladdin-ADD:
I just created a PR #89 to do it. (maybe need update the docs).

btmills:
@plmrry thanks for doing this! As you've figured out, and to clarify for anyone else reading, the `endLine` key wasn't added to messages until [eslint@3.1.0](https://github.com/eslint/eslint/pull/6640), but this plugin is currently tested all the way back to node 0.10, which means it's tested against eslint@^2, since eslint@3.0.0 [dropped support for node < 4](https://eslint.org/docs/user-guide/migrating-to-3.0.0#dropping-support-for-nodejs--4). That prevents a full integration test, though this change can still be unit tested by manually supplying `endLine` in the processor tests.

I left https://github.com/eslint/eslint-plugin-markdown/pull/89#issuecomment-379581801 trying to figure out what to do with this, and I'd like to hear your thoughts as well.

If we do decide against dropping support for eslint < 4, it's possible with a small tweak to implement this change in a way that's fully backwards-compatible with eslint prior to 3.1.0:

```diff
diff --git a/lib/processor.js b/lib/processor.js
index 113b289..a13b795 100644
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -129,16 +129,17 @@ function adjustBlock(block) {
     return function adjustMessage(message) {

         var lineInCode = message.line - leadingCommentLines;
-        var endLine = message.endLine - leadingCommentLines;
         if (lineInCode < 1) {
             return null;
         }

         var out = {
             line: lineInCode + blockStart,
-            endLine: endLine ? endLine + blockStart : endLine,
             column: message.column + block.position.indent[lineInCode - 1] - 1
         };
+        if (message.endLine >= 0) {
+            out.endLine = message.endLine + blockStart;
+        }

         return assign({}, message, out);
     };
```

tests/lib/plugin.js
```diff
@@ -45,6 +45,51 @@ describe("plugin", function() {
         assert.equal(report.results[0].messages[0].line, 2);
     });
 
+    it.skip("should emit correct line numbers", function() {
```
btmills: These two tests should now pass with ESLint 4!

tests/lib/processor.js
```diff
@@ -514,13 +514,13 @@ describe("processor", function() {
         ].join("\n");
         var messages = [
             [
-                { line: 1, column: 1, message: "Use the global form of \"use strict\".", ruleId: "strict" },
-                { line: 3, column: 5, message: "Unexpected console statement.", ruleId: "no-console" }
+                { line: 1, endLine: NaN, column: 1, message: "Use the global form of \"use strict\".", ruleId: "strict" },
```
btmills: Now that #89 dropped support for ESLint < v4, we should be able to test without any `NaN` here, so this would be 1, and the assertion below would become 4.

PR #89, opened 2018-04-01
Breaking: required node version 6+

Aladdin-ADD:


btmills:
Thanks for this, @Aladdin-ADD! I'll wait until eslint/eslint#10074 is merged and then do here whatever ends up being merged there. I would be happy to merge PRs that upgrade to `eslint^4` in `devDependencies` and add Node 8 and 9 to Travis right now though!

btmills:
> eslint v4 has already dropped node 0.x

Ack, you're totally right. Umm, okay, opinion time: breaking changes in beta releases are bad semver, so technically we can't drop node@0.x support until a hypothetical 2.0, meaning we can't test with eslint@3 or above yet. Both node@0.x and eslint@2 are old, and I'd rather drop node < 4 and upgrade to eslint@4, preferably in a beta. How bad is that?

On a related note, full integration tests for #88 requires [eslint@>=3.1.0](https://github.com/eslint/eslint/pull/6640) so that it emits `endLine`. I've verified that it's possible to do a fully backward-compatible implementation that only moves `endLine` if present, and it can be verified in unit tests, but as long as this runs Travis tests on node < 4, we can't use eslint@>=3.1.0 for full integration tests.

not-an-aardvark:
I think doing a lot of breaking changes in betas is discouraged because it makes migration difficult for early adopters, but it's not strictly a violation of semver. I think it would be fine to drop support for ESLint and/or Node versions in 1.0.

Alternatively, we could just do a 1.0 release as-is and jump directly to 2.0 beta releases.

btmills:
> Alternatively, we could just do a 1.0 release as-is and jump directly to 2.0 beta releases.

I also considered that. In fact our CI might do that automatically if it sees a `Breaking:` commit in a beta? Not sure haha. :soon: Based on my reading of https://github.com/eslint/eslint-release/blob/80a2e80bf374874592fee4dbf3a919e30b48f285/lib/release-ops.js#L92, it just trusts whatever we specify. So I can go along with either. Any preferences? If none, I'll default to doing it in a 1.0 beta.

not-an-aardvark:
I have no strong preferences, although I'd slightly prefer to do it as a 1.0 beta.

Aladdin-ADD:
@btmills sounds good! I will update asap. 😄

Aladdin-ADD:
updated `engines` in `package.json`, friendly ping @btmills 😄

PR #92, opened 2018-04-08
Build: Upgrade to eslint-release@0.11.1

btmills:
This brings in @not-an-aardvark's eslint/eslint-release#20 so that release notes get published as a GitHub release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "chai": "^3.0.0",
     "eslint": "^2.2.0",
     "eslint-config-eslint": "^3.0.0",
-    "eslint-release": "^0.10.2",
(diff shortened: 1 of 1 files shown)


not-an-aardvark:
It's not a big deal, but I generally try to avoid putting usernames in commit messages so that the user doesn't get pinged every time a commit is merged or rebased onto a different branch. (This probably matters more for projects where there are multiple active branches/forks, since old commits are rebased often.)

btmills:
Ah, I didn't know they pinged for commit messages too - I just pushed an amended commit that removes your username. Will merge as soon as CI passes.

PR #93, opened 2018-04-13
Docs: Updating CLA link

pnevares:
Howdy! 👋 

Was looking at contributing to the project but the CLA link in CONTRIBUTING.md wasn't loading. This update looks like the right one.

Thank you!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing Code
 
-Please sign the ESLint [Contributor License Agreement](https://cla.js.foundation/eslint/eslint)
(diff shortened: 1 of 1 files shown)


pnevares:
I can definitely update it to that path if I can get confirmation it's the correct one. Thanks for the note!

platinumazure:
@kborchers Sorry to bother you, but I just want to confirm that the CLA URL paths should generally look like https://cla.js.foundation/<org>/<repo>, right? So the CLA for this repo should be eslint/eslint-plugin-markdown? Thanks!

kborchers:
@platinumazure that is correct though CLA signatures are valid on an org basis so technically if they went to /eslint/eslint and signed it would work on this repo going forward as well.

pnevares:
Thank you for the merge!

PR #95, opened 2018-05-26
Docs: Suggest disabling strict in .md files (fixes #94)

### LINKED ISSUE #94: [question/docs]: suppress `strict`
In docs often avoiding `use strict` directive. Should we suppress this by default or documentation this? Looks even https://eslint.org/ doesn't use `use strict` in docs.

btmills:


README.md
```diff
@@ -129,6 +129,23 @@ Since code blocks are not files themselves but embedded inside a Markdown docume
...
+// .eslintrc.json
+{
+    // ...
+    "overrides": [{
+        "files": ["**/*.md"],
+        "rules": {
+            "strict": "off"
+        }
+    }]
+}
```
not-an-aardvark: Should we also recommend `parserOptions: { impliedStrict: true }` so that users can effectively always have strict mode?

btmills: Great idea! Updated.

PR #97, opened 2018-06-05
New: Enable autofix (--fix)

### LINKED ISSUE #58: Plugin does not work with autofix
ref: https://github.com/AtomLinter/linter-eslint/issues/815

Everything just started with https://github.com/AtomLinter/linter-eslint/issues/779. 

I just have a markdown file, javascript file and eslintrc.

I'm using:

```
% ❯ npm ls | grep eslint                                                                               February 12, 12:34:53
├── eslint@3.15.0
├── eslint-config-standard@6.2.1
├── eslint-plugin-markdown@1.0.0-beta.3
├── eslint-plugin-promise@3.4.1
├── eslint-plugin-standard@2.0.1
```

and have this config

```json
{
  "extends": ["standard"],
  "plugins": ["markdown"],
  "rules": {
    "semi": [
      2,
      "always"
    ]
  }
}
```

Intentionally makes `standard` to always want semicolons so it is easy to see if autofixing works.

Initially I was thinking it is Linter-ESLint bug, but seems it isn't. Because when run the following ESLint still report errors and does not fix them.

```
eslint --fix --ext md README.md
eslint --fix --ext md .
eslint --fix -c .eslintrc --ext md README.md
```

the code fences in the markdown file is both `jsx` and `js`

byzyk:
Fixes #58

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -121,6 +121,8 @@ function adjustBlock(block) {
     }, 0);
 
     var blockStart = block.position.start.line;
+    var blockEnd = block.position.end.line;
+    var blockLines = blockEnd - blockStart;
 
     /**
      * Adjusts ESLint messages to point to the correct location in the Markdown.
@@ -143,11 +145,13 @@ function adjustBlock(block) {
 
         var adjustedFix = {};
         if (message.fix) {
+            var blockOffset = block.position.end.offset - block.position.end.column + (block.position.start.column - 1);
             var codeLength = block.value.length;
             var commentsLength = block.comments.reduce(function(acc, curr) {
                 return acc + curr.length + 1;
             }, 0);
-            var offset = block.position.end.offset - block.position.end.column - codeLength - commentsLength;
+            var indentLength = (block.position.indent[lineInCode - 1] - 1) * (blockLines - lineInCode);
+            var offset = blockOffset - codeLength - commentsLength - indentLength;
             adjustedFix.fix = {
                 range: message.fix.range.map(function(range) {
                     return range + offset;
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -143,7 +143,11 @@ function adjustBlock(block) {
 
         var adjustedFix = {};
         if (message.fix) {
-            var offset = block.position.end.offset - 4 - block.value.length;
+            var codeLength = block.value.length;
+            var commentsLength = block.comments.reduce(function(acc, curr) {
+                return acc + curr.length + 1;
+            }, 0);
+            var offset = block.position.end.offset - block.position.end.column - codeLength - commentsLength;
             adjustedFix.fix = {
                 range: message.fix.range.map(function(range) {
                     return range + offset;
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -88,8 +88,8 @@ var leadingWhitespaceRegex = /^\s*/;
  * whitespace than the opening fence, so not all lines are guaranteed to have
  * the same column offset as the opening fence.
  *
- * The source code of a non-configuration-comment line in the linted JS is
- * a suffix of the corresponding line in the Markdown code block. There are no
+ * The source code of a non-configuration-comment line in the linted JS is a
+ * suffix of the corresponding line in the Markdown code block. There are no
  * differences within the line, so the mapping need only provide the offset
  * delta at the beginning of each line.
  *
@@ -97,11 +97,10 @@ var leadingWhitespaceRegex = /^\s*/;
  * @param {ASTNode} node A Markdown code block AST node.
  * @param {comments} comments List of configuration comment strings that will be
  *     inserted at the beginning of the code block.
- * @returns {number[]} A flattened list of 2-tuples, where the even-indexed
- *     elements are range offsets in the linted JS used as a lookup key, and the
- *     next odd-indexed element is the offset delta from linted JS to Markdown.
- *     Add this value to a range from the linted JS to get the equivalent
- *     location in the Markdown source.
+ * @returns {object[]} A list of offset-based adjustments, where lookups are
+ *     done based on the `js` key, which represents the range in the linted JS,
+ *     and the `md` key is the offset delta that, when added to the JS range,
+ *     returns the corresponding location in the original Markdown source.
  */
 function getBlockRangeMap(text, node, comments) {
     var baseIndent,
@@ -157,7 +156,10 @@ function getBlockRangeMap(text, node, comments) {
      * the lookup index will also be 0, and the lookup should always go to the
      * last range that matches, skipping this initialization entry.
      */
-    rangeMap = [0, 0];
+    rangeMap = [{
+        js: 0,
+        md: 0
+    }];
 
     // Start the JS offset after any configuration comments.
     jsOffset = commentLength;
@@ -181,13 +183,13 @@ function getBlockRangeMap(text, node, comments) {
         // fence, so keep any additional indentation beyond that.
         trimLength = Math.min(baseIndent, leadingWhitespaceLength);
 
-        rangeMap.push(
-            jsOffset,
+        rangeMap.push({
+            js: jsOffset,
             // Advance `trimLength` character from the beginning of the Markdown
             // line to the beginning of the equivalent JS line, then compute the
             // delta.
-            mdOffset + trimLength - jsOffset
-        );
+            md: mdOffset + trimLength - jsOffset
+        });
 
         // Accumulate the current line in the offsets, and don't forget the
         // newline.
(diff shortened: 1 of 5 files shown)


olstenlarck:
Such little code? Coool. :)

btmills:
Wow, this is excellent, thank you! :tada: I had no idea it could take so little code. Can you add some tests? I'm thinking two sets:

1. In `tests/lib/processor.js`, manually define fixes and make sure the transformed messages have the correct ranges.
1. In `tests/lib/plugin.js`, check the output to make sure the fixed code looks as intended.

I care a lot more about checking edge cases in the processor tests, and a couple integration tests for the plugin ought to be sufficient.

byzyk:
Yeah, turns out that's all we need to enable the feature. Already tried this in our webpack docs and it worked as expected - fixed around 500 linter issues in under 4 seconds.

Sure thing will add some test cases, probably tomorrow. Gonna ping you once ready for review :)

byzyk:
@btmills please have a look at the test cases I've added.

I had to move `new CLIEngine` part into a separate helper function since I needed to initiate CLI with `fix: true` to be able to test actual autofix feature. Let me know if you feel like we should move it to separate file, perhaps to `/tests/utils/cli-engine.js`?

tests/lib/plugin.js
```diff
@@ -10,6 +10,31 @@ var assert = require("chai").assert,
     path = require("path"),
     plugin = require("../..");
 
+/**
+ * Helper function which creates CLIEngine instance with enabled/disabled autofix feature.
+ * @param {boolean} [isAutofixEnabled=false] Whether to enable autofix feature.
+ * @returns {CLIEngine} CLIEngine instance to execute in tests.
+ */
+function initCLI(isAutofixEnabled) {
```
byzyk: I feel like this function should be moved to its own file.

btmills: Since this isn't used anywhere else right now, I have no complaints about leaving it right here where you put it 👍

byzyk:
Hey @btmills! Thanks again for edge test cases, I've added them to the PR. I've managed to solve all of them except this one: `inside a list`. 

I tried to make use of `block.position.indent` property to adjust offset for indented blocks accordingly but failed. Seems like there is a bug in the parser as no matter how many spaces I indent the code block with it always equal to 3 for every line in the code block. At the same time it always equal to 1 for not indented blocks.

On a side note, the test passes when it's only one line in the code block, which is really weird.

Perhaps I'm missing something here, I'd appreciate if you could be of any help. I've added two test cases: `inside a list single line` and `inside a list multi line` so you could see the difference between them.

Thanks!

byzyk:
Hey @btmills! 

I've finally managed to overcome all the obstacles and have covered all edge cases you shared with me. Please have a look at the updated code and let me know if this would be good enough to proceed.

Thanks

btmills:
Hey @byzyk! Great work covering those edge cases :clap: I'm really excited about this :tada:

Based on your comment about the indent weirdness, and building on your two list indent tests, I started digging to see what the boundaries of the indent edge cases might be, starting with the [CommonMark spec](https://spec.commonmark.org/0.28/#fenced-code-blocks). Eliding heavily, here's the relevant portions:

> A fenced code block begins with a code fence, indented no more than three spaces. If the leading code fence is indented N spaces, then up to N spaces of indentation are removed from each line of the content (if present). (If a content line is not indented, it is preserved unchanged. If it is indented less than N spaces, all of the indentation is removed.)

I can see the code as it currently stands works with fenced code blocks indented by 2 spaces, so with that as a starting point and the spec as a boundary, I tried 3 other cases:

1. One-space indentation

	````md
    This is Markdown.

     ```js
     console.log('Hello, world!')
     console.log('Hello, world!')
     ```
    ````

1. Three-space indentation

	````md
    This is Markdown.

       ```js
       console.log('Hello, world!')
       console.log('Hello, world!')
       ```
    ````

1. And finally, abandoning all decency, inconsistent indentation

	````md
    This is Markdown.

       ```js
     console.log('Hello, world!')
      console.log('Hello, world!')
        console.log('Hello, world!')
       ```
    ````

    For reference, after parsing, the three lines above should have indentation within the code block of 0, 0, and 1 spaces due to the truncation of underindents described in the spec quote above.

In all three code blocks, `position.start.column` is `1` and - similar to your observation [above](https://github.com/eslint/eslint-plugin-markdown/pull/97#issuecomment-396040724) - `position.indent` is an array of all `1`s. I'm really not sure what to make of that. 😩

Other avenues I'm researching:

- We're a couple versions behind on the parser. Is this perhaps a bug that's been fixed?
- The [AST spec](https://github.com/syntax-tree/unist#position) used by the parser includes definitions for `Position`. Is it possible I'm misinterpreting the intended meaning of the properties?
- There's a [utility](https://github.com/syntax-tree/unist-util-source) to pull the original source code for a Markdown AST node. Does that give us the information we need?
- I'm pretty sure these problems have existed all along, but nobody really noticed if an error message was off by a character or two in strangely-indented blocks, and autofix is the first time the need for positional accuracy has exposed these issues. Do we say "weird indentation is not supported, sorry" and move on?

So that's where I'm stuck. Thoughts on this are appreciated, brilliant insights even more so! :sparkles:

byzyk:
Hey @btmills! Thanks for digging into it on such deep level.

I assume none of the above examples are working correctly with the current implementation, right? That's really frustrating as there is no clear definition of `indent` property in AST specs, so I don't think we're misinterpreting anything here.

I suggest we can open an issue in the parser, asking for clarifications from its maintainers and add a note to the readme file, something like:
> This plugin supports `--fix` feature... blah-blah... to make it work inside lists please indent fenced code block with 2 spaces, because of [link to the parser issue here]...

And merge it as it (with readme note) just to get this going. 

Let's see where it's gonna take us, perhaps we just discovered a bug in parser or we just really misunderstood something. Regardless, some help from the outside won't harm. What do you think?

byzyk:
Hey @btmills, just a friendly ping 🙂 What's the best way to move forward with this feature in your opinion?

btmills:
Hi @byzyk! Sorry for the delay on this. I just went back in with fresh eyes to see if I could figure anything out, and I made no more progress than either one of us did last time. With that in mind, I think your suggestion makes sense:

1. Open an issue on the parser, linking this PR.
1. Add a note to the readme to specify that we're only able to support 2-space indentation pending resolution of the newly-created issue above.
1. Merge this and release!

Feel free to go ahead and do the first two, or let me know if you'd like me to handle those!

SimenB:
Any news here? 🙂

byzyk:
@SimenB I guess maintainers still have to review changes and unfortunately we haven't been able to make the feature work for code blocks inside lists.

btmills:
:tada: d2243ac :unicorn:

First off, sorry for the lack of responsiveness here. I really want to release v1.0.0, and I really want it to include autofix, but I didn't like that it could break code in certain valid, albeit rare, edge cases. I'd been coming back to this off and on and couldn't figure out a solution, and I kept putting it off until now.

I think I've finally figured out how to handle the edge cases. Please try to break it, and send a test case if you can! I did my best to comment the code to explain what's going on, but let me know if I missed anything.

I'll give this some time for people to look over then merge and publish a release candidate if all goes well.

byzyk:
Awesome work man! This is really cool that you've managed to overcome it in the end 👍🏽

I'll try to scrap some time this week to play around it. Will keep you posted.

lib/processor.js
```diff
@@ -66,6 +67,136 @@ function getComment(html) {
...
+     * For each line, determine how much leading whitespace was trimmed due to
+     * indentation. Increase the JS lookup offset by the length of the line
+     * post-trimming and the Markdown offset by the total line length.
+     */
+    for (i = 0; i + 1 < lines.length; i++) {
+        line = lines[i + 1];
+        leadingWhitespaceLength = leadingWhitespaceRegex.exec(line)[0].length;
+        // The parser trims leading whitespace up to the level of the opening
+        // fence, so keep any additional indentation beyond that.
+        trimLength = Math.min(baseIndent, leadingWhitespaceLength);
```
byzyk: 👍🏽

lib/processor.js
```diff
@@ -66,6 +67,136 @@ function getComment(html) {
...
+ * @param {string} text The text of the file.
+ * @param {ASTNode} node A Markdown code block AST node.
+ * @param {comments} comments List of configuration comment strings that will be
+ *     inserted at the beginning of the code block.
+ * @returns {number[]} A flattened list of 2-tuples, where the even-indexed
+ *     elements are range offsets in the linted JS used as a lookup key, and the
+ *     next odd-indexed element is the offset delta from linted JS to Markdown.
+ *     Add this value to a range from the linted JS to get the equivalent
+ *     location in the Markdown source.
+ */
```
byzyk: Just a small suggestion, do you think we can use a simple array of objects here like this `[{jsKey: 1,mdOffset:1},...]`?

I guess if we make it more explicit it will improve readability because took me a bit to get the concept here.

lib/processor.js
```diff
@@ -140,7 +274,25 @@ function adjustBlock(block) {
...
+        var adjustedFix = {};
+        if (message.fix) {
+            adjustedFix.fix = {
+                range: message.fix.range.map(function(range) {
+                    // Advance through the block's range map to find the last
+                    // matching range.
+                    var i = 0;
+                    while (i < block.rangeMap.length && block.rangeMap[i] < range) {
+                        i += 2;
+                    }
```
byzyk: See my comment above.

Then here we can just simply iterate through the array, and be more explicit. What do you think?

lib/processor.js
btmills: Good idea. 9c221e7 implements this change. What do you think?

byzyk: Awesome, that's exactly what I had in mind 👍🏽

byzyk:
Hey @btmills, just wanted to give a quick update.

I removed the note about usage in lists in README. Also played a bit with new functionality and it looks pretty solid to me. That said, it's good to go from my end 👏🏼

Let me know if you need any help before we finally merge this bad boy 😃

tunnckoCore:
sweet! i'll try it

PR #99, opened 2018-10-15
Support multiple words in code's language.

### LINKED ISSUE #98: Should ignore lang string after the first space
If you have a code fence like this:

````markdown see here!
```javascript more stuff
5 + 5
```
````

Most markdown interpreters simply ignore everything after the space and treat the syntax coloring based only on the first part. For example, if you view the source of this very comment, you'll see that the code fence above is "markdown see here!" and not just "markdown". This is very useful because it allows one to provide additional functionality. In particular, on the node site we are going to have ````
javascript runkit` to turn on "runnability" on certain code blocks ( https://github.com/nodejs/node/pull/22831 ). The fix is very simple, just changing:

```javascript
if (node.lang && SUPPORTED_SYNTAXES.indexOf(node.lang.toLowerCase()) >= 0) {
```

to:

```javascript
if (node.lang && SUPPORTED_SYNTAXES.indexOf(node.lang.split(" ")[0].toLowerCase()) >= 0) {
```

I'm happy to submit this change but wanted to run it by everyone here first.

tolmasky:
This fixes #98. I tested this in the node source and it appears to work fine (currently pushed as a forked module to be able to do this test: https://github.com/nodejs/node/pull/22831/commits/6361ed8588dcdc203c41f829f9925ebf0dba51ff )


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "eslint-plugin-markdown",
+  "name": "eslint-plugin-markdown-runkit",
```
btmills: Looks like this snuck in 😈

PR #107, opened 2018-10-30
Fix: Allowing eslint-plugin-prettier to work

simlu:


not-an-aardvark:
Hi, thanks for the pull request!

I think @btmills would be the best person to review this, so I'll leave that to him. Just as a minor note, it would be a bit better to push new updates to a PR rather than opening new ones if there's something wrong with a PR, since that avoids creating a bunch of notifications for many opened PRs.

PR #110, opened 2018-11-06
Chore: Upgrade to eslint-config-eslint@5.0.1

btmills:
The prior config was still enforcing ES5 coding standards. The earliest supported version of Node.js is 6.14.0, so this lets us use all the nice ES6 features.

I split this into multiple commits to ease review:

- b90a6d6 is just `npm install`ing the right versions.
- 02ce704 was done entirely with `npm run lint -- --fix`.
- 24fb9b7 was a bulk find replace `assert.equal` to `assert.strictEqual`.
- 6ca28b5 narrowed variable declaration scope where possible.
- 2612d95 took advantage of the array spread operator.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,10 +5,10 @@
 
 "use strict";
 
-var processor = require("./processor");
+const processor = require("./processor");
 
 module.exports = {
-    "processors": {
+    processors: {
         ".markdown": processor,
         ".mdown": processor,
         ".mkdn": processor,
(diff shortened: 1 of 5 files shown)


klappy:
Is this PR required to enable ES6 linting?

platinumazure:
@klappy I don't think so. eslint-config-eslint is used for linting within this project, so this change benefits contributors to this project but not users of this plugin.

PR #120, opened 2019-09-07
Chore: Add failing test case for indent of multiline fixes

lydell:
If the replacement text of an autofix contains newlines, all lines but
the first in the replacement text will unexpectedly have 0 indentation
if a code block is indented. I guess not only the _range_ of fixes but
also the _text_ of fixes need to be adjusted.

If the code block is placed within a blockquote, a `>` character also
needs to be inserted.

This issue was found while investigating
lydell/eslint-plugin-simple-import-sort#22. That plugin fixes several lines
of imports in one go. This means that most of the imports will have too
little indentation after autofix if the code block is indented (such as
when placed in a list).

### DIFF
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -554,6 +554,52 @@ describe("plugin", () => {
                 assert.strictEqual(actual, expected);
             });
 
+            it("multiline autofix", () => {
+                const input = [
+                    "This is Markdown.",
+                    "",
+                    "   ```js",
+                    "   console.log('Hello, \\",
+                    "   world!')",
+                    "   ```"
+                ].join("\n");
+                const expected = [
+                    "This is Markdown.",
+                    "",
+                    "   ```js",
+                    "   console.log(\"Hello, \\",
+                    "   world!\")",
+                    "   ```"
+                ].join("\n");
+                const report = cli.executeOnText(input, "test.md");
+                const actual = report.results[0].output;
+
+                assert.strictEqual(actual, expected);
+            });
+
+            it("multiline autofix in blockquote", () => {
+                const input = [
+                    "This is Markdown.",
+                    "",
+                    ">   ```js",
+                    ">   console.log('Hello, \\",
+                    ">   world!')",
+                    ">   ```"
+                ].join("\n");
+                const expected = [
+                    "This is Markdown.",
+                    "",
+                    ">   ```js",
+                    ">   console.log(\"Hello, \\",
+                    ">   world!\")",
+                    ">   ```"
+                ].join("\n");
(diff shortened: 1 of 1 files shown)


lydell:
FYI: I don't intend to try to fix this issue at the moment. Just wanted to let you know about it.

btmills:
Well this is fun. Thank you for creating the tests! I've started playing with a potential solution in [a branch](https://github.com/eslint/eslint-plugin-markdown/compare/issue120), and I've gotten the tests passing without breaking any other tests. What do you think? You're right that it needs to adjust the *text* of fixes, which it wasn't doing before. That branch inserts whatever leading text was found on the first line of the code block after each newline in a multi-line fix. I can't decide if that's so simple it can't possibly work, or simple enough that it just might do the trick.

lydell:
Will it work with a underindented code? This one is tricky because the first line has less indentation than the second. Does it matter?

``````md
   ```js
 console.log('Hello, world!')
  console.log("Hello, \
  world!")
     console.log('Hello, world!')
   ```
`````

Here’s a similar case:

``````md
   ```js
  console.log('Hello, world!')
 console.log("Hello, \
 world!")
     console.log('Hello, world!')
   ```
`````

This time the first line has _more_ indenation. We must make sure that we don’t accidentally end up with more spaces in the JS string, altering its `.length`.

Here are two other test cases I’d like to see:

- CRLF tests.
- A code block inside a blockquote inside _another_ blockquote.

**Comment by lydell** [comment]
Also, what does this `while` loop do?

```js
    while (startOffset > 0 && text[startOffset - 1] !== "\n") {
        startOffset--;
    }
```

btmills:
Take a look at #124. I added tests for underindenting and nested code blocks. The only caveat for underindenting is that it isn't preserved, though the fix is applied correctly. [Example](https://github.com/eslint/eslint-plugin-markdown/pull/124/files#diff-50f17d3033cf86483cf4afecb228f6a1R597) where the line was previously underindented by one level, and after the fix it matches the level of the fence. Extra indents are correctly preserved.

Good call on CRLF. I'm going to address that separately in #125 since it doesn't even require any indenting to fail. Pretty sure the cause is here:
https://github.com/eslint/eslint-plugin-markdown/blob/dc909618aa8f39e84279f5bdeb4a888d56d919b1/lib/processor.js#L180-L183

That line of code was a hack to get to the beginning of the line so I knew whether to add a blockquote character. Turns out I don't need to search backwards through the string; I can just [do the math](https://github.com/eslint/eslint-plugin-markdown/pull/124/files#diff-1ea033eede7961e81fac67f7cf52fd51R79) with the node's position, so it's gone in #124.

PR #124, opened 2019-10-08
Fix: Indent multiline fixes (fixes #120)

btmills:
Thanks to @lydell for discovering this bug and writing the tests to reproduce it in #120.

If I lint the following Markdown source text containing a JS code block inside a blockquote:

````md
This is Markdown.

>   ```js
>   console.log('Hello, \
>   world!');
>   ```
````

What ESLint sees is this:

```js
console.log('Hello, \
world!');
```

The fix text sent to the processor will contain the string, now with double quotes, and preserving the newline as-is:

```
"Hello, \\\nworld!"
```

The fix would then be applied incorrectly because it didn't account for the leading `>   ` in the Markdown source text:

````md
This is Markdown.

>   ```js
>   console.log("Hello, \
world!" world!');
>   ```
````

This change tracks the leading text that precedes the first line of any fenced code block and inserts the same prefix after any newline in a fix, so the fix text in the above example would become:
```
"Hello, \\\n>   world!"
````

### DIFF
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -590,6 +590,11 @@ describe("plugin", () => {
                     "     console.log('Hello, world!')",
                     "   ```"
                 ].join("\n");
+
+                // The Markdown parser doesn't have any concept of a "negative"
+                // indent left of the opening code fence, so autofixes move
+                // lines that were previously underindented to the same level
+                // as the opening code fence.
                 const expected = [
                     "   ```js",
                     " console.log(\"Hello, world!\")",
@@ -644,6 +649,11 @@ describe("plugin", () => {
                     "> >    world!')",
                     "> >   ```"
                 ].join("\n");
+
+                // The Markdown parser doesn't have any concept of a "negative"
+                // indent left of the opening code fence, so autofixes move
+                // lines that were previously underindented to the same level
+                // as the opening code fence.
                 const expected = [
                     "This is Markdown.",
                     "",
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -554,6 +554,52 @@ describe("plugin", () => {
                 assert.strictEqual(actual, expected);
             });
 
+            it("multiline autofix", () => {
+                const input = [
+                    "This is Markdown.",
+                    "",
+                    "   ```js",
+                    "   console.log('Hello, \\",
+                    "   world!')",
+                    "   ```"
+                ].join("\n");
+                const expected = [
+                    "This is Markdown.",
+                    "",
+                    "   ```js",
+                    "   console.log(\"Hello, \\",
+                    "   world!\")",
+                    "   ```"
+                ].join("\n");
+                const report = cli.executeOnText(input, "test.md");
+                const actual = report.results[0].output;
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/lib/plugin.js [hunk 1]
```diff
@@ -554,6 +554,114 @@ describe("plugin", () => {
...
+                    "  console.log('Hello, \\",
+                    "  world!')",
+                    "     console.log('Hello, world!')",
+                    "   ```"
+                ].join("\n");
+                const expected = [
+                    "   ```js",
+                    " console.log(\"Hello, world!\")",
+                    "  console.log(\"Hello, \\",
+                    "   world!\")",
```
lydell: I’m not 100% sure how leading whitespace works in weird code blocks like this, but didn’t we cause one more space to appear in the JS string here?

tests/lib/plugin.js
```diff
@@ -554,6 +554,114 @@ describe("plugin", () => {
...
+                    "> >   ```"
+                ].join("\n");
+                const expected = [
+                    "This is Markdown.",
+                    "",
+                    "> This is a nested blockquote.",
+                    ">",
+                    "> >   ```js",
+                    "> >  console.log(\"Hello, \\",
+                    "> >   world!\")",
```
lydell: Same thing here? Is this 1–2 too many spaces?

lydell:
Btw, I’d be fine not bothering with the two potential edge cases above. It’s better to fix the bug in the common case, and leave the very uncommon case for later.

tests/lib/plugin.js
btmills: I just pushed 80edd59 that adds comments explaining why this occurs for the benefit of future readers of these tests.

When a line inside a fenced code block is indented less than the opening fence, Markdown treats it as having 0 indent, even though its indent in the source is negative. The parser doesn't provide any information to distinguish lines that begin at the same level as the opening fence from those that are underindented. Whenever one of those lines gets autofixed, it gets shifted to align with the opening fence. Lines that that have positive indentation inside the code fence retain their original indentation because those are representable.

tests/lib/plugin.js
btmills: Yep, same cause. I added comments in 80edd59 and explained further in https://github.com/eslint/eslint-plugin-markdown/pull/124#discussion_r333814825.

lydell:
Looks like this is ready to go, then?

tests/lib/plugin.js
lydell: I checked this case in https://spec.commonmark.org/dingus/ and the actual and expected code snippets produce the same rendered JS, with the same amount of spaces between "Hello," and "world!".

tests/lib/plugin.js
lydell: I checked this case in https://spec.commonmark.org/dingus/ and the actual and expected code snippets produce the same rendered JS, with the same amount of spaces between "Hello," and "world!".

lydell:
As mentioned in the inline comments above, after checking on https://spec.commonmark.org/dingus/ there seems to be no problem with the current implementation! :tada:

PR #126, opened 2019-10-31
Base indent for multiple line breaks; npm

brettz9:
- Fix: Apply base indent to multiple line breaks; amend test to check
- Testing: Get plugins tests to work with eslint 6
- Teting: Expect an extra line break since `remark-parse@7` started preserving them (see https://github.com/remarkjs/remark/issues/351 )
- Linting: Remove call to lint absent `Makefile.js`
- Linting: Add ignore file and explicitly exclude undesired
- npm: Update object-assign, unified, remark-parse deps.
- npm: Update eslint-release, chai, mocha, eslint devDeps
- npm: Change from deprecated istanbul to nyc
- npm: Add ignore file

Note that the only dep./devDep. that had an upgrade but which I didn't upgrade was `eslint-plugin-node`, as `eslint-config-eslint` is still tied to an older version thereof.

Btw, the reason I changed to explicitly ignoring more files was so that the likes of my IDE linter wouldn't try to lint files which you hadn't excluded, even though your npm script wasn't targeted at them previously. I also figured it safer to lint all by default and just exclude the ones you don't want linted.


brettz9:
Btw, I didn't update `travis.yml` to handle the eslint@6 upgrade (though I can if you wish), since a number of projects (such as eslint itself) have been dropping Node 6 support with it reaching end-of-life, and I didn't know if you might wish to also (in which case, it would just be a matter of dropping that Node line from `travis.yml`).

And I'm not sure why the commit message is still failing despite my attempting to follow the eslint guidelines.

btmills:
Hi @brettz9, thanks for taking the initiative to do all of this! As-is, this pull request does too many different things, so let's split it up into a couple:

1. I want to get #127 fixed right away, so let's start with a pull request with just 538c442.
1. A pull request with f924c99 would also be good - I see why the `.eslintignore` approach makes it easier to work on this with editor integrations. If you want to include it in that same PR, I'd also welcome a second commit that stops ignoring `tests/lib/processor.js` and runs `eslint --fix` and fixes any other issues so that it passes. I'm not sure why that file was ever omitted in the first place.

I'm less convinced of need for the `.npmignore` file and dependency upgrades, but if you'd like to either open issues or pull requests for those I'm happy to discuss them there. Short version here:

- We already have use the [`files` array in `package.json`](https://github.com/eslint/eslint-plugin-markdown/blob/2d306bad6416f88ae8d25a5e4a188fc5d719fd41/package.json#L33-L37), so [the published package doesn't contain unnecessary files](https://unpkg.com/browse/eslint-plugin-markdown@1.0.1/).
- We'll need to do a major version bump to support the [new processor API](https://github.com/eslint/rfcs/tree/master/designs/2018-processors-improvements), so my inclination is to do all the dependency upgrades as part of that release.

I'll keep this open for now to track the changes to fix #127 and modify `.eslintignore`, and then this can be closed once those pull requests exist.

brettz9:
I've filed #128 and #129 (though not sure why *that* commit message is failing unless it needs an issue despite the docs suggesting it wouldn't). Re: npmignore, I always forget to check for `files`, yeah, I see not needed.

brettz9:
Btw, I see the processor API is already ready with https://github.com/eslint/eslint/pull/11552 and the docs updated: https://eslint.org/docs/developer-guide/working-with-plugins . 

FWIW, in applying the new processor API, you may still find my work updating the deps/devDeps helpful as it took me some time to figure out, e.g., my not finding anywhere that `plugins` was now needed with `CLIEngine`; I have the branch at https://github.com/brettz9/eslint-plugin-markdown/tree/deps-devDeps .

btmills:
> you may still find my work updating the deps/devDeps helpful as it took me some time to figure out

Thanks so much for the reference! I appreciate having that available as a reference.

btmills:
If you're willing to pull just the `devDependencies` changes from 9a65621 into another `Chore: ...` PR I'd be happy to merge that too! I should've thought about that distinction in https://github.com/eslint/eslint-plugin-markdown/pull/126#issuecomment-548576076 - it's the `remark` and `unified` upgrades that I want to wait to do.

PR #130, opened 2019-11-01
Upgrade: Update devDeps and change istanbul -> nyc

brettz9:
Upgrade: Update devDeps and change istanbul -> nyc

Update eslint-release, chai, mocha, eslint devDeps. In process of updating,
get plugins tests to work in anticipation of eslint 6.

Note that this still doesn't include logic to exclude ESLint 6 in Node 6 on Travis (or, if you preferred, to bump `engines` to 8.10.0 and drop Travis testing of 6).

If you like, I can also upgrade the deps. to the versions prior to breaking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ coverage/
 node_modules/
 npm-debug.log
 .eslint-release-info.json
+.nyc_output
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "scripts": {
     "lint": "eslint .",
     "test": "npm run lint && npm run test-cov",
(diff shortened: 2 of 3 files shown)


btmills:
> If you like, I can also upgrade the deps. to the versions prior to breaking.

Good idea, using ESLint v5 will probably be the easiest route. The other option could be some form of build matrix that uses older versions of ESLint on Node.js v6 and ESLint v6 everywhere else, but I don't think that's worth the effort since supporting the new processor API will require dropping support for ESLint < v6 soon anyway.

brettz9:
Cool... Yeah, I tried with different versions of eslint including reverting, but a Travis issue makes sense.

PR #133, opened 2019-11-29
New: support more syntaxes for vue

Amour1688:
I'd like to use Vue.js in markdown.

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -9,7 +9,7 @@ const assign = require("object-assign");
 const unified = require("unified");
 const remarkParse = require("remark-parse");
 
-const SUPPORTED_SYNTAXES = ["js", "javascript", "node", "jsx"];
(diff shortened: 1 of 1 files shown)


btmills:
:wave: Thanks for taking the time to open this pull request, @Amour1688! The first version of ESlint's processor API didn't have a great solution for extracting code blocks like Vue code that aren't plain JS. That's why the hard-coded `SUPPORTED_SYNTAXES` list had to exist in v1 of this plugin. The good news is that a new processor API shipped in ESLint v6 that will allow this plugin to extract code blocks containing any syntax, and you'll be able to configure what should happen inside your `.eslintrc.js` without this plugin acting as gatekeeper. I'm now actively working on implementing support for the new API, so if you're still interested in linting Vue code inside your Markdown documents, follow #138 for updates. I'll close this now that a first-class supported solution is available and use #138 as the primary tracking issue for that work.

PR #152, opened 2020-07-18
Docs: Add React example

btmills:

PR #155, opened 2020-07-28
Docs: Add TypeScript example

btmills:
#153 and #152 should be merged first.


btmills:
@kaicataldo today I learned a bit about how type-aware linting works! Unfortunately I can't get it fully working. I can get the first `ts` code block in `README.md` to lint, but the second code block reports a configuration error even though I tried to configure it the same way. I left some comments in 1c78364. Any ideas?

btmills:
@kaicataldo I tried a few more things and still wasn't able to get type-aware linting to work outside the first code block. Unless you have any ideas, what do you think of adding a note to this saying non-type-aware rules work with TS but type-aware rules don't for now?

kaicataldo:
That works for me! I looked at it for a bit and wasn't able to figure it out either. Maybe @bradzacher has an idea?

bradzacher:
I don't believe type-aware linting will work. Type-aware linting relies upon a file on disk, but preprocessors produce "fake" files.

---

Long answer:
typescript-eslint is built on top of typescript itself. For standard linting, TS has APIs which let as parse raw text (and we then convert the TS AST to an ESTree AST).

For type-aware linting, we have to leverage TS's heavy compiler API, which works around having a `tsconfig` which references files on disk.

ESLint preprocessors are a bit of a hack in that a preprocessor essentially splits one file into many "fake" files that don't exist on disk. Because there's no file on disk, the TS API can't find the file specified in its data structures, so we throw an error.

btmills:
Thank you @bradzacher, that's very helpful! Knowing that, I'm curious how type-aware rules worked _at all_ in the first code block in 1c7836429388e4236773b4e2e76641b7b9f80019. Gremlins, apparently.

In da4b9b08250986161a281ef6046a1d042d41109b, I removed `tsconfig.json` and the configuration for type-aware linting and added a brief note to the example's readme. With that, I think this is ready. GitHub auto-dismissed @kaicataldo's approval of the plan that I implemented in the commit, so I'll hold off merging for a couple days in case someone has time to drop by and give it a :+1:

PR #156, opened 2020-08-03
Add rules to recommended config

arvigeus:
It's basically what documentation suggests. One rule I would add, but in documentation only, is `"import/no-unresolved": "off"`


arvigeus:
Well, I managed to screw this up royally.

PR #163, opened 2020-11-03
chore: fix failing tests

Trott:
Two tests are written that require ESLint@7, but the project specifies
ESLint@6. Move to ESLint@7.

BREAKING CHANGE: ESLint@7 drops support for Node.js 8.x.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,9 +1,9 @@
 "use strict";
 
-var fs = require("fs");
-var path = require("path");
-var PACKAGE_NAME = require("./package").name;
-var SYMLINK_LOCATION = path.join(__dirname, "node_modules", PACKAGE_NAME);
+const fs = require("fs");
+const path = require("path");
+const PACKAGE_NAME = require("./package").name;
+const SYMLINK_LOCATION = path.join(__dirname, "node_modules", PACKAGE_NAME);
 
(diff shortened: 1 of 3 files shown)


Trott:
(Fixed the issues flagged by the bot and force pushed. This is ready to be reviewed.)

btmills:
Hi @Trott, thanks for pointing this out! This plugin is compatible with the ESLint v6 processor API, and most of the unit tests use that. The examples use the friendlier ESLint v7 processor API. As a way to verify both work, we [run the examples](https://github.com/eslint/eslint-plugin-markdown/blob/2749b4deb8a8f8015721ecb5eb49bec8de2042c4/package.json#L23) as an additional test step. CI runs the [`install-examples` npm script](https://github.com/eslint/eslint-plugin-markdown/blob/2749b4deb8a8f8015721ecb5eb49bec8de2042c4/package.json#L23) before `npm test` so that the examples have ESLint v7 installed. However, I see `npm run install-examples` is missing from the contributing docs, so `git clone ...; cd ...; npm install; npm test` will fail. Any ideas?

_Ideally_ there'd be a way to automatically run `npm install` for the examples as part of a local `npm install` but not when the plugin is installed as a dependency. Then the separate `install-examples` step wouldn't be necessary. Maybe we could abuse the `prepare` script for that? Worst case we can add `npm run install-examples` to the contribution section at the bottom of the readme, but something automatic would be much better.

PR #164, opened 2020-11-15
Chore: Automatically install example dependencies

btmills:
Previously, CI was explicitly running the `install-examples` script as its own step. This was also necessary but not documented locally, so `git clone; npm install; npm test` would show a couple failures when the examples inherited the wrong ESLint dependency. The examples' dependencies should now be installed automatically when running `npm install` locally without any arguments.

Originally reported in #163.

**For reviewers:** Based on my reading of https://docs.npmjs.com/cli/v6/using-npm/scripts, this is a safe usage of the `prepare` script, but I'd appreciate a second set of eyes on that. The goal is to run as part of `npm install` locally but not when the plugin is a dependency. The additional runs before packing and publishing are unnecessary but I don't think they should cause a problem. I'll do a `rc.1` release after merging this just to make sure there aren't any unexpected kinks. The fallback plan is just to document the second step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - name: Install Packages
-      run: npm install && npm run install-examples
+      run: npm install
       env:
         CI: true
     - name: Test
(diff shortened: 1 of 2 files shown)


nzakas (approved):
Yup, that should work. "prepare" will be called before "install" when running npm install.

PR #167, opened 2020-12-16
Fix: Ignore words in info string after syntax (fixes #166)

### LINKED ISSUE #166: [v2.0.0-rc.0] Broken support for extended fence language
This is a regression of the feature introduced by a5d0ccec6f14f4808a535c0f06273d1e9892be15. 

For example, using the same input as the unit test:

````
```js more words are ignored
var answer = 6 * 7;
```

```js
var answer = 6 * 7;
```
````

Running ESLint on the markdown above:

```
> eslint *.md


test.md
  6:1  error  Unexpected var, use let or const instead  no-var

```

Should report 2 errors with the following config:

```js
module.exports = {
  extends: 'plugin:markdown/recommended',
  rules: {
    'no-var': 'error'
  }
};
```

I haven't been able to run tests locally (failed to `npm install` this repo on Windows) but I think that the [associated unit test](https://github.com/eslint/eslint-plugin-markdown/blob/d30c50f46237af2fdef0a8a21fb547ed8e6c4d80/tests/lib/processor.js#L271-L280) doesn't guarantee that the returned block will be processed by ESLint. I guess the returned `block.filename` is `'0.js more words are ignored'` which doesn't pass the glob matching on this virtual block filename.

btmills:
This was working correctly in v1 but stopped working when v2 switched to the new processor API. The generated filename was `0.js more words are ignored`. The test case for this was insufficiently strict: it asserted that the code block was extracted, which was sufficient with hard-coded extensions in v1, but it didn't assert the generated name was `0.js` now that we're taking the extension from the info string.

Per [CommonMark 0.29]:

> The line with the opening code fence may optionally contain some text following the code fence; this is trimmed of leading and trailing whitespace and called the info string.

I've strengthened the tests for generated filename extensions and added a new test to ensure that leading whitespace is trimmed correctly.

[CommonMark 0.29]: https://spec.commonmark.org/0.29/#info-string

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -252,7 +252,9 @@ function preprocess(text) {
     });
 
     return blocks.map((block, index) => ({
-        filename: `${index}.${block.lang.trimStart().split(" ")[0]}`,
+
+        // TODO: Use trimStart() after dropping Node.js 8.x support.
+        filename: `${index}.${block.lang.replace(/^\s+/u, "").split(" ")[0]}`,
         text: [
             ...block.comments,
             block.value,
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -252,7 +252,7 @@ function preprocess(text) {
     });
 
     return blocks.map((block, index) => ({
-        filename: `${index}.${block.lang}`,
+        filename: `${index}.${block.lang.trimStart().split(" ")[0]}`,
         text: [
             ...block.comments,
             block.value,
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -252,9 +252,7 @@ function preprocess(text) {
     });
 
     return blocks.map((block, index) => ({
-
-        // TODO: Use trimStart() after dropping Node.js 8.x support.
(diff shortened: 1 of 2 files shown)


lib/processor.js
```diff
@@ -252,7 +252,9 @@ function preprocess(text) {
     });
 
     return blocks.map((block, index) => ({
-        filename: `${index}.${block.lang}`,
+
+        // TODO: Use trimStart() after dropping Node.js 8.x support.
+        filename: `${index}.${block.lang.replace(/^\s+/u, "").split(" ")[0]}`,
```
simonbrunel, line 257: I don't think you need the `/u` regex flag but can't you just use `trim()` since there is no reason to keep trailing spaces?

`${index}.${block.lang.trim().split(" ")[0]}`

btmills: Good idea, there's no reason this _has_ to trim only the leading whitespace. I went ahead and added a test for trailing whitespace while I was at it.

nzakas (approved):
LGTM

PR #168, opened 2020-12-16
Fix: npm prepare script on Windows (refs #166)

btmills:
#166 originally reported being unable to `npm install` in the repository on Windows. This was due to the changes in #164, which worked fine on Linux and macOS but not Windows. This PR first adds macOS and Windows to the test matrix and then fixes the npm prepare script by moving the previous bash-only logic into JS.

### DIFF
--- a/npm-prepare.js
+++ b/npm-prepare.js
@@ -1,3 +1,9 @@
+/**
+ * @fileoverview Install examples' dependencies as part of local npm install for
+ * development and CI.
+ * @author btmills
+ */
+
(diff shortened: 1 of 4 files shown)


npm-prepare.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
```
nzakas, line 1: Can you add a header comment explaining what this file does?

nzakas (approved):

Issue #78, opened 2017-07-05
First empty line in code fragment silents rules about first empty line in code


vsemozhetbyt:
From `.eslintrc.js:`
```js
  'lines-around-directive': ['error', 'never'],
```
In `.js` file:
```js
<empty line>
'use strict';
console.log(1);

```
Message:
```console
  2:1  error  Unexpected newline before "use strict" directive  lines-around-directive
```
In `.md` file:
<pre>
### Heading

```js

'use strict';
console.log(1);
```
</pre>
No error messages.

nzakas:
@btmills what do you want to do with this issue?

btmills:
`remark-parse` v7 includes https://github.com/remarkjs/remark/pull/423, which should fix this issue.

Issue #77, opened 2017-07-05
First empty line in code fragment breaks line numeration in error messages


vsemozhetbyt:
Test doc:

<pre>
### Heading

```js

console.log('a')
```
</pre>

Message:

```console
  4:17  error  Missing semicolon  semi
```

This should be `5:17`.

btmills:
[`remark-parse` v7](https://github.com/remarkjs/remark/releases/tag/remark-parse%407.0.0) includes https://github.com/remarkjs/remark/pull/423, which should fix this issue and #78.

PR #171, opened 2021-02-02
Chore: upgrade remark-parse to 9.0.0

aduh95:
Changelog: https://github.com/remarkjs/remark/releases

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -252,7 +252,7 @@ function preprocess(text) {
     });
 
     return blocks.map((block, index) => ({
-        filename: `${index}.${block.lang.trim().split(" ")[0]}`,
+        filename: `${index}.${block.lang}`,
         text: [
             ...block.comments,
             block.value,
@@ -286,7 +286,7 @@ function adjustBlock(block) {
 
         const out = {
             line: lineInCode + blockStart,
-            column: message.column + block.position.indent[lineInCode - 1] - 1
+            column: message.column + block.baseIndentText.length
         };
 
         if (Number.isInteger(message.endLine)) {
--- a/tests/lib/processor.js
+++ b/tests/lib/processor.js
@@ -199,7 +199,7 @@ describe("processor", () => {
 
             assert.strictEqual(blocks.length, 1);
             assert.strictEqual(blocks[0].filename, "0.js");
-            assert.strictEqual(blocks[0].text, "\n\n \n  \n");
+            assert.strictEqual(blocks[0].text, "\n\n\n \n  \n");
         });
 
         it("should ignore code fences with unspecified info string", () => {
@@ -362,7 +362,7 @@ describe("processor", () => {
             const blocks = processor.preprocess(code);
 
             assert.strictEqual(blocks[0].filename, "0.js");
-            assert.strictEqual(blocks[0].text, "var answer = 6 * 7;\nconsole.log(answer);\n");
+            assert.strictEqual(blocks[0].text, "var answer = 6 * 7;\r\nconsole.log(answer);\n");
         });
 
         it("should unindent space-indented code fences", () => {
--- a/tests/lib/processor.js
+++ b/tests/lib/processor.js
@@ -646,8 +646,8 @@ describe("processor", () => {
             const result = processor.postprocess(messages);
 
             assert.strictEqual(result[2].column, 9);
-            assert.strictEqual(result[3].column, 2);
-            assert.strictEqual(result[4].column, 2);
+            assert.strictEqual(result[3].column, 4);
+            assert.strictEqual(result[4].column, 4);
         });
 
         it("should adjust fix range properties", () => {
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "nyc": "^14.1.1"
   },
   "dependencies": {
-    "remark-parse": "^5.0.0",
+    "remark-parse": "^9.0.0",
     "unified": "^6.1.2"
   },
   "peerDependencies": {


aduh95:
Blocked on https://github.com/syntax-tree/mdast-util-from-markdown/issues/12.

tests/lib/processor.js
```diff
@@ -199,7 +199,7 @@ describe("processor", () => {
 
             assert.strictEqual(blocks.length, 1);
             assert.strictEqual(blocks[0].filename, "0.js");
-            assert.strictEqual(blocks[0].text, "\n\n \n  \n");
+            assert.strictEqual(blocks[0].text, "\n\n\n \n  \n");
```
btmills, line 202: Aha! This reminded me of both #77 and #78. After some digging, it turns out [`remark-parse@7.0.0`](https://github.com/remarkjs/remark/releases/tag/remark-parse%407.0.0) includes https://github.com/remarkjs/remark/pull/423, which fixes both of those issues. I opened #175 with a few additional tests specifically demonstrating those fixes.

tests/lib/processor.js
```diff
@@ -362,7 +362,7 @@ describe("processor", () => {
             const blocks = processor.preprocess(code);
 
             assert.strictEqual(blocks[0].filename, "0.js");
-            assert.strictEqual(blocks[0].text, "var answer = 6 * 7;\nconsole.log(answer);\n");
+            assert.strictEqual(blocks[0].text, "var answer = 6 * 7;\r\nconsole.log(answer);\n");
```
btmills, line 365: This test is... strange. It says it "preserves" the original line endings, yet the assertion _normalized_ them to `\n`. However, it was consistent. Here I see there's one `\r\n` and another `\n`. Any idea why that is? If v9 of the parser now preserves original line endings, I'd expect both to be `\r\n`. If this behavior changes, I'll want to add some autofix tests that span lines to make sure we're using correct substitutions. That'd be a likely place to have an off-by-one error that causes syntax errors in the output.

lib/processor.js
```diff
@@ -286,7 +286,7 @@ function adjustBlock(block) {
 
         const out = {
             line: lineInCode + blockStart,
-            column: message.column + block.position.indent[lineInCode - 1] - 1
+            column: message.column + block.baseIndentText.length
```
btmills, line 289: I don't think this change is valid in the case of "negative" indents, i.e. lines of code in a code block that are less indented than the opening fence. The parser previously gave us those on a per-line basis so we could make the correct adjustment. That shows up in the change to the "should translate indented column numbers" test, where one of the messages' columns (4) is now greater than the line length (2).

btmills (commented):
@aduh95 thank you for opening this PR. It led me to a fix in v7 of the parser that fixes two long-outstanding bugs! v8 also appears to produce the same results as v7 including the fixes. v9 being a complete rewrite makes me a bit nervous, however, and I commented on two places that might have issues.

tests/lib/processor.js
aduh95: Yep, I agree it's weird. I think the last line of the code block is always normalized to `\n`, while the other line endings are preserved. It can be observed by adding another line inside the code block. Looks like an upstream issue.

aduh95: Oh no, it's actually due to this:

https://github.com/eslint/eslint-plugin-markdown/blob/f6a8ca0824b8a62d0af2a9eb512104564588b49e/lib/processor.js#L256-L260

It always adds a blank line and to the node value, and use `\n` without checking if it's relevant. I'm not sure why we add a blank line here…

For reference, here's the snippet I used to check the upstream implementation is working as expected:

```js
console.log(
  require('mdast-util-from-markdown')([
    "```js",
    "var answer = 6 * 7;",
    "console.log(answer);",
    "```"
  ].join("\r\n")).children[0].value
); // 'var answer = 6 * 7;\r\nconsole.log(answer);'
```

JounQin:
Can we support `remark-parser<9` and `remark-parse>=9` at the same time?

Besides, should `unified` be upgraded too?

btmills:
Update on this: I really want to at least get to v7 to fix those two bugs you made me realize were related. I'm waiting for feedback from the rest of the team in the discussion for #175, which demonstrates those exact fixes, about how comfortable we are bumping parser major versions in a semver-minor release of `eslint-plugin-markdown`. While v7 fixes two known bugs in this plugin, I'm not sure yet about going all the way to v9, partially because it's a rewrite, which introduces more risk, and partially because of the negative indent changes that would require further work.

> Can we support `remark-parser<9` and `remark-parse>=9` at the same time in case the change seems very small?

As long as we're comfortable bumping the dependency to at least v7 in a semver-minor release (hypothetically `eslint-plugin-markdown@2.1.0`), then we can do that immediately to fix those two parser-related bugs. If for some reason we don't want to bump the minimum dependency past v7 but need to support v9, we could look at supporting multiple major versions. That would bring significant implementation and test complexity, so there would need to be some compelling advantage to justify that tradeoff.

> Besides, should `unified` be upgraded too?

That's lower priority than upgrading `remark-parse` because I'm not aware of any bugs caused by `unified`, but if someone wanted to look through the `unified` changelogs and verify that none of the changes will break `eslint-plugin-markdown`, sure.

btmills:
I just merged #175, which includes the two fixes for #77 and #78 that you discovered, so thank you! Some of the changes with the new parser in `remark-parse` v9 don't seem to be compatible with this plugin's use case, particularly the code block indentation stuff. I don't yet know how non-trivial that work would be, but I'd welcome the research if someone wants to figure out. I think that's as as we can go for now, so I'll close this. Thanks for the push!

Issue #170, opened 2021-02-02
Add an option to specify different rules inside a markdown file


aduh95:
I'm trying to set up a rule set that is able to discriminate between ESM examples and CJS code snippets inside the markdown documentation:

````markdown
```js esm
export default {};
```
```js cjs
'use strict';
module.exports = {};
```
````

The issue is that only the `js` is passed through `unfied` `lang` data, and the part after the space is pass to the `meta` option:

```js
{
  type: 'html',
  lang: 'js',
  meta: 'esm',
  value: 'export default {};',
  position: {
    start: { line: 1, column: 1, offset: 0 },
    end: { line: 3, column: 4, offset: 32 }
  }
}
{
  type: 'html',
  lang: 'js',
  meta: 'cjs',
  value: "'use strict';\nmodule.exports = {};",
  position: {
    start: { line: 4, column: 1, offset: 33 },
    end: { line: 7, column: 4, offset: 81 }
  }
}
```

I wish there was a way to create separate rules for those blocks, something like:

```js
// .eslintrc.js
module.exports = {
    plugins: ["markdown"],
    overrides: [
        {
            files: ["**/*.md/*.js:cjs"],
            parserOption: { sourceType: "script" },
            rules: {
                // ...
            }
        },
        {
            files: ["**/*.md/*.js:esm"],
            parserOption: { sourceType: "module" },
            rules: {
                // ...
            }
        },
    ]
};
```

EDIT: I'm using `remark-parse` 9.0.0, the `meta` property doesn't seem to be available in `remark-parse` 5.x this repo is using.

btmills:
Are you able to use ```` ```cjs```` and ```` ```mjs```` fences instead of ```` ```js````? That would allow you to configure `**/*.md/*.cjs` and `**/*.md/*.mjs` separately.

aduh95:
Sure, but that doesn't address the underlining issue: I'd still want to have the option to specify different set of overrides for a same language. Ignoring those meta strings may be fine for some people, but I think it can be a very useful feature for others.

If we want to make a breaking change, I think now would be the right time as it's going to make folks rewrite their ESLint configuration anyway.
If such a breaking change is not acceptable, maybe we could make this configurable/opt-in?

btmills:
While I agree switching functionality based on the meta string is neat, there'd need to be some important use case for which that route is the only option. One downside is that it claims exclusive use of the meta string. What if there are other tools that want to use it as well?

[...]

PR #172, opened 2021-02-04
New: add support for meta info in code blocks

aduh95:
Fixes: #170
Blocked on #171


btmills:
If someone already has extra meta info after the opening fence, adding it to the virtual filename would be a breaking change. I've labeled this as such, though I'm optimistic #170 could be solved in a non-breaking way. Let's focus discussion there for now.

btmills:
Closing as the underlying issue has been solved another way!

PR #177, opened 2021-03-01
Fix: More reliable comment attachment (fixes #76)

### LINKED ISSUE #76: List entry without followed empty line eliminates configuration/skip comments
`eslint@4.1.1`, `eslint-plugin-markdown@1.0.0-beta.7`

Test doc 1:
<pre>
* List entry

&lt;!-- eslint-disable strict --&gt;
```js
console.log('a');
```

* List entry

&lt;!-- eslint-skip --&gt;
```js
> console.log('a');
```
</pre>
No errors.

Test doc 2:
<pre>
* List entry
&lt;!-- eslint-disable strict --&gt;
```js
console.log('a');
```

* List entry
&lt;!-- eslint-skip --&gt;
```js
> console.log('a');
```
</pre>

Comments are ignored:

```console
   4:1  error  Use the global form of 'use strict'     strict
  10:1  error  Parsing error: Unexpected token >
```
No other block or span Markdown elements have this issue. All list elements (unordered with `* + -` or ordered with `1. ...`) have the same issue (both one- and multiline).

btmills:
Comment attachment was sensitive to whitespace around the code block and preceding comments. In some cases, the parser would place comments as descendants of code blocks' preceding sibling nodes. However, a depth-first traversal of the tree will still encounter the comments in linear order, which is sufficient for our purposes. This approach no longer traverses backwards from code blocks looking for preceding comments, so we no longer need to pass the `parent` node and index into its children.

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -22,25 +22,27 @@ let blocks = [];
  * Performs a depth-first traversal of the Markdown AST.
  * @param {ASTNode} node A Markdown AST node.
  * @param {Object} callbacks A map of node types to callbacks.
- * @param {Object} [parent] The node's parent AST node.
  * @returns {void}
  */
-function traverse(node, callbacks, parent) {
+function traverse(node, callbacks) {
     if (callbacks[node.type]) {
-        callbacks[node.type](node, parent);
+        callbacks[node.type](node);
+    } else {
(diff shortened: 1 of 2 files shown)


nzakas (approved):
LGTM

PR #178, opened 2021-03-12
Update: compatibility with eslint-mdx

JounQin:
resolve #134, fix #176 and https://github.com/mdx-js/eslint-mdx/issues/259

close #180

related https://github.com/mdx-js/eslint-mdx/pull/283

I added a `getShortLang` helper, but I don't know if you'll agree. Personally I think it's useful for example I'm writing `JavaScript` language, but the file extension should be `.js`. Or, we can provide an option.

See also https://github.com/eslint/eslint-plugin-markdown/pull/178#discussion_r593347558

PR #182, opened 2021-03-18
Docs: improve jsdoc, better for typings

JounQin:
split part of #178

@btmills

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@ node_modules/
 npm-debug.log
 .eslint-release-info.json
 .nyc_output
-/yarn.lock
-/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ node_modules/
 npm-debug.log
 .eslint-release-info.json
 .nyc_output
+/yarn.lock
+/package-lock.json
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -1,10 +1,25 @@
 /**
  * @fileoverview Processes Markdown files for consumption by ESLint.
  * @author Brandon Mills
+ */
+
+/**
+ * @typedef {import('eslint/lib/shared/types').LintMessage} Message
  *
- * @typedef {import('unist').Node} ASTNode
- * @typedef {import('eslint').Linter.LintMessage} Message
- * @typedef {ASTNode & {lang: string, baseIndentText: string, comments: string[], rangeMap: Array<{js: number, md: number}>}} Block
+ * @typedef {Object} ASTNode
+ * @property {string} type
+ * @property {string} [lang]
+ *
+ * @typedef {Object} RangeMap
+ * @property {number} js
+ * @property {number} md
+ *
+ * @typedef {Object} BlockBase
+ * @property {string} baseIndentText
+ * @property {string[]} comments
+ * @property {RangeMap[]} rangeMap
+ *
+ * @typedef {ASTNode & BlockBase} Block
  */
 
 "use strict";
@@ -20,6 +35,9 @@ const SUPPORTS_AUTOFIX = true;
 
 const markdown = unified().use(remarkParse);
 
+/**
+ * @type {Block[]}
+ */
 let blocks = [];
 
 /**
@@ -124,7 +142,7 @@ function getIndentText(text, node) {
  * @param {ASTNode} node A Markdown code block AST node.
  * @param {string[]} comments List of configuration comment strings that will be
  *     inserted at the beginning of the code block.
- * @returns {Array<{js: number, md: number}>} A list of offset-based adjustments, where lookups are
+ * @returns {RangeMap[]} A list of offset-based adjustments, where lookups are
  *     done based on the `js` key, which represents the range in the linted JS,
  *     and the `md` key is the offset delta that, when added to the JS range,
  *     returns the corresponding location in the original Markdown source.
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -1,6 +1,10 @@
 /**
  * @fileoverview Processes Markdown files for consumption by ESLint.
  * @author Brandon Mills
+ *
+ * @typedef {import('unist').Node} ASTNode
+ * @typedef {import('eslint').Linter.LintMessage} Message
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@ node_modules/
 npm-debug.log
 .eslint-release-info.json
 .nyc_output
+/yarn.lock
+/package-lock.json
```
btmills, line 8: https://github.com/eslint/eslint-plugin-markdown/blob/610cffd93d29cc10bf9f2b51daeecb1b42338a34/.npmrc#L1

`package-lock.json` is disabled in `.npmrc`, so npm shouldn't be creating it. If someone does accidentally end up with a lock file (for either npm or yarn), I'd like that to show up in `git status` so that they notice it. Ensuring all installs use the latest compatible version will avoid one possible cause of weird hard-to-reproduce errors due to slightly different versions.

```suggestion
```

package.json
```diff
@@ -37,6 +37,8 @@
     "lib/processor.js"
   ],
   "devDependencies": {
+    "@types/eslint": "^7.2.7",
+    "@types/unist": "^2.0.3",
```
btmills, line 41: ESLint projects use JSDoc types without adopting TypeScript/Flow/etc-specific semantics, so I'd like to avoid installing TypeScript declarations from `@types/*`. If that requires re-defining JSDoc `@typedef`s for the `ASTNode` and `Message` types, that duplication is not a problem.

```suggestion
```

lib/processor.js
```diff
@@ -1,6 +1,11 @@
 /**
  * @fileoverview Processes Markdown files for consumption by ESLint.
  * @author Brandon Mills
+ *
+ * @typedef {import('unist').Node} ASTNode
+ * @typedef {ASTNode & {children: ASTNode[]}} ASTParentNode
+ * @typedef {import('eslint').Linter.LintMessage} Message
```
btmills, line 7: I have no idea how JSDoc imports work, but instead of `@types/eslint`, does it support importing the `LintMessage` `@typedef` provided by ESLint in [`lib/shared/types.js`](https://github.com/eslint/eslint/blob/cd2e584aa5a87c425177fb84a9cd6b05624fec43/lib/shared/types.js#L84-L96)? We don't strip comments from our builds, so those comments are all present in `node_modules` installs.

btmills (changes requested):

.gitignore
JounQin: It makes a bit difficult to add commits, I'm using `git add .` sometimes, so it work commit `yarn.lock` file accidentally. Did you mean that I should remove `yarn.lock` every time after installing? Or can I add a `.yarnrc` like `.npmrc`?

btmills: Repositories under the ESLint GitHub organization have historically used npm in development and to run tests and releases. That means we've already solved this friction point for npm by automatically disabling `package-lock.json`, and I suspect that we'd set a similar option for yarn if one existed. I typically use yarn for other projects but use npm when working on ESLint repositories out of habit so hadn't noticed the `git add .` pain point until you just pointed it out.

I just checked a couple other repositories in the organization and see that they've included [both package managers' lockfiles in `.gitignore`](
https://github.com/eslint/eslint/blob/cd2e584aa5a87c425177fb84a9cd6b05624fec43/.gitignore#L24-L25). Given that precedent, if you'd like to submit a standalone PR that adds those, I'd be happy to merge it 🙂

lib/processor.js
```diff
@@ -16,12 +37,15 @@ const SUPPORTS_AUTOFIX = true;
...
+/**
+ * @type {Block[]}
+ */
 let blocks = [];
 
 /**
  * Performs a depth-first traversal of the Markdown AST.
  * @param {ASTNode} node A Markdown AST node.
- * @param {Object} callbacks A map of node types to callbacks.
+ * @param {{[key: string]: (node: ASTNode, parent: ASTParentNode) => void}} callbacks A map of node types to callbacks.
```
btmills, line 48:
```suggestion
 * @param {{[key: string]: (node: ASTNode) => void}} callbacks A map of node types to callbacks.
```

lib/processor.js
```diff
@@ -3,6 +3,27 @@
...
 
+/**
+ * @typedef {import('eslint/lib/shared/types').LintMessage} Message
+ *
+ * @typedef {Object} ASTNode
+ * @property {string} type
+ * @property {string} [lang]
+ *
+ * @typedef {ASTNode & {children: ASTNode[]}} ASTParentNode
+ *
```
btmills, line 14: Since #177 removed the need for the parent node, can you confirm that we're good to get rid of the `ASTParentNode` type? If so, hopefully you can commit this and the other suggestion inline to save effort.

```suggestion
```

.gitignore
JounQin: > I suspect that we'd set a similar option for yarn if one existed

Unfortunately, there is and will be no option for yarn.

I'd like to raise a PR for it.

btmills (commented):
Thanks for making those changes! I'm glad you were able to create typedefs without pulling in TypeScript definitions.

I merged #177 earlier tonight, and I'm sorry that it caused a couple merge conflicts. I believe I was able to fix those with GitHub's built-in merge conflict editor, but please make sure I did that correctly. It also removed the parent node from the traversal step, so I left two inline suggestions to remove the `ASTParentNode` type that I'll let you apply after looking over them.

Once you've confirmed the merge conflict fixes, this'll be good to merge! Sorry again about the extra iteration from merge conflicts.

lib/processor.js
btmills: I should also note that #177 added a fallback `*` callback for any node types that don't have their own callback. Right now, that callback doesn't need access to the AST node, so it isn't passed in the `else` case. If that complicates the type definition, feel free to get around that by passing `node` in the `else` case as well if that's the easiest solution.

https://github.com/eslint/eslint-plugin-markdown/blob/79be776331cf2bb4db2f265ee6cf7260e90e3d5e/lib/processor.js#L28-L32

https://github.com/eslint/eslint-plugin-markdown/blob/79be776331cf2bb4db2f265ee6cf7260e90e3d5e/lib/processor.js#L233-L235

JounQin:
@btmills I've rebased just now.

btmills (approved):
LGTM, thanks @JounQin!

PR #175, opened 2021-02-27
Update: Upgrade remark-parse to v7 (fixes #77, fixes #78)

btmills:
[`remark-parse@7.0.0`](https://github.com/remarkjs/remark/releases/tag/remark-parse%407.0.0) includes https://github.com/remarkjs/remark/pull/423, which fixes handling of leading and trailing newlines in fenced code blocks.

**For reviewers:**

1. `remark-parse` v6, v7, and v8 were semver-major fixes. v9 was a full rewrite. Are we comfortable with a parser major version bump in a semver-minor release? This is currently tagged as `Update:` so it's semver-minor rather than semver-patch.
2. If so, how far? This PR bumps to v7 because it fixes two known issues, #77 and #78. I'm not aware of any `eslint-plugin-markdown` issues that would be fixed by v8 or v9.
3. If we're fine upgrading past v7 for speculative parsing accuracy improvements that don't knowingly affect `eslint-plugin-markdown`, should we stop at v8 or go all the way to the re-written parser in v9? If the latter, I'll close this in favor of #171.

**Changelogs:**

- [`remark-parse@6.0.0`](https://github.com/remarkjs/remark/releases/tag/remark-parse%406.0.0)
- [`remark-parse@7.0.0`](https://github.com/remarkjs/remark/releases/tag/remark-parse%407.0.0)
- [`remark-parse@8.0.0`](https://github.com/remarkjs/remark/releases/tag/remark-parse%408.0.0)
- [`remark-parse@9.9.0`](https://github.com/remarkjs/remark/releases/tag/13.0.0)

PR #183, opened 2021-03-18
Fix: use blocksCache instead of single blocks instance (fixes #181)

### LINKED ISSUE #181: [bug] `blocks` will be reset on `preprocess` every time and will break `.md` in `.md`
The  will not work when writing markdown codes in code blocks.

Lifecycle:

```
preprocess('.md') -> preprocess('.md/0_0.md') -> postprocess('.md/0_0.md') -> postprocess('.md')
```

This was not a problem before because `.md` files are not linted by eslint before. By adding `esling-plugin-mdx` which support `.md` files, it will be a problem.

_Originally posted by @JounQin in https://github.com/eslint/eslint-plugin-markdown/pull/178#discussion_r593347558_

---

@btmills 

It's clear to be an bug

https://github.com/eslint/eslint-plugin-markdown/blob/main/lib/processor.js#L219

JounQin:
pass through lint messages from other plugins

close #181

---

split part of #178

@btmills


JounQin:
This issue will only occur when integrating with other plugins like `eslint-plugin-mdx` which will lint original `.md` files by `preprocess` too.

Here is the fake version for testing.

https://github.com/eslint/eslint-plugin-markdown/pull/183/files#diff-83e0a2056017ee834945f51308072e4a883a4ed96193581fe4aabc41cb98710a

btmills:
Thanks to your discussion in #181 and https://github.com/eslint/eslint/pull/14227/files#r602802758, we can reproduce this in a test without stubbing another mdx plugin! So let's simplify this:

<pre><code>&lt;!-- test.md --&gt;

````markdown
&lt;!-- test.md/0_0.markdown --&gt;

```js
// test.md/0_0.markdown/0_0.js
console.log("Hello");
```
````
</code></pre>

With that as the test input, if you configure the CLIEngine to run this plugin's processor on `.md` _and_ `.markdown` files (the `recommended` config only does `.md`), ESLint should run the processor recursively. That should reproduce the bug without having to stub another plugin and reduce the number of moving parts in this PR.

If that doesn't work, mention me and I'll see if I can put together a reproduction case that you can use to prove this implementation works.

I'm glad you were able to fix this by keying the blocks cache on the filename. There are two implementation tweaks that I'd like to request:

1. Can you store the cache in a `Map` instead of an object? That's supported all the way back to Node 8.10.
2. To save memory, would it be possible to remove blocks from the cache once `postprocess()` is done with them?

**Comment by JounQin** [comment]
@btmills All done!

File: package.json [hunk 1]
```diff
@@ -47,8 +47,8 @@
     "nyc": "^14.1.1"
   },
   "dependencies": {
-    "remark-parse": "^5.0.0",
-    "unified": "^6.1.2"
+    "remark-parse": ">=5.0.0 <9.0.0",
+    "unified": ">=6.0.0"
```
btmills, line 51: Are these version changes a necessary part of this PR? If the PR works without changing them, we'll be able to merge this as a semver-patch bugfix and release it as v2.0.1.

---

For more context in case you're interested, I've opened #175 to upgrade the parser to v7. At the very least that'll be a semver-minor change, so v2.1.0. I'm waiting for confirmation from the team that we're comfortable doing it as only a minor version bump. There's some risk that a parser change in one of its major versions will cause this processor to behave differently in a way that would require us to do a major bump to v3.0.0 here as well. Since v6, v7, and v8 are just bug fixes, I'm less concerned about noticeable breakage there. v7 fixes a bug in this processor, so to reduce risk in #175, I only went as far as needed to fix the processor bugs.

examples/typescript/README.md
```diff
@@ -1,4 +1,4 @@
-# React Example
+# TypeScript Example
```
btmills, line 1: Ha, whoops 🤦  thanks for the quick fix.

lib/processor.js
```diff
@@ -370,14 +373,28 @@ function excludeUnsatisfiableRules(message) {
...
-    return [].concat(...messages.map((group, i) => {
-        const adjust = adjustBlock(blocks[i]);
+function postprocess(messages, filename) {
+    const blocks = blocksCache.get(filename) || [];
+
+    const filteredMessages = [].concat(...messages.map((group, i) => {
+        const block = blocks[i];
+
+        // non code block message, parsed by `eslint-mdx` for example
+        if (!block) {
```
btmills, line 386: Can you elaborate on what scenarios could cause a cache miss above (requiring the `|| []` fallback) or `block` to be missing here? At least so far in this processor's history, those would both indicate bugs where there's a mismatch between the logic in `preprocess()` and `postprocess()`. If that's still the case, I'd like to fail as close to the source as possible so that we get the most accurate stack traces and bug reports. While the fallbacks are nice in that they probably prevent a crash, they could lead to undefined behavior that shows incorrect results.

JounQin: It's about integration with `eslint-plugin-mdx`, not a bug for `eslint-plugin-markdown`, I want to reuse `eslint-plugin-markdown`'s processor in `eslint-plugin-mdx`.

See https://github.com/mdx-js/eslint-mdx/blob/master/packages/eslint-plugin-mdx/src/processors/remark.ts#L29

package.json
JounQin: OK, it will revert this change.

lib/processor.js
btmills: Ah, I see that `eslint-plugin-mdx` appends one additional virtual file from `preprocess()`. In `eslint-plugin-mdx`'s `postprocess()`, could you then trim the last `Message[]` from `lintMessages`? I think that would ensure that `eslint-plugin-markdown`'s `postprocess()` here would only get messages that came from its own `preprocess()`, and these fallbacks wouldn't be needed anymore.

As-is, these fallbacks are implementation details around error handling in the `postprocess()` method. They're not part of the public API mandated by ESLint. A refactor of this logic could hypothetically change this behavior in a way that's fully compatible with the public API yet unintentionally invalidates an assumption that `eslint-plugin-mdx` was making. That's why I'd prefer to have the "extra block" behavior fully owned and controlled within `eslint-plugin-mdx` if at all possible.

JounQin: Thanks for your great suggestion! I'll refactor it ASAP.

JounQin: @btmills Done!

btmills (approved):
Thanks for working so hard to get this down to just the essential components! I'm really happy with where it ended up. LGTM.

JounQin:
@btmills Any schedule for a release then?

btmills:
Now! https://github.com/eslint/eslint-plugin-markdown/releases/tag/v2.0.1

PR #185, opened 2021-03-30
Chore: Reproduce nested blocks failure (refs #181)

btmills:
@JounQin I put together a reproduction test tonight to make sure I fully understand the problem, and I'm sharing it here in case it helps as you're working on #183.

### DIFF
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -12,19 +12,22 @@ const path = require("path");
 const plugin = require("../..");
 
 /**
+ * @typedef {import('eslint/lib/cli-engine/cli-engine').CLIEngineOptions} CLIEngineOptions
+ */
+
+/**
  * Helper function which creates CLIEngine instance with enabled/disabled autofix feature.
  * @param {string} fixtureConfigName ESLint JSON config fixture filename.
- * @param {boolean} [isAutofixEnabled=false] Whether to enable autofix feature.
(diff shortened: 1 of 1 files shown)


JounQin:
Thanks, I was thinking that has been validated via https://github.com/eslint/eslint-plugin-markdown/pull/183/files#diff-8c591ed3533957b84298776fda1d6d11b2da563b3222e471b4725fe248c0e294

I'll add this test case in that PR.

btmills:
That fixture was the right `.md` to reproduce the issue, but by itself it wasn't being executed because no `describe()`/`it()` test was using it. If you were to add that fixture by itself without implementing the blocks cache fix yet and run the tests, it wouldn't fail as hoped.

Here I wrote a `describe()`/`it()` test that will be executed as part of the test run to prove that 1) `preprocess` is extracting nested blocks, 2) ESLint is finding the expected lint messages in those nested blocks, and 3) the two layers of `postprocess` calls map those lint messages back to the correct line in `test.md`.

The test would function just as well using a file in `tests/fixtures/test.md` instead of creating the file text with `[...].join("\n");`, but I personally find it easier to see what the test is doing when the file text and assertions are right next to each other.

JounQin:
@btmills No worry, I've added this test case at #181 as you suggested.

btmills:
Closing as this has been integrated into #183, which contains the fix.

PR #192, opened 2021-09-29
Chore: use `actions/setup-node@v2` in CI

snitin315:
CI-related change.

updated setup-node action to v2 for CI node workflow.

https://github.com/marketplace/actions/setup-node-js-environment

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v1
     - name: Install Node.js
-      uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


btmills (approved):
Neat, thanks! LGTM as-is, but since this version supports caching, do you have any interest in enabling that too?

snitin315:
I think we need a `package-lock.json` to enable caching, and the repository doesn't have one. There was a similar attempt in eslint repo for this in https://github.com/eslint/eslint/pull/14843

btmills:
> I think we need a package-lock.json to enable caching, and the repository doesn't have one.

Ah, never mind then.

PR #198, opened 2021-11-23
test: test with `ESLint` instead of `CLIEngine` when available

MichaelDeBoey:
As asked by @btmills in https://github.com/eslint/eslint-plugin-markdown/pull/195#issuecomment-955802046

### DIFF
--- a/tests/lib/plugin.js
+++ b/tests/lib/plugin.js
@@ -7,7 +7,7 @@
 
 const assert = require("chai").assert;
 const execSync = require("child_process").execSync;
-const CLIEngine = require("eslint").CLIEngine;
+const { CLIEngine, ESLint } = require("eslint");
 const path = require("path");
 const plugin = require("../..");
 
@@ -19,9 +19,20 @@ const plugin = require("../..");
  * Helper function which creates CLIEngine instance with enabled/disabled autofix feature.
  * @param {string} fixtureConfigName ESLint JSON config fixture filename.
  * @param {CLIEngineOptions} [options={}] Whether to enable autofix feature.
- * @returns {CLIEngine} CLIEngine instance to execute in tests.
+ * @returns {ESLint} ESLint instance to execute in tests.
  */
-function initCLI(fixtureConfigName, options = {}) {
+function initESLint(fixtureConfigName, options = {}) {
+    if (ESLint) { // ESLint v7+
+        return new ESLint({
+            cwd: path.resolve(__dirname, "../fixtures/"),
+            ignore: false,
+            useEslintrc: false,
+            overrideConfigFile: path.resolve(__dirname, "../fixtures/", fixtureConfigName),
+            plugins: { markdown: plugin },
+            ...options
+        });
+    }
+
     const cli = new CLIEngine({
         cwd: path.resolve(__dirname, "../fixtures/"),
         ignore: false,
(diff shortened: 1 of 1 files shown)


nzakas:
Tests are failing

tests/lib/plugin.js
```diff
@@ -19,19 +19,27 @@ const plugin = require("../..");
...
+    return new CLIEngine({
         cwd: path.resolve(__dirname, "../fixtures/"),
         ignore: false,
         useEslintrc: false,
         configFile: path.resolve(__dirname, "../fixtures/", fixtureConfigName),
         ...options
-    });
-
-    cli.addPlugin("markdown", plugin);
-    return cli;
```
btmills, line 34: I suspect tests are failing because these lines were changed. (`addPlugin()` [returns `void`](https://github.com/eslint/eslint/blob/9738f8cc864d769988ccf42bb70f524444df1349/lib/cli-engine/cli-engine.js#L675-L681), not the `CLIEngine` instance.)

btmills: Upon further investigation, looks like this is only part of it. In addition to having different method names than `CLIEngine`, `ESLint`'s methods are also `async` and return [`LintResult[]`](https://eslint.org/docs/developer-guide/nodejs-api#return-value-1) instead of a `report` object where `.results` is the array. To avoid that repetitive ternary, it should be possible to return a wrapper that adapts the `CLIEngine` instance to match the `ESLint` API for the limited purposes of our tests:

```js
return {
    async calculateConfigForFile(filename) {
        return cli.getConfigForFile(filename);
    },
    async lintFiles(files) {
        return cli.executeOnFiles(files).results;
    },
    async lintText(text, { filePath }) {
        return cli.executeOnText(text, filePath).results;
    }
};
```

Then all the tests can use the `ESLint` API but still work with `CLIEngine` in `eslint@6`:

```js
it("description", async () => {
    const results = await eslint.lintText(text, { filePath: "test.md" });
    assert.strictEqual(results.length, 1);
    // ...
});
```

- Renamed `initCli()` to `initESLint` and its return value `cli` to `eslint`.
- Used `ESLint` methods like `lintText()` instead of `CLIEngine` methods like `executeOnText()`, no ternary needed thanks to the wrapper.

tests/lib/plugin.js
```diff
@@ -192,76 +211,76 @@ describe("plugin", () => {
...
+        assert.strictEqual(results[0].messages[0].endLine, 7);
+        assert.strictEqual(results[0].messages[1].message, "'blah' is not defined.");
+        assert.strictEqual(results[0].messages[1].line, 11);
+        assert.strictEqual(results[0].messages[1].endLine, 11);
     });
 
-    it("should run on .mkdn files", () => {
-        const report = cli.executeOnText(shortText, "test.mkdn");
+    it("should run on .mkdn files", async () => {
+        const results = await eslint.lintText(shortText, { filePath: "test.mkdnd" });
```
btmills, line 225: Hopefully the last test failure is just due to a typo here from the API conversion 🤞

```suggestion
        const results = await eslint.lintText(shortText, { filePath: "test.mkdn" });
```

btmills (changes requested):
Ideally we're one typo fix away from all tests passing 🤞 If that's true, LGTM once that change is made!

MichaelDeBoey:
@btmills All tests are green now 🥳🎉

btmills (approved):
Woohoo, no implementation changes needed to work with the `ESLint` API! Thanks for proving that in these tests.

nzakas:
Thanks @MichaelDeBoey! We selected you to receive payment for your work, please send an email to contact@eslint.org so we can send you the info.

PR #203, opened 2022-05-21
feat!: drop node v8 and v10

amareshsm:
**Description:** 

Drop node v8.x and v10.x


nzakas:
@btmills thoughts?

btmills:
**TSC summary**: `eslint-plugin-markdown` currently tests against [ESLint 6+ and Node 8+](https://github.com/eslint/eslint-plugin-markdown/blob/87c2b536fd80b15e134766d92b90048ae45cbe1f/.github/workflows/ci.yml#L32-L33). This pull request would drop support for end-of-life versions of Node.js. Normally, we drop support for EOL engines as part of regular major releases. However, we have no other breaking changes queued up for this package.

**TSC question**: Should we do a semver-major release just to drop support for EOL Node.js versions?

btmills (changes requested):
Thank you, @amareshsm! The team had no objections to doing a v3 with dropping EOL versions as the only breaking change.

It looks like there are some compat issues with Node 18. If necessary, feel free to drop 8/10/<12.22.0 here and split adding 18 support into another PR.

amareshsm:
Drafted a separate PR for node v18 support.

.github/workflows/ci.yml
```diff
@@ -30,18 +30,7 @@ jobs:
...
-            node: 10
-          - eslint: 8
-            node: 8
-          - eslint: 8
-            node: 8.10.0
-          - eslint: 7
-            node: 8
-          - eslint: 7
-            node: 8.10.0
+        node: [12, 14, 16]
```
btmills, line 33: To ensure it still works against the absolute lowest supported version:

```suggestion
        node: [12.22.0, 12, 14, 16]
```

(Previously we've discussed testing against the lowest supported patch of _each_ major Node version. We decided that we'd only add e.g. `16.0.0` in addition to `16` if we accidentally broke `16.0.0` but didn't catch it because it still passed in `16`.)

btmills (changes requested):
One inline change, then LGTM!

.github/workflows/ci.yml
amareshsm: done

amareshsm:
Bump

btmills:
This and your other PR appear to be failing due to behavior changes in new npm versions that were released since the previous merges to `main`. Nothing about your change caused these failures. I've submitted https://github.com/eslint/eslint-plugin-markdown/pull/206 to fix the changes on `main`. Once that's in, I bet rebasing both of your branches will let them pass :+1:

btmills:
Closing and reopening to see if that'll fix CI without having to rebase now that #206 is merged.

btmills (approved):
That's more like it! Thanks for your patience, @amareshsm. LGTM, but since this is a breaking change, I'll leave it open another couple days to give others a chance to review.

nzakas (approved):
LGTM

PR #206, opened 2022-06-19
ci: Prevent recursive prepare script invocations

btmills:
A change between npm 8.9 and 8.10 started running the main package's `prepare` script as part of the examples' installs. This broke CI when the install step launched a fork bomb and got killed after several minutes.

I've left a comment on https://github.com/npm/cli/issues/4895, which appears to be related.

This change should fix builds on `main`.

### DIFF
--- a/npm-prepare.js
+++ b/npm-prepare.js
@@ -6,17 +6,23 @@
 
 "use strict";
 
-const childProcess = require("child_process");
-const fs = require("fs");
-const path = require("path");
+if (!process.env.NO_RECURSIVE_PREPARE) {
+    const childProcess = require("child_process");
+    const fs = require("fs");
+    const path = require("path");
 
-const examplesDir = path.resolve(__dirname, "examples");
-const examples = fs.readdirSync(examplesDir)
-    .filter(exampleDir => fs.statSync(path.join(examplesDir, exampleDir)).isDirectory())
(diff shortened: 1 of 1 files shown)


btmills:
Looks like there are _two_ behavior changes in npm since 8.5 that are breaking our CI. I'm not yet confident enough that our usage is intended to be supported to call them regressions or breaking changes, so I'm sticking with behavior changes for now.

I have a workaround for the second change that is still causing CI to fail on this branch. (Spoiler alert: peer deps, or course.) I needed to step away for a bit but will push the fix later.

tests/lib/plugin.js
```diff
@@ -82,7 +82,7 @@ describe("recommended config", () => {
                 // eslint-disable-next-line no-invalid-this
                 this.timeout(30000);
 
-                execSync("npm link && npm link eslint-plugin-markdown");
+                execSync("npm link && npm link eslint-plugin-markdown --legacy-peer-deps");
```
btmills, line 85: We have to use `--legacy-peer-deps` because there is no version of `eslint-plugin-jsdoc` that supports ESLint 6, 7, and 8. `--legacy-peer-deps` disables additional peer dependency checks that were added in npm 8.6.0 and caused our CI to begin failing.

snitin315 (approved):
LGTM.

nzakas (approved):

PR #210, opened 2022-09-10
feat: support new config system

jjangga0214:
Supporting eslint's new config system of eslint.

The legacy config system always has `require()`d plugins and sharable configs, while the new system is available in both ESM and CJS.

Therefore [conditional export](https://nodejs.org/api/packages.html#conditional-exports) is great to keep compatibility while introducing the new config.

`plugin:` protocol(e.g. `plugin:markdown/recommended`) is not valid any more. In the new config system, a bi-directional dependency between a plugin object and a preset config object is detached. From now on, only a config object should depend on a plugin object, not vice-versa. Thus, the new plugin (a new entry point index.js) doesn't have `configs` property, following the spec, though having the property does not mean incompatibility.

Users should have a way to import shareable config independently.
Hence `eslint-plugin-markdown/recommended` is introduced as a new entry point for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,14 +218,12 @@ You might want to check out the official blog posts,
 
 and the [official docs](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new).
 
-If your `eslint.config.js` is CJS, change it to ESM.
-
 ### Shareable config
 
 Extending the the `recommended` config will enable the Markdown processor on all .md files.
 
 ```ts
-import markdown from 'eslint-plugin-markdown/recommended'
+import markdown from 'eslint-plugin-markdown/recommended' // <== trailing '/recommended'
 
 export default [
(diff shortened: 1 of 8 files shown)


jjangga0214:
I also wrote README. And I've taken care to make sure this PR is not a breaking change.

jjangga0214:
@nzakas May I ping? 
I want to get feedback from you about the design in this PR, as I also want to apply similar changes to other plugins.

kecrily:
Can you add a unit test for new system config?

kecrily (commented):
Can you add a unit test for new config system?

jjangga0214:
@kecrily Maybe later?
Or I expect whoever is available can do that before I do.

nzakas:
Superseded by #232

PR #229, opened 2023-12-01
feat!: bump `mdast-util-from-markdown` with `synckit`

### LINKED ISSUE #218: request: update dependencies
In my nuxt project this plugin is causing conflicts with `@nuxtjs/content`.
Updating this dependency should fix the issue
https://github.com/syntax-tree/mdast-util-from-markdown

JounQin:
close #218, drop node 12

`mdast-util-from-markdown` is ESM only now, [`synckit`](https://github.com/un-ts/synckit) makes it possible to require ESM only module in commonjs without `await import()`.

cc @btmills @nzakas

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -3,8 +3,19 @@
 module.exports = [
     ...require("eslint-config-eslint/cjs").map(config => ({
         ...config,
-        files: ["**/*.js"]
+        files: ["**/*.js", "**/*.mjs"]
     })),
+    {
+        files: ["**/*.mjs"],
+        languageOptions: {
+            sourceType: "module",
+            parserOptions: {
+                ecmaFeatures: {
+                    impliedStrict: true
+                }
+            }
+        }
+    },
     {
         plugins: {
             markdown: require(".")
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -27,8 +27,11 @@
 
 "use strict";
 
-const parse = require("mdast-util-from-markdown");
+const { createSyncFn } = require("synckit");
+
 const pkg = require("../package.json");
+
+const parse = createSyncFn(require.resolve("./worker.mjs"));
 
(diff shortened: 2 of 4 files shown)


nzakas:
We don't want to drop Node 12 until we release ESLint v9, which will also drop Node.js 12. 

Is there a reason we can't use mdast-util-from-markdown 0.8.5 for now?

JounQin:
It requires outdated unified related packages may results conflict/duplicate dependencies.

nzakas:
Okay, We'll still need to wait until ESLint v9 to drop Node.js 12 support. At that point, we need to update Node.js support for all packages.

JounQin:
@nzakas 

image

`synckit@v0.7` supports Node 12, would that be acceptable? I can patch Node v12 support back on v0.8 either.

JounQin:
Friendly ping @nzakas

nzakas:
We just released ESLint v9.0.0-alpha, so I think we can formally update the Node.js support for this package now. 

Can you submit a PR like this? https://github.com/eslint/eslint-scope/pull/115

JounQin:
Wow, that's really a big change, I was thinking to keep compatible, not ESLint v9 only. Because in case the big breaking changes on eslint v9, I believe the migration could be very slow.

JounQin:
@nzakas Any further discussion?

aladdin-add:
@JounQin it has dropped node.js < 18 supports. what's the next steps here?

JounQin:
@aladdin-add Didn't notice that, I've updated the PR accordingly.

remcohaszing (approved):

nzakas:
At this point, I don't think we should move further with this PR. It's not fixing any bugs and adding in a new dependency just to upgrade some other module doesn't seem like a good tradeoff.

PR #247, opened 2024-03-27
fix: check upper bounds of message line numbers for code blocks

### LINKED ISSUE #244: Bug: Uncaught undefined for reporting after a code block
### Issue
Currently there is a [guard](https://github.com/eslint/eslint-plugin-markdown/blob/bb5c3d4d9b0283bbb44fd75d1c04a3ec7e789fe1/lib/processor.js#L334) for messages which report before the code block, but not if the message reports for a line after the code block.
This leads to an error in the [block.rangeMap[lineInCode].indent](https://github.com/eslint/eslint-plugin-markdown/blob/bb5c3d4d9b0283bbb44fd75d1c04a3ec7e789fe1/lib/processor.js#L340) access,
as block.rangeMap[lineInCode] evaluates to undefined.
I am willing to create an PR for this.

### Minimal Reproduction
Markdown (I had to create an invalid document as the markdown renderer does not support nested code blocks):
```markdown
Empty code block:
`` `js
`` `
```

Rule:
```javascript
export default {
	create(context) {
		return {
			Program() {
				context.report({ message: 'message', loc: { start: { line: 2, column: 1 } } })
			}
		};
	}
}
```

DMartens:
Fixes #244.
Added an upper guard to make sure the access to `block.rangeMap[lineNumber]` is defined as well as add a simple test case (I manually set the state via `preprocess` in the test case and not use the "shared" state from the beforeEach to make this test case clearer).

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -331,7 +331,7 @@ function adjustBlock(block) {
 
         const lineInCode = message.line - leadingCommentLines;
 
-        if (lineInCode < 1) {
+        if (lineInCode < 1 || lineInCode >= block.rangeMap.length) {
             return null;
         }
 
--- a/tests/lib/processor.js
+++ b/tests/lib/processor.js
@@ -755,6 +755,17 @@ describe("processor", () => {
             });
         });
 
+        it("should ignore messages after the code block", () => {
+            const empty = [
+                '```javascript',
+                '```',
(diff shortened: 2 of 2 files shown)


nzakas (changes requested):
Code changes look good to me. Can you check the linting error?

PR #256, opened 2024-07-04
feat: add name to flat configs

ntnyq:
Adding names to configs makes them easier to check with @eslint/config-inspector.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -70,15 +70,18 @@ const plugin = {
 
 plugin.configs.recommended = [
     {
+        name: "markdown/recommended/plugin",
         plugins: {
             markdown: plugin
         }
     },
     {
+        name: "markdown/recommended/processor",
         files: ["**/*.md"],
         processor: "markdown/markdown"
     },
     {
+        name: "markdown/recommended/code-blocks",
         files: ["**/*.md/**"],
         languageOptions: {
             parserOptions: {


lib/index.js
```diff
@@ -70,15 +70,18 @@ const plugin = {
...
             markdown: plugin
         }
     },
     {
+        name: "markdown/processor",
         files: ["**/*.md"],
         processor: "markdown/markdown"
     },
     {
+        name: "markdown/rules/recommended",
```
nzakas, line 84:
```suggestion
        name: "markdown/recommended",
```

nzakas, line 84:
```suggestion
        name: "markdown/recommended",
```

nzakas: Other plugins just use "recommended" for this name.

ntnyq: > Other plugins just use "recommended" for this name.

Resolved.

lib/index.js
```diff
@@ -70,15 +70,18 @@ const plugin = {
...
             markdown: plugin
         }
     },
     {
+        name: "markdown/processor",
         files: ["**/*.md"],
         processor: "markdown/markdown"
     },
     {
+        name: "markdown/recommended",
```
nzakas, line 84:
```suggestion
        name: "markdown/recommended/code-blocks",
```

lib/index.js
```diff
@@ -70,15 +70,18 @@ const plugin = {
 
 plugin.configs.recommended = [
     {
+        name: "markdown/plugin",
         plugins: {
             markdown: plugin
         }
     },
     {
+        name: "markdown/processor",
```
nzakas, line 79:
```suggestion
        name: "markdown/recommended/processor",
```

lib/index.js
```diff
@@ -70,15 +70,18 @@ const plugin = {
 
 plugin.configs.recommended = [
     {
+        name: "markdown/plugin",
```
nzakas, line 73:
```suggestion
        name: "markdown/recommended/plugin",
```

nzakas (commented):
Okay, after thinking about this a bit, I think we should put "recommended" in each name.

nzakas (approved):
Thanks!

PR #258, opened 2024-07-16
build: Update release-please

nzakas:
This upgrades release-please to v4 and switches to the official API.

I've also updated to use a manifest file.

Last, I switched to static values for the name and version in the JavaScript files. release-please can update the version number directly during a release, which removes the dependency on `package.json`. That's important for converting to ESM, which is my next step.


mdjermanovic (approved):
LGTM, thanks!

Issue #325, opened 2025-03-25
Change Request: Support Front Matter


lumirlumir:
### Environment

ESLint version: `v9.23.0`
@eslint/markdown version: `v6.3.0`
Node version: `v20.18.0`
npm version: `v10.9.2`
Operating System: Windows 11

### What problem do you want to solve?

Hello,  

[Front Matter](https://jekyllrb.com/docs/front-matter/) is widely used in Markdown and is compatible with both CommonMark and GFM.  

Additionally, with [mdast-util-frontmatter](https://github.com/syntax-tree/mdast-util-frontmatter) available, integrating it into `@eslint/markdown` should be straightforward.  

Therefore, I’d like to propose adding support for Front Matter syntax in `@eslint/markdown`.

### What do you think is the correct solution?

Let's use [mdast-util-frontmatter](https://github.com/syntax-tree/mdast-util-frontmatter).  

We can add support for the language `markdown/commonmark/front-matter` or `markdown/commonmark/gfm` (the names are flexible!).

### Participation

- [x] I am willing to submit a pull request for this change.

### Additional comments

[markdownlint](https://github.com/DavidAnson/markdownlint) also supports Front Matter.

nzakas:
This is something I've been thinking about for a while, so yes, let's do it!

We should implement this as a new option in `languageOptions` (`frontmatter: true`) that applies in both CommonMark and GFM languages rather than a new language.

Note that we need to use both `mdast-util-frontmatter` and [`micromark-extension-frontmatter`](https://github.com/micromark/micromark-extension-frontmatter) together to get the correct effect.

PR #329, opened 2025-03-28
test: create tests for `MarkdownLanguage`

lumirlumir:
Hello,  

I've created basic test cases for `MarkdownLanguage`. This test was necessary to implement #325 correctly but was missing. I referenced tests for [JSON](https://github.com/eslint/json/blob/main/tests/languages/json-language.test.js) and [CSS](https://github.com/eslint/css/blob/main/tests/languages/css-language.test.js).  

I tried to create a failure case for the `parse()` method, but unlike JSON and CSS parsers, every string value is recognized as text. So, I couldn't make one fail. I'd appreciate it if you could suggest a possible failure case.

### DIFF
--- a/tests/language/markdown-language.test.js
+++ b/tests/language/markdown-language.test.js
@@ -38,11 +38,7 @@ describe("MarkdownLanguage", () => {
 
 			assert.strictEqual(result.ok, true);
 			// The table should not be parsed and should be recognized as plain text.
-			assert.strictEqual(result.ast.children[0].type === "table", false);
-			assert.strictEqual(
-				result.ast.children[0].type === "paragraph",
-				true,
-			);
+			assert.strictEqual(result.ast.children[0].type, "paragraph");
 		});
 
 		it("should parse gfm features in gfm mode", () => {
--- a/tests/language/markdown-language.test.js
+++ b/tests/language/markdown-language.test.js
@@ -50,11 +50,7 @@ describe("MarkdownLanguage", () => {
 
 			assert.strictEqual(result.ok, true);
 			// The table should be parsed correctly.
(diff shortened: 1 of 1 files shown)


tests/language/markdown-language.test.js
```diff
@@ -0,0 +1,108 @@
...
+				path: "test.md",
+			});
+
+			assert.strictEqual(result.ok, true);
+			// The table should not be parsed and should be recognized as plain text.
+			assert.strictEqual(result.ast.children[0].type === "table", false);
+			assert.strictEqual(
+				result.ast.children[0].type === "paragraph",
+				true,
+			);
```
nzakas, line 45: We only need the second assertion here.
```suggestion
			assert.strictEqual(result.ast.children[0].type, "paragraph");
```

tests/language/markdown-language.test.js
```diff
@@ -0,0 +1,108 @@
...
+				path: "test.md",
+			});
+
+			assert.strictEqual(result.ok, true);
+			// The table should be parsed correctly.
+			assert.strictEqual(result.ast.children[0].type === "table", true);
+			assert.strictEqual(
+				result.ast.children[0].type === "paragraph",
+				false,
+			);
```
nzakas, line 61: Same here. We just need the first assertion.
```suggestion
			assert.strictEqual(result.ast.children[0].type, "table");
```

nzakas (commented):
Thanks for putting this together. I just left a couple suggestions for cleaning things up a bit.

lumirlumir:
I've resolved the suggestions. Thank you!

nzakas (approved):
LGTM. Thanks!

PR #367, opened 2025-05-09
refactor: align rule typedefs with conventions from other plugins

lumirlumir:
#### Prerequisites checklist

Hello,

In this PR, I've aligned rule type definitions with conventions from other plugins like [CSS](https://github.com/eslint/css/blob/main/src/rules/use-layers.js) and [JSON](https://github.com/eslint/json/blob/main/src/rules/no-duplicate-keys.js)

I've used `@import` syntax and added support for `MessageIds`.

### DIFF
--- a/src/rules/fenced-code-language.js
+++ b/src/rules/fenced-code-language.js
@@ -8,8 +8,9 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ required?: string[]; }]; }>}
- * FencedCodeLanguageRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"missingLanguage" | "disallowedLanguage"} FencedCodeLanguageMessageIds
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ required?: string[]; }], MessageIds: FencedCodeLanguageMessageIds }>} FencedCodeLanguageRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
--- a/src/rules/heading-increment.js
+++ b/src/rules/heading-increment.js
@@ -8,8 +8,9 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
- * HeadingIncrementRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"skippedHeading"} HeadingIncrementMessageIds
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [], MessageIds: HeadingIncrementMessageIds }>} HeadingIncrementRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
--- a/src/rules/no-duplicate-headings.js
+++ b/src/rules/no-duplicate-headings.js
@@ -8,8 +8,9 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
- * NoDuplicateHeadingsRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"duplicateHeading"} NoDuplicateHeadingsMessageIds
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [], MessageIds: NoDuplicateHeadingsMessageIds }>} NoDuplicateHeadingsRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
--- a/src/rules/no-empty-images.js
+++ b/src/rules/no-empty-images.js
@@ -8,8 +8,9 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
- * NoEmptyImagesRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"emptyImage"} NoEmptyImagesMessageIds
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [], MessageIds: NoEmptyImagesMessageIds }>} NoEmptyImagesRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
--- a/src/rules/no-empty-links.js
+++ b/src/rules/no-empty-links.js
@@ -7,8 +7,9 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
- * NoEmptyLinksRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"emptyLink"} NoEmptyLinksMessageIds
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [], MessageIds: NoEmptyLinksMessageIds }>} NoEmptyLinksRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
--- a/src/rules/no-html.js
+++ b/src/rules/no-html.js
@@ -14,8 +14,9 @@ import { findOffsets } from "../util.js";
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ allowed?: string[]; }]; }>}
- * NoHtmlRuleDefinition
+ * @import { MarkdownRuleDefinition } from "../types.ts";
+ * @typedef {"disallowedElement"} NoHtmlMessageIds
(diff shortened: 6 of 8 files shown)


tools/dedupe-types.js
lumirlumir, line 1: I've followed the script from [JSON](https://github.com/eslint/json/blob/main/tools/dedupe-types.js).

tests/types/types.test.ts
```diff
@@ -1,14 +1,12 @@
...
+	Toml,
 	MarkdownRuleDefinition,
 	MarkdownRuleVisitor,
-	SourceLocation,
-	SourceRange,
-	type RuleModule,
-} from "@eslint/markdown";
-import { Toml } from "@eslint/markdown/types";
+	RuleModule,
+} from "@eslint/markdown/types";
```
lumirlumir, line 7: Would this be considered a BREAKING CHANGE?

The type module path was changed from the original to match the conventions used in the CSS and JSON plugins.

(It would be great if we could avoid making this a BREAKING CHANGE by using a different approach.)

snitin315: So we no longer export `SourceLocation`, `SourceRange` from `@eslint/markdown` path?

lumirlumir: Yes. Currently [CSS](https://github.com/eslint/css/blob/main/tests/types/types.test.ts#L3) and [JSON](https://github.com/eslint/json/blob/main/tests/types/types.test.ts#L23) follow it.

snitin315: I think it would be a breaking change if we remove an export from public API. So what I feel we can do is keep both exports,  mark this one as deprecated for now, and remove it in next major version.

lumirlumir: Thanks for your suggestion. I'll go with your recommendation.

Actually, I had been trying to maintain both APIs, but when I use the original code, I keep running into type errors related to duplicate types. If you're available, or if someone else on the team can take a look, I’d really appreciate some help with this.

fasttime: @lumirlumir I think you could use a `@typedef` tag to generate `export` statements in the bundle. For example, you could typedef `SourceLocation` in `src\language\markdown-source-code.js`, where the type is also imported.

```diff
 /**
  * @import { Root, Node, Html } from "mdast";
  * @import { TraversalStep, SourceLocation, FileProblem, DirectiveType, RulesConfig } from "@eslint/core";
  * @import { MarkdownLanguageOptions } from "../types.js";
+ * @typedef {SourceLocation} SourceLocation
  */
```

To export a type that is defined locally, you can use the `import()` syntax inside a typedef, and the `dedupe-types.js` script should take care to handle that properly. For example, in `src\index.js` you could use

```js
/**
 * @typedef {import("./types.ts").MarkdownRuleDefinition} MarkdownRuleDefinition
 */
```

and then the generated .d.ts file should include an export for `MarkdownRuleDefinition`.

lumirlumir: @fasttime 

Thanks a lot! I’ll go through them carefully and refactor the code to avoid any breaking changes.

src/types.ts
```diff
@@ -177,3 +177,5 @@ export type MarkdownRuleDefinition<
 			Omit<MarkdownRuleDefinitionTypeOptions, keyof Options>
 	>
 >;
+
+export type RuleModule = MarkdownRuleDefinition;
```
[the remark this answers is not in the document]
lumirlumir: I've removed it thank to #383

lumirlumir:
I've updated this PR to incorporate the latest changes from #383.

There are no longer any breaking changes in this PR.

src/index.js
```diff
@@ -17,16 +17,18 @@ import rules from "./build/rules.js";
...
 //-----------------------------------------------------------------------------
 
-/** @typedef {import("eslint").Linter.RulesRecord} RulesRecord*/
-/** @typedef {import("eslint").Linter.Config} Config*/
-/** @typedef {import("eslint").ESLint.Plugin} Plugin */
 /**
- * @typedef {import("./types.ts").MarkdownRuleDefinition<Options>} MarkdownRuleDefinition<Options>
- * @template {Partial<import("./types.ts").MarkdownRuleDefinitionTypeOptions>} [Options={}]
+ * @import { Linter } from "eslint";
+ * @import { MarkdownRuleVisitor as MRV, MarkdownRuleDefinition as MRD, MarkdownRuleDefinitionTypeOptions } from "./types.js";
```
nzakas: Can we just use the regular names? I think that makes things a lot easier to understand. 

```suggestion
 * @import { MarkdownRuleVisitor , MarkdownRuleDefinition, MarkdownRuleDefinitionTypeOptions } from "./types.js";
```

src/index.js
```diff
@@ -17,16 +17,18 @@ import rules from "./build/rules.js";
...
+ * @typedef {Linter.RulesRecord} RulesRecord
+ * @typedef {MarkdownRuleDefinition} RuleModule
+ * @typedef {MRV} MarkdownRuleVisitor
+ */
+
+/**
+ * @typedef {MRD<Options>} MarkdownRuleDefinition<Options>
```
nzakas: Then I don't think we need these.

src/rules/no-duplicate-definitions.js
```diff
@@ -70,7 +72,9 @@ export default {
 			context.options[0]?.allowFootnoteDefinitions,
 		);
 
+		/** @type {Set<string>} */
 		const definitions = new Set();
+		/** @type {Set<string>} */
```
nzakas, line 78: Can we add an extra blank line here? 
```suggestion

		/** @type {Set<string>} */
```

src/index.js
lumirlumir: This is because the type tests in `types.test.js` currently depend on these type definitions.
So, removing them in this PR would introduce a breaking change:

[https://github.com/eslint/markdown/blob/47f06b743c77067e30ce8249327da45fe9cb0a8d/tests/types/types.test.ts#L1-L6](https://github.com/eslint/markdown/blob/47f06b743c77067e30ce8249327da45fe9cb0a8d/tests/types/types.test.ts#L1-L6)

Personally, I think it would be better to keep this PR non-breaking and create a separate PR to remove these types and move them under the `@eslint/markdown/types` path when importing.

Both the JSON and CSS plugins follow a similar approach, so it would make sense to organize the types this way:

* JSON plugin:
  https://github.com/eslint/json/blob/3f754f7f5f8e182ce88fc65a909b58a0116d04ac/tests/types/types.test.ts#L3-L7
* CSS plugin:
  https://github.com/eslint/css/blob/211bf21f3c72530e60105a4f89c708bcbb00fc82/tests/types/types.test.ts#L55

If it’s acceptable, I can open a separate PR with the breaking changes by tomorrow.

src/index.js
lumirlumir: I initially tried using the original name, but without an alias, it results in a `duplicate type definitions` error when running the `tsc` command, since the same name is declared again in the following lines using the `@typedef` syntax.

(This redeclaration is intentional, for the reason I mentioned [here](https://github.com/eslint/markdown/pull/367#discussion_r2161769941).)

However, if we decide to remove these types from the `@eslint/markdown` export path and move them to the `@eslint/markdown/types` path, then these lines can be safely removed. (ref: [here](https://github.com/eslint/markdown/pull/367#discussion_r2161769941))

fasttime: Maybe instead of renaming the imported types we could import the whole `types` module as a namespace to avoid the naming collision? I.e:

```js
/**
 * @import { Linter } from "eslint";
 * @import * as Types from "./types.js";
 * @typedef {Linter.RulesRecord} RulesRecord
 * @typedef {Types.MarkdownRuleDefinition} RuleModule
 * @typedef {Types.MarkdownRuleVisitor} MarkdownRuleVisitor
 */
```

And similarly for the `MarkdownRuleDefinition` typedef below.

lumirlumir: Awesome, it works 👍

I've added a new commit https://github.com/eslint/markdown/pull/367/commits/a001891ad6627f5e7a2e5c886df22536bc635670

src/index.js
```diff
@@ -17,16 +17,19 @@ import rules from "./build/rules.js";
...
+ * @import { Linter } from "eslint";
+ * @import * as Types from "./types.js";
+ * @typedef {Linter.RulesRecord} RulesRecord
+ * @typedef {Types.MarkdownRuleDefinition} RuleModule
+ * @typedef {Types.MarkdownRuleVisitor} MarkdownRuleVisitor
+ * @typedef {Types.MarkdownRuleDefinitionTypeOptions} MarkdownRuleDefinitionTypeOptions
+ */
+
+/**
+ * @typedef {Types.MarkdownRuleDefinition<Options>} MarkdownRuleDefinition<Options>
```
lumirlumir, line 30:
```suggestion
 * @typedef {Types.MarkdownRuleDefinition<Options>} MarkdownRuleDefinition<Options>
```

This line should be separated from the typedefs above; otherwise, it causes a type error like the one below:

image

src/index.js
nzakas: Sounds good. Because the next release will be a major release, this is the time to get all the breaking changes in. :+1:

snitin315 (approved):
LGTM

src/index.js
lumirlumir: I've made a separate PR https://github.com/eslint/markdown/pull/446 👍

lumirlumir:
@fasttime @JoshuaKGoldberg

Hi everyone, is there anything I need to do on my side? If so, I’d be happy to take a look.

lumirlumir:
@JoshuaKGoldberg Thanks for the references!

I’ve explored some alternatives, but I think #453 is the best solution for now 👍

---

@nzakas I’d appreciate it if you could take a final look when you have time 😄

(FYI, I’ve resolved the merge conflicts, so the previous review has gone stale)

Issue #369, opened 2025-05-10
New Rule: link-fragments-should-be-valid


SwetaTanwar:
### Rule details

This rule should warn when a link fragment does not match any of the fragments that are automatically generated for headings in a document

### What type of rule is this?

Warns about a potential problem

### Example code

```markdown
Incorrect Code


# Heading Name

[Link](#fragment)


Correct Code


# Heading Name

[Link](#heading-name)
```

### Participation

- [x] I am willing to submit a pull request to implement this rule.

### Additional comments

Prior At: [Rule](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md051---link-fragments-should-be-valid)

snitin315:
I'm 👍🏻 for this rule & I feel this should also be recommended.

lumirlumir:
Yes, I agree that this rule is worth implementing.

However, I’m concerned that it has some tricky aspects.

Some Markdown parsers handle heading link fragments in their own ways. For example, one parser might convert `Heading with .dot` to `heading-with-dot`, while another might interpret it as `heading-with-.dot`. This might not be the exact case, but I’ve seen many websites generate link fragments differently.

Additionally, in general usage, we need to consider that many Markdown parsers support custom link fragments like this:

```md
# Heading 1 {#custom-heading-1-it-is-possible-for-many-markdown-parsers}
```

This is commonly used syntax, supported by popular parsers like `markdown-it` and `remark`.

---

So in conclusion, it would be great to implement and recommend this rule if we can account for these cases properly. But until we find a concrete way to handle these situations, I think we should hold off on proceeding and have a discussion about how to implement the rule.

SwetaTanwar:
> I think this is a good idea. @SwetaTanwar Are you planning to submit a PR for this? If not, I can take it on.

Yes, I will be picking it up

PR #380, opened 2025-05-19
feat: add no-missing-link-fragments rule

SwetaTanwar:
#### Prerequisites checklist

- [ ] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).


docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,69 @@
+# no-missing-link-fragments
+
+Disallow link fragments that don't exist in the document.
+
+## Rule Details
+
+This rule verifies that all link fragments (using the `#` syntax in URLs) reference valid headings that exist in the document. Link fragments are commonly used to navigate to specific sections within a document, and when they're invalid, users won't be directed to the intended location.
+
+### Options
```
snitin315, line 9: This should be h2, 

```suggestion
## Options
```

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,69 @@
...
+
+This rule has an options object with the following properties (parameter naming follows markdownlint's MD051 rule for compatibility):
+
+* `ignoreCase`: Boolean, default `false` - When `true`, fragment case will be ignored when comparing to heading IDs
+* `allowPattern`: String, default `""` - Regular expression pattern for fragments to ignore
```
snitin315, line 14: To keep the option naming consistent with other rules.

```suggestion
* `ignoreCase: boolean` - When `true`, fragment case will be ignored when comparing to heading IDs (default `false`).
* `allowPattern: string` - Regular expression pattern for fragments to ignore (default `""`).
```

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,69 @@
...
+## Heading with spaces
+
+<!-- With ignoreCase: true -->
+[Case insensitive link](#INTRODUCTION)
+
+## Introduction
+
+<!-- With allowPattern: "^section-" -->
+[Ignored pattern link](#section-123)
+```
```
snitin315, line 61: Let's improve the example section by separating the types of examples for each configuration. For example:

- Let's add the basic incorrect/correct code examples for `markdown/no-missing-link-fragments: "error"` in the Rule details section.

- Under the options section, we can add separate example code blocks for:
    - Examples for **correct code** when configured as `"markdown/no-missing-link-fragments": ["error", { ignoreCase: true } ]`
    
    
     ```md
     ## Introduction

     [Case insensitive link](#INTRODUCTION)
     ```
     - Examples for **correct code** when configured as `"markdown/no-missing-link-fragments": ["error", { allowPattern: "^section-" } ]` 
     
     ```md
     [Ignored pattern link](#section-123)
     ```

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Further Reading
+
+This rule is similar to the markdownlint MD051 rule, which checks that link fragments are valid. It ensures that all links that use fragment identifiers point to actual headings in the document.
```
snitin315, line 65: Let's remove this section in favor of the `Prior Art` section.

snitin315:
```suggestion
### Prior Art

- [MD051 - Link fragments should be valid](https://github.com/DavidAnson/markdownlint/blob/main/doc/md051.md)
```

snitin315: Also let's add this section at the bottom after When not to use it section.

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,69 @@
...
+
+## When Not To Use It
+
+You might want to disable this rule if you're linking to fragments that are generated by other means, such as manually created anchor elements in HTML within the markdown. Alternatively, you can use the `allowPattern` option to exclude specific fragment patterns from validation.
```
snitin315, line 69: Let's update the language for consistency with other rule documentations:

```suggestion
If you aren't concerned with fragments generated by other means, like manually created anchor elements in HTML within the markdown, you can safely disable this rule.
```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,138 @@
...
+		docs: {
+			recommended: true,
+			description:
+				"Disallow link fragments that don't exist in the document",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-missing-link-fragments.md",
+		},
+
+		defaultOptions: [
+			{
+				ignoreCase: false,
```
snitin315, line 37: As per the docs.

```suggestion
				ignoreCase: false,
				allowPattern: "",
```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,138 @@
...
+		messages: {
+			missingFragment:
+				"Link fragment '#{{fragment}}' does not exist in the document.",
+		},
+	},
+
+	create(context) {
+		const options = context.options[0] || {};
+		const ignoreCase = options.ignoreCase;
+		const allowPattern = options.allowPattern || "";
```
snitin315, line 67: We will get it from the `defaultOptions`.

```suggestion
		const allowPattern = options.allowPattern;
```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,138 @@
...
+
+		const allowedRegex = allowPattern
```
snitin315, line 69:
```suggestion
		const allowedRegex = Boolean(allowPattern)
```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		return {
+			heading(node) {
+				if (node.children && node.children.length > 0) {
+					const headingText = node.children
+						.filter(child => child.type === "text")
+						.map(child => child.value)
+						.join("");
+
```
snitin315, line 97: We're filtering only `child.type === "text"`, but markdown headings can contain emphasis, inline code, etc. This may lead to missed fragments. For example the followings are false positive:

```md
<!-- eslint markdown/no-missing-link-fragments: "error" -->

## `intro`

[abc](#intro)

## _intro1_

[abc](#intro1)

## __intro2__

[abc](#intro2)

## ~intro3~

[abc](#intro3)

## *intro4*

[abc](#intro4)
```

```console
   5:1  error  Link fragment '#intro' does not exist in the document   markdown/no-missing-link-fragments
   9:1  error  Link fragment '#intro1' does not exist in the document  markdown/no-missing-link-fragments
  13:1  error  Link fragment '#intro2' does not exist in the document  markdown/no-missing-link-fragments
  21:1  error  Link fragment '#intro4' does not exist in the document  markdown/no-missing-link-fragments
```

snitin315 (changes requested):
Thank you for taking this up. This looks like a good start. I have added a few suggestions on documentation & few false postives. I think there are few more cases we should cover that [MD051](https://github.com/DavidAnson/markdownlint/blob/main/doc/md051.md) handles as well:

1. HTML ids:

    ```md
    <a id="bookmark"></a>
    
    [Link](#bookmark) // valid
    ```

2. `#top`:  [HTML links to #top scroll to the top of a document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scrolling-to-a-fragment).

    ```md
    [Link](#top)
    ```

3. Custom fragment syntax used by GitHub to highlight [specific content in a document](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet#linking-to-markdown#linking-to-markdown).
    
    ```md
    [Link](#L20) // Valid
    
    [Link](#L19C5-L21C11) // Valid
    ```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,138 @@
...
+		 * @returns {string} The normalized heading ID
+		 */
+		function normalizeHeadingId(text) {
+			return text
+				.toLowerCase()
+				.replace(/[^\w\s-]/gu, "") // Remove non-word chars except spaces and hyphens
+				.replace(/\s+/gu, "-") // Replace spaces with hyphens
+				.replace(/-+/gu, "-") // Replace multiple hyphens with a single hyphen
+				.trim();
+		}
```
snitin315, line 88: This regex is fragile because various tools handle heading slugs differently, for example:

 1. GitHub / GFM (GitHub Flavored Markdown)
 
GitHub uses [this algorithm](https://github.com/gjtorikian/html-pipeline/blob/f13a1534cb650ba17af400d1acd3a22c28004c09/lib/html/pipeline/toc_filter.rb) where:
 
 - convert heading to lowercase
 - remove punctuation
 - convert spaces to dashes
 - Unicode characters are allowed and lowercased
 - Duplicate headings get appended `-1`, `-2`, etc.
 
 ```md
 ## Résumé
 
 [link](#résumé) // Error, but this is what github generates
 
 [link2][#rsum] // No error because this is what the current regexp generates.
 
 ```
 
 So I think instead of write a custom regex, maybe we can use something like [github-slugger](https://github.com/Flet/github-slugger) for more reliable results.

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,139 @@
...
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+/** @type {NoMissingLinkFragmentsRuleDefinition} */
+export default {
+	meta: {
+		type: "problem",
+
+		docs: {
+			recommended: true,
```
lumirlumir, line 29:
```suggestion

```
Should we include this rule in the recommended set? 

As I mentioned in the issue comments (https://github.com/eslint/markdown/issues/369#issuecomment-2868904376), it doesn’t cover all general cases—so it might be best not to mark it as recommended.

README.md
```diff
@@ -81,6 +81,7 @@ export default defineConfig([
 | [`no-html`](./docs/rules/no-html.md) | Disallow HTML tags | no |
 | [`no-invalid-label-refs`](./docs/rules/no-invalid-label-refs.md) | Disallow invalid label references | yes |
 | [`no-missing-label-refs`](./docs/rules/no-missing-label-refs.md) | Disallow missing label references | yes |
+| [`no-missing-link-fragments`](./docs/rules/no-missing-link-fragments.md) | Disallow link fragments that don't exist in the document | yes |
```
lumirlumir, line 84:
```suggestion
| [`no-missing-link-fragments`](./docs/rules/no-missing-link-fragments.md) | Disallow link fragments that don't exist in the document | no |
```

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,68 @@
...
+## Options
+
+This rule has an options object with the following properties:
+
+* `ignoreCase: boolean` - When `true`, fragment case will be ignored when comparing to heading IDs (default `false`).
+* `allowPattern: string` - Regular expression pattern for fragments to ignore (default `""`).
+
+Example configuration:
+
+```json
```
lumirlumir, line 18:
```suggestion
```js
```

Should we use `json` for the example? I think using `js` is more appropriate, since ESLint v9 only supports `js` and `ts` configuration files.

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,139 @@
...
+		messages: {
+			missingFragment:
+				"Link fragment '#{{fragment}}' does not exist in the document.",
+		},
+	},
+
+	create(context) {
+		const options = context.options[0] || {};
+		const ignoreCase = options.ignoreCase;
+		const allowPattern = options.allowPattern;
```
lumirlumir, line 68:
```suggestion
		const [{ ignoreCase, allowPattern }] = context.options;
```

We can use destructuring syntax here, as the options are validated by the JSON schema beforehand, so the fallback `|| {}` is unnecessary.

src/rules/no-missing-link-fragments.js
lumirlumir: @snitin315

I also prefer using a boolean cast. Unfortunately, though, an internal ESLint rule currently prevents us from doing so.

tests/rules/no-missing-link-fragments.test.js
```diff
@@ -0,0 +1,242 @@
...
+	},
+	language: "markdown/commonmark",
+});
+
+ruleTester.run("no-missing-link-fragments", rule, {
+	valid: [
+		// Simple case with matching fragment
+		`# Heading One
+		
+		[Link to heading](#heading-one)`,
```
lumirlumir, line 30:
```suggestion
		dedent`# Heading One
		
		[Link to heading](#heading-one)`,
```

If you're planning to use indented template literals, you should use `dedent`; otherwise, it may lead to unintended results. ([AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiXCIsXCJqc29uXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgSlNPTiBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG4gKiB0aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcbiAqXFxuICogVGhlIHRhYnMgYXJlOlxcbiAqXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICpcXG4gKiBZb3UgY2FuIGNoYW5nZSB0aGUgd2F5IHRoYXQgdGhlIEpTT04gY29kZSBpcyBpbnRlcnByZXRlZFxcbiAqIGJ5IGNsaWNraW5nIFxcXCJKU09OXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKlxcbiAqIFRoaXMgZXhhbXBsZSBpcyBpbiBKU09OQyBtb2RlLCB3aGljaCBhbGxvd3MgY29tbWVudHMuXFxuICovXFxuXFxue1xcbiAgICBcXFwia2V5MVxcXCI6IFt0cnVlLCBmYWxzZSwgbnVsbF0sXFxuICAgIFxcXCJrZXkyXFxcIjoge1xcbiAgICAgICAgXFxcImtleTNcXFwiOiBbMSwgMiwgXFxcIjNcXFwiLCAxZTEwLCAxZS0zXVxcbiAgICB9XFxufVwiLFwibWFya2Rvd25cIjpcIiMgSGVhZGluZyBPbmVcXG5cXHRcXHRcXG5cXHRcXHRbTGluayB0byBoZWFkaW5nXSgjaGVhZGluZy1vbmUpXCIsXCJjc3NcIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBDU1MgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKlxcbiAqIFRoZSB0YWJzIGFyZTpcXG4gKlxcbiAqIC0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG4gKiAgIGJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIHRoZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUuIFlvdVxcbiAqICAgY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxcbiAqXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBDU1MgY29kZSBpcyBpbnRlcnByZXRlZFxcbiAqIGJ5IGNsaWNraW5nIFxcXCJDU1NcXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqL1xcblxcbkBpbXBvcnQgdXJsKCdodHRwczovL2ZvbnRzLmdvb2dsZWFwaXMuY29tL2NzczI%2FZmFtaWx5PVJvYm90bzp3Z2h0QDQwMDs3MDAmZGlzcGxheT1zd2FwJyk7XFxuXFxuYm9keSB7XFxuXFx0Zm9udC1mYW1pbHk6IHNhbnMtc2VyaWY7XFxufVxcblxcbmgxIHtcXG5cXHRjb2xvcjogIzMzMztcXG59XFxuXFxucCB7XFxuXFx0bWFyZ2luOiAxZW0gMDtcXG59XCIsXCJodG1sXCI6XCI8IURPQ1RZUEUgaHRtbD5cXG48IS0tXFxuVHlwZSBvciBwYXN0ZSBzb21lIEhUTUwgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG5cXG5UaGUgdGFicyBhcmU6XFxuXFxuLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbmJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIHRoZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUuIFlvdVxcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4tLT5cXG5cXG48aHRtbCBsYW5nPVxcXCJlblxcXCI%2BXFxuICAgIDxoZWFkPlxcbiAgICAgICAgPG1ldGEgY2hhcnNldD1cXFwiVVRGLThcXFwiPlxcbiAgICAgICAgPHRpdGxlPkhUTUw8L3RpdGxlPlxcbiAgICA8L2hlYWQ%2BXFxuICAgIDxib2R5PlxcbiAgICAgICAgPHA%2BVGV4dDwvcD5cXG4gICAgPC9ib2R5PlxcbjwvaHRtbD5cIn0sXCJsYW5ndWFnZVwiOlwibWFya2Rvd25cIixcImpzT3B0aW9uc1wiOntcInBhcnNlclwiOlwiZXNwcmVlXCIsXCJzb3VyY2VUeXBlXCI6XCJtb2R1bGVcIixcImVzVmVyc2lvblwiOlwibGF0ZXN0XCIsXCJpc0pTWFwiOnRydWV9LFwianNvbk9wdGlvbnNcIjp7XCJqc29uTW9kZVwiOlwianNvbmNcIn0sXCJjc3NPcHRpb25zXCI6e1wiY3NzTW9kZVwiOlwiY3NzXCJ9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJnZm1cIn0sXCJ3cmFwXCI6dHJ1ZSxcInZpZXdNb2Rlc1wiOntcImFzdFZpZXdcIjpcInRyZWVcIixcInNjb3BlVmlld1wiOlwiZmxhdFwiLFwicGF0aFZpZXdcIjpcImNvZGVcIn0sXCJwYXRoSW5kZXhcIjp7XCJpbmRleFwiOjAsXCJpbmRleGVzXCI6MX0sXCJlc3F1ZXJ5U2VsZWN0b3JcIjp7XCJzZWxlY3RvclwiOlwiXCJ9fSxcInZlcnNpb25cIjowfSI%3D))

The same applies to all test cases.

src/rules/no-missing-link-fragments.js
nzakas: :+1: to adding to recommended

docs/rules/no-missing-link-fragments.md
SwetaTanwar: Changing the type to JS gives eslint error as below:

Screenshot 2025-05-20 at 11 31 14 PM

docs/rules/no-missing-link-fragments.md
```diff
@@ -1,68 +1,83 @@
 # no-missing-link-fragments
 
-Disallow link fragments that don't exist in the document.
+Ensures that link fragments (URLs that start with `#`) reference valid headings or anchors in the document. This rule helps prevent broken internal links.
```
nzakas, line 3: For our documentation, this should be a single sentence. I think the original was fine.

```suggestion
Disallow link fragments that don't exist in the document.
```

docs/rules/no-missing-link-fragments.md
```diff
@@ -1,68 +1,83 @@
...
 
 ## Rule Details
```
nzakas, line 5: For our documentation format, we also have a "Background" section that gives some context as to what the rule is meant to address. Please see other rules in the plugin for examples.

docs/rules/no-missing-link-fragments.md
```diff
@@ -1,68 +1,83 @@
...
-    "ignoreCase": true,
-    "allowPattern": "^section-"
-  }]
-}
+# Some Heading
+[Case Mismatch](#SOME-HEADING) <!-- Default: case-sensitive -->
 ```
 
-### Examples
+### Correct Code Examples
```
nzakas, line 20:
```suggestion
Examples of **correct** code for this rule:
```

docs/rules/no-missing-link-fragments.md
```diff
@@ -1,68 +1,83 @@
...
 
-If you aren't concerned with fragments generated by other means, like manually created anchor elements in HTML within the markdown, you can safely disable this rule.
+You might consider disabling this rule if:
+
+* You are using a Markdown processor or static site generator that has a significantly different algorithm for generating heading IDs, and this rule produces too many false positives.
+* You have many dynamically generated links or fragments that cannot be easily covered by the `allowPattern` option.
+
+## Further Reading
 
-### Prior Art
```
nzakas: We should keep this section.

docs/rules/no-missing-link-fragments.md
```diff
@@ -1,68 +1,83 @@
...
-- [MD051 - Link fragments should be valid](https://github.com/DavidAnson/markdownlint/blob/main/doc/md051.md)
\ No newline at end of file
+* [GitHub's heading anchor links](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links)
+* [CommonMark Specification](https://spec.commonmark.org/)
```
nzakas, line 83: This link doesn't point to anything specific about link fragments?

docs/rules/no-missing-link-fragments.md
nzakas, line 1: I think overall the recent changes to this file went in the wrong direction. Please take a look at my comments and please double-check the format with other rule docs.

src/rules/no-missing-link-fragments.js
```diff
@@ -57,81 +78,101 @@ export default {
...
-		const allowedRegex = allowPattern
-			? new RegExp(allowPattern, "u")
+		const ignoreCase = options.ignoreCase || false;
+		const allowPattern = options.allowPattern
+			? new RegExp(options.allowPattern, "u")
 			: null;
 
-		const headingIds = new Set();
-		const lowercaseHeadingIds = new Set();
+		const slugger = new GithubSlugger();
```
nzakas, line 93: Let's move this into the global scope. There's no need to create a new reference each time the rule is created.

src/rules/no-missing-link-fragments.js
```diff
@@ -57,81 +78,101 @@ export default {
 		],
 
 		messages: {
-			missingFragment:
-				"Link fragment '#{{fragment}}' does not exist in the document.",
+			invalidFragment:
+				"Link fragment '{{fragment}}' does not reference a valid heading or anchor.",
```
nzakas, line 82: I think this message is confusing.

```suggestion
				"Link fragment '{{fragment}}' does not reference a heading or anchor in this document.",
```

src/rules/no-missing-link-fragments.js
```diff
@@ -57,81 +78,101 @@ export default {
...
+				if (!url || !url.startsWith("#")) {
+					return;
+				}
+
+				const fragment = url.slice(1);
+				if (!fragment) {
+					return;
+				}
+
+				if (allowPattern && allowPattern.test(fragment)) {
```
nzakas, line 158:
```suggestion
				if (allowPattern?.test(fragment)) {
```

tests/rules/no-missing-link-fragments.test.js
```diff
@@ -0,0 +1,357 @@
...
+		dedent`
+            # Heading with 🚀 emoji
+            [Link](#heading-with--emoji)
+
+            # Héading with àccènt chàrâctérs
+            [Link](#héading-with-àccènt-chàrâctérs)
+
+            # Mix: _Héading_ with 🚀 & \`code\`
+            [Link](#mix-héading-with---code)
+        `,
```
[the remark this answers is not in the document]
snitin315: I checked this locally, and it seems that the slug generate by GithubSlugger is different in the case of bun:

Screenshot 2025-05-25 at 3 50 08 PM

I think we can report this in the GithubSlugger repo & for now we can skip this test case for bun by something like `(typeof process !== 'undefined') && (process.release.name === 'node')`

docs/rules/no-missing-link-fragments.md
```diff
@@ -0,0 +1,99 @@
+# no-missing-link-fragments
+
+Disallow link fragments that don't exist in the document.
+
+## Background
+
+Ensures that link fragments (URLs that start with `#`) reference valid headings or anchors in the document. This rule helps prevent broken internal links.
```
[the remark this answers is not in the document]
SwetaTanwar: Added

nzakas:
@SwetaTanwar please double-check the merge conflict.

nzakas (approved):

snitin315 (dismissed):
LGTM, thanks!

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,174 @@
...
+ * }>} NoMissingLinkFragmentsRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const githubLineReferencePattern = /^L\d+(?:C\d+)?(?:-L\d+(?:C\d+)?)?$/u;
+const customHeadingIdPattern = /\{#([a-z0-9_-]+)\}\s*$/u;
+const markdownInlineFormattingPattern = /[*_~`]/gu;
```
lumirlumir, line 30:
```suggestion

```

Could you remove this regex pattern?

Directly stripping Markdown inline formatting using regex is quite fragile.

For example, `github-slugger` does **not** strip underscore (`_`) characters:

[https://github.com/Flet/github-slugger/blob/master/test/fixtures.json#L18-L21](https://github.com/Flet/github-slugger/blob/master/test/fixtures.json#L18-L21)

If we apply this logic directly in the `getHeadingId` function, it ends up stripping intentional underscores. While the original intention might have been to remove Markdown syntax, this causes unintended removal of valid characters. As a result, the generated slug may differ from what GitHub actually uses, potentially leading to false negatives or positives for users.

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,174 @@
...
+		 * @returns {string} The normalized heading ID.
+		 */
+		function getHeadingId(headingText) {
+			const customIdMatch = headingText.match(customHeadingIdPattern);
+			if (customIdMatch) {
+				return customIdMatch[1];
+			}
+			const plainText = headingText
+				.replace(markdownInlineFormattingPattern, "")
+				.trim();
```
lumirlumir, line 105:
```suggestion
```

Could you remove this line as mentioned above?

If we remove this line, one might ask: *How do we handle inline Markdown formatting syntax like backticks (\`), asterisks (`*`), tildes (`~`), and underscores (`_`)?\*

Fortunately, the `github-slugger` package already handles most of these cases—it strips backticks, asterisks, and tildes automatically. However, it does **not** strip underscores:

```js
import { slug } from "github-slugger";

// Stripped
console.log(slug("`")); // ""
console.log(slug("*")); // ""
console.log(slug("~")); // ""

// Not stripped
console.log(slug("_")); // "_"
```

So how do we distinguish between underscores used intentionally and those used for Markdown formatting?

This distinction is important because the Markdown heading below generates a slug of `test_`, not `__test___`:

```md
# __test___
```

To handle this accurately, we can't rely on regex. Instead, we need to traverse to the deepest `text` nodes, extract their raw values, and reassemble them—effectively reconstructing the heading text **without** interpreting Markdown syntax manually. This ensures the slug generation mirrors GitHub's behavior and avoids false positives or negatives.

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,174 @@
...
+				if (!fragmentIds.has(normalizedFragment)) {
+					context.report({
+						loc: node.position,
+						messageId: "invalidFragment",
+						data: { fragment },
+					});
+				}
+			},
+		};
+	},
```
lumirlumir, line 173: Could you refactor this logic to avoid false positives and negatives?

It might be better to collect all `heading`, `html`, and `link` nodes sequentially during traversal, and then perform the actual validation in the `root:exit` visitor. This way, we ensure that all headings and references are available before any comparisons are made.

The current implementation depends on the order in which `heading` and `html` nodes appear. So if a link is defined earlier in the document and the corresponding heading appears later, it can lead to false positives or negatives.

Also, it would be great if you could add more test cases for this.

tests/rules/no-missing-link-fragments.test.js
```diff
@@ -0,0 +1,366 @@
...
+	language: "markdown/commonmark",
+});
+
+ruleTester.run("no-missing-link-fragments", rule, {
+	valid: [
+		// Basic heading match
+		dedent`
+            # Heading Name
+            [Link](#heading-name)
+        `,
```
lumirlumir, line 32:
```suggestion
		dedent`
                # Heading Name
                [Link](#heading-name)`,
```

Could you align the strings using `dedent`?

It would be great if you could follow the same alignment style you used in [`no-missing-atx-heading-space.test.js`](https://github.com/eslint/markdown/blob/main/tests/rules/no-missing-atx-heading-space.test.js).

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,174 @@
...
+ *   }];
+ * }>} NoMissingLinkFragmentsRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const githubLineReferencePattern = /^L\d+(?:C\d+)?(?:-L\d+(?:C\d+)?)?$/u;
+const customHeadingIdPattern = /\{#([a-z0-9_-]+)\}\s*$/u;
```
lumirlumir, line 29:
```suggestion
const customHeadingIdPattern = /\{#([a-z0-9_-]+)\}\s*$/u;
```

Could you update `customHeadingIdPattern` to allow a wider range of characters?

Some users may use sluggers that include non-English characters, so supporting a broader character set would be helpful.

src/rules/no-missing-link-fragments.js
SwetaTanwar: Thanks for pointing this out, have fixed it

src/rules/no-missing-link-fragments.js
SwetaTanwar: Nice catch, fixed it

nzakas:
@lumirlumir please re-review this when you can.

lumirlumir:
@nzakas Sure, but I think my reviews haven’t been addressed yet? It seems like a bit more time might be needed.

@SwetaTanwar If you don’t mind, and if you’re still working on it, could you mark this PR as a draft while you continue making progress? :)

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,183 @@
...
+						htmlText.endsWith("-->")
+					) {
+						return;
+					}
+
+					for (const match of htmlText.matchAll(htmlIdNamePattern)) {
+						const extractedId = match[1];
+						const finalId = slugger.slug(extractedId);
+						fragmentIds.add(
+							ignoreCase ? finalId.toLowerCase() : finalId,
```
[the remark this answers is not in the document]
SwetaTanwar: Already checking that, added more test cases in https://github.com/eslint/markdown/pull/380/commits/b9ec4da2c692f78c3b512b54e2184e7cb40f9929

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,183 @@
...
+
+			html(node) {
+				if (node.value) {
+					const htmlText = node.value.trim();
+					if (
+						htmlText.startsWith("<!--") &&
+						htmlText.endsWith("-->")
+					) {
+						return;
+					}
```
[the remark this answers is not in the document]
SwetaTanwar: Fixed in https://github.com/eslint/markdown/pull/380/commits/757a61bc218fddc658d1f066002a64abc912c07e

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,183 @@
...
+
+				if (customIdMatch) {
+					baseId = customIdMatch[1];
+				} else {
+					const tempSlugger = new GithubSlugger();
+					baseId = tempSlugger.slug(rawHeadingText);
+				}
+
+				const finalId = slugger.slug(baseId);
+				fragmentIds.add(ignoreCase ? finalId.toLowerCase() : finalId);
```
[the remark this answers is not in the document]
SwetaTanwar: Can you share any example of a false positive or negative, the following case works as expected:

```md
<!-- eslint markdown/no-missing-link-fragments: "error" -->

# foo

## foo

[Link](#foo) // No error

[Link](#foo-1) // No error

[Link](#foo-2) // Error Link fragment '#foo-2' does not reference a heading or anchor in this document
```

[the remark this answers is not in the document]
SwetaTanwar: Added test cases in https://github.com/eslint/markdown/pull/380/commits/ff82b1b133d73bf6fe7199d10c13c6649cfef615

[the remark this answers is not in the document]
SwetaTanwar: I got it, fixed the behavior in https://github.com/eslint/markdown/pull/380/commits/4c987306861da835e6093be08bc37f2c98f19eb2

SwetaTanwar:
@lumirlumir The checks are passing now, fixed the CI failure due to `Duplicate identifier 'MarkdownNode`

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+//-----------------------------------------------------------------------------
+
+import GithubSlugger from "github-slugger";
+
+//-----------------------------------------------------------------------------
+// Type Definitions
+//-----------------------------------------------------------------------------
+
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{
```
lumirlumir:
```suggestion

/** @typedef {import("mdast").Node} Node */
/**
 * @typedef {import("../types.ts").MarkdownRuleDefinition<{
```

Could you move the `Node` type definition to the top level?

References:

- https://github.com/eslint/markdown/blob/main/src/rules/no-invalid-label-refs.js#L16-L17
- https://github.com/eslint/markdown/pull/367#discussion_r2081797899

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+ * @param {string} fragment The fragment to check
+ * @returns {boolean} Whether the fragment is a valid GitHub line reference
+ */
+function isGitHubLineReference(fragment) {
+	return githubLineReferencePattern.test(fragment);
+}
+
+/**
+ * Extracts the text from a heading node
+ * @param {import("mdast").Node} node The heading node to extract text from
```
lumirlumir:
```suggestion
 * @param {Node} node The node from which to recursively extract text
```

Could you update the comment? Since, current logic looks into other node type like `text` recursively, I think it's more accurate.

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+ * @returns {string} The extracted text
+ */
+function extractText(node) {
+	if ("value" in node && typeof node.value === "string") {
```
lumirlumir:
```suggestion
	if ("value" in node) {
```

Could you remove `node.value==="string"? The `value`'s type is always `string`, so the check might be unnecessary.

image

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+		return node.value;
+	}
+	if ("children" in node && Array.isArray(node.children)) {
```
lumirlumir:
```suggestion
	if ("children" in node) {
```

Same here.

image

tests/rules/no-missing-link-fragments.test.js
```diff
@@ -0,0 +1,484 @@
...
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+	plugins: {
+		markdown,
+	},
+	language: "markdown/commonmark",
```
lumirlumir:
```suggestion
	language: "markdown/gfm",
```

Could you mark it with `gfm`? We're using strikethrough syntax now, so that would be more appropriate.

(Alternatively, you can manually set the `language` option for the test cases using strikethrough syntax.)

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+ * Checks if the fragment is a valid GitHub line reference
+ * @param {string} fragment The fragment to check
+ * @returns {boolean} Whether the fragment is a valid GitHub line reference
+ */
+function isGitHubLineReference(fragment) {
+	return githubLineReferencePattern.test(fragment);
+}
+
+/**
+ * Extracts the text from a heading node
```
lumirlumir:
```suggestion
 * Extracts the text recursively from a node
```

src/rules/no-missing-link-fragments.js
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const githubLineReferencePattern = /^L\d+(?:C\d+)?(?:-L\d+(?:C\d+)?)?$/u;
+const customHeadingIdPattern = /\{#([^}\s]+)\}\s*$/u;
+const htmlCommentPattern = /<!--[\s\S]*?-->/gu;
+const htmlIdNamePattern = /<(?:[^>]+)\s+(?:id|name)="([^"]+)"/gu;
```
lumirlumir:
```suggestion
const htmlIdNamePattern = /<(?:[^>]+)\s+(?:id|name)="([^"]+)"/gu;
```

Could you update the regex to also handle single quotes?

Reference: https://github.com/eslint/markdown/blob/main/src/rules/require-alt-text.js#L33

src/rules/no-missing-link-fragments.js
SwetaTanwar: Was getting typescript error thus added this check:

Screenshot 2025-06-10 at 10 18 11 AM

Screenshot 2025-06-10 at 10 18 02 AM

If we are removing this, I need to add explicit type as string, let me know if this is fine

Screenshot 2025-06-10 at 10 19 43 AM

src/rules/no-missing-link-fragments.js
SwetaTanwar: Same for this as well, have explicitly specified the type now

src/rules/no-missing-link-fragments.js
lumirlumir: Your suggestion looks good to me 👍. Let's use an explicit type.

snitin315 (approved):
LGTM. Nice job, and thanks for being so patient with all the back and forth.

PR #392, opened 2025-05-24
feat: added new rule table-column-count

### LINKED ISSUE #389: New Rule: table-column-count
### Rule details

This rule is triggered when a GitHub Flavored Markdown table does not have the same number of cells in every row

### Example code

```markdown
This table's second data row has too few cells and its third data row has too many cells:

| Header | Header |
| ------ | ------ |
| Cell   | Cell   |
| Cell   |
| Cell   | Cell   | Cell   |
To fix these issues, ensure every row has the same number of cells:

| Header | Header |
| ------ | ------ |
| Cell   | Cell   |
| Cell   | Cell   |
| Cell   | Cell   |
```

### Participation

### Additional comments

Prior Art: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md056---table-column-count

SwetaTanwar:
#### Prerequisites checklist

This PR adds a new rule `table-column-count` to disallow data rows in a [GitHub Flavored Markdown table](https://github.github.com/gfm/#tables-extension-) from having more cells than the header row

Added the `table-column-count` rule, along with documentation and tests.

#### Related Issues
fixes #389

### DIFF
--- a/docs/rules/table-column-count.md
+++ b/docs/rules/table-column-count.md
@@ -4,7 +4,7 @@ Disallow data rows in a GitHub Flavored Markdown table from having more cells th
 
 ## Background
 
-In GitHub Flavored Markdown [tables](https://github.github.com/gfm/#tables-extension-), rows should maintain a consistent number of cells. While variations are sometimes tolerated, data rows having *more* cells than the header can lead to lost data or rendering issues. This rule focuses on preventing data rows from exceeding the header's column count.
+In GitHub Flavored Markdown [tables](https://github.github.com/gfm/#tables-extension-), rows should maintain a consistent number of cells. While variations are sometimes tolerated, data rows having *more* cells than the header can lead to lost data or rendering issues. This rule prevents data rows from exceeding the header's column count.
 
 ## Rule Details
 
@@ -14,7 +14,7 @@ In GitHub Flavored Markdown [tables](https://github.github.com/gfm/#tables-exten
 
 This rule is triggered if a data row in a GFM table contains more cells than the header row. It does not flag rows with fewer cells than the header.
 
-Examples of **incorrect** code for this rule (i.e., will be flagged):
+Examples of **incorrect** code for this rule:
 
 ```markdown
 <!-- eslint markdown/table-column-count: "error" -->
@@ -28,7 +28,7 @@ Examples of **incorrect** code for this rule (i.e., will be flagged):
 | 1 | 2 | <!-- This data row has 2 cells, header has 1 -->
 ```
 
-Examples of **correct** code for this rule (i.e., will NOT be flagged):
+Examples of **correct** code for this rule:
 
 ```markdown
 <!-- eslint markdown/table-column-count: "error" -->
@@ -48,7 +48,7 @@ Examples of **correct** code for this rule (i.e., will NOT be flagged):
 | Col A | Col B | Col C |
 | ----- | ----- | ----- |
 | 1     |       | 3     |
-| 4     | 5     |       |
+| 4     | 5     |
 
 <!-- Single column table -->
 | Single Header |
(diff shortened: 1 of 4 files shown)


tests/rules/table-column-count.test.js
```diff
@@ -0,0 +1,142 @@
...
+					messageId: "inconsistentColumnCount",
+					data: {
+						dataRowIndex: "1",
+						actualCells: "3",
+						expectedCells: "2",
+					},
+					line: 5,
+					column: 1,
+					endLine: 5,
+					endColumn: 30,
```
snitin315, line 137: Instead of reporting the whole row, we should report only the extra cells, similar to markdownlint:

Screenshot 2025-05-25 at 2 29 39 PM

src/rules/table-column-count.js
```diff
@@ -0,0 +1,68 @@
...
+				"Disallow data rows in a GitHub Flavored Markdown table from having more cells than the header row",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/table-column-count.md",
+		},
+
+		fixable: null,
+		schema: [],
+
+		messages: {
+			inconsistentColumnCount:
+				"Data row {{dataRowIndex}} (1-indexed) has {{actualCells}} cells, but header has {{expectedCells}} cells (should not exceed header count).",
```
snitin315, line 35: I think we can improve the message:

```suggestion
				"Table column count mismatch (Expected: {{expectedCells}}, Actual: {{actualCells}}), extra data will be ignored.",
```

snitin315 (commented):
This looks like a good start, I have a few suggestions.

tests/rules/table-column-count.test.js
```diff
@@ -0,0 +1,130 @@
...
+				{
+					messageId: "inconsistentColumnCount",
+					data: { actualCells: "3", expectedCells: "2" },
+					line: 5,
+					column: 21,
+					endLine: 5,
+					endColumn: 30,
+				},
+			],
+		},
```
snitin315, line 128: Let's add few more cases on different type of table syntax:


```js
<p>| abc | def |
| --- |
| bar |</p>


| abc | defghi |
:-: | -----------:
bar | baz

> | f\|oo  |
> | ------ |
> | b `\|` az |
> | b **\|** im |



```

docs/rules/table-column-count.md
```diff
@@ -0,0 +1,65 @@
...
+
+In GitHub Flavored Markdown [tables](https://github.github.com/gfm/#tables-extension-), rows should maintain a consistent number of cells. While variations are sometimes tolerated, data rows having *more* cells than the header can lead to lost data or rendering issues. This rule prevents data rows from exceeding the header's column count.
+
+## Rule Details
+
+> [!IMPORTANT] <!-- eslint-disable-line -- This should be fixed in https://github.com/eslint/markdown/issues/294 -->
+>
+> This rule relies on the `table` AST node, typically available when using a GFM-compatible parser (e.g., `language: "markdown/gfm"`).
+
+This rule is triggered if a data row in a GFM table contains more cells than the header row. It does not flag rows with fewer cells than the header.
```
lumirlumir, line 15: > "It does not flag rows with fewer cells than the header."

Can I ask why this rule doesn’t report rows with fewer cells than the header? Is there a specific constraint?

src/rules/table-column-count.js
```diff
@@ -0,0 +1,79 @@
...
+				"Table column count mismatch (Expected: {{expectedCells}}, Actual: {{actualCells}}), extra data starting here will be ignored.",
+		},
+	},
+
+	create(context) {
+		return {
+			table(node) {
+				if (!node.children || node.children.length < 1) {
+					return;
+				}
```
lumirlumir, line 41:
```suggestion
				if (node.children.length < 1) {
					return;
				}
```

Can I ask why this line is needed? I can't think of any cases where `node.children.length < 1`.

Also, the `Table` node always has a `children` property as mentioned in its type definition. Can we remove the `!node.children` check?

image

image

src/rules/table-column-count.js
```diff
@@ -0,0 +1,79 @@
...
+
+				const headerRow = node.children[0];
+				const expectedCells = headerRow.children.length;
+
+				for (let i = 1; i < node.children.length; i++) {
+					const currentRow = node.children[i];
+					const actualCells = currentRow.children.length;
```
lumirlumir, line 48:
```suggestion
					const actualCellLength = currentRow.children.length;
```

lumirlumir (changes requested):
Nice rule👍 I've left a few comments.

src/rules/table-column-count.js
SwetaTanwar: Got it, makes sense, applying your suggestion

snitin315 (dismissed):
LGTM, thanks!

Leaving it open for @lumirlumir

nzakas:
@SwetaTanwar please give folks a couple days to circle back to your PR. We have a lot that we're reviewing at any time.

tests/rules/table-column-count.test.js
```diff
@@ -0,0 +1,204 @@
...
+				{
+					messageId: "inconsistentColumnCount",
+					data: { actualCells: "3", expectedCells: "2" },
+					line: 3,
+					column: 13,
+					endLine: 3,
+					endColumn: 22,
+				},
+			],
+		},
```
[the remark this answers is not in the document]
SwetaTanwar: Added

snitin315 (approved):
Thank you!

PR #393, opened 2025-05-29
feat: add checkSiblingsOnly option to no-duplicate-headings rule

### LINKED ISSUE #390: Rule Change: add siblingsOnly option to no-duplicate-headings rule

no-duplicate-headings

Generate fewer warnings

A new option

### Example code

```markdown
# Change log

## 1.0.0

### Features
### Bug Fixes

## 2.0.0

### Features
### Bug Fixes
```

Currently, the no-duplicate-headings rule would flag all duplicate headings at the same level, regardless of their parent heading context. For the example above, it would report:

Duplicate heading "Features" found
Duplicate heading "Bug Fixes" found

With the new siblingsOnly: true option enabled,  the rule will only check for duplicates among headings that share the same immediate parent heading.

For the example changelog:

- No violations will be reported
- Each "Features" heading is allowed because they have different parent headings ("1.0.0" vs "2.0.0")
- Similarly, the "Bug Fixes" headings are allowed

### Participation

### Additional comments

Prior Art: [MD024](https://github.com/DavidAnson/markdownlint/blob/main/doc/md024.md)

xbinaryx:
#### Prerequisites checklist

This PR adds a new siblingsOnly option to the no-duplicate-headings rule to make it more flexible. Currently, the rule flags any duplicate headings in the entire document, which can be too restrictive in some cases. With this new option, users can configure the rule to only check for duplicate headings among siblings (headings at the same level under the same parent heading).

Added a new siblingsOnly boolean option to the rule's schema and refactored the implementation to track headings by level, resetting tracking when moving between levels when siblingsOnly is enabled. Updated documentation and added test cases for both ATX and setext-style headings to verify the new behavior.

#### Related Issues

Fixes #390

### DIFF
--- a/src/rules/no-duplicate-headings.js
+++ b/src/rules/no-duplicate-headings.js
@@ -9,7 +9,7 @@
 
 /** @typedef {import("mdast").Heading} HeadingNode */
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ siblingsOnly?: boolean; }]; }>}
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ checkSiblingsOnly?: boolean; }]; }>}
  * NoDuplicateHeadingsRuleDefinition
  */
 
@@ -35,7 +35,7 @@ export default {
 			{
 				type: "object",
 				properties: {
-					siblingsOnly: {
+					checkSiblingsOnly: {
 						type: "boolean",
 					},
 				},
@@ -43,14 +43,16 @@ export default {
 			},
 		],
 
-		defaultOptions: [{ siblingsOnly: false }],
+		defaultOptions: [{ checkSiblingsOnly: false }],
 	},
 
 	create(context) {
-		const [{ siblingsOnly }] = context.options;
+		const [{ checkSiblingsOnly }] = context.options;
 		const { sourceCode } = context;
 
-		const headingsByLevel = [null, []];
+		const headingsByLevel = checkSiblingsOnly
+			? new Array(7).fill(null).map(() => new Set())
+			: [null, new Set()];
 		let lastLevel = 1;
 		let currentLevelHeadings = headingsByLevel[lastLevel];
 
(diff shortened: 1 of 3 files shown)


src/rules/no-duplicate-headings.js
```diff
@@ -29,46 +30,91 @@ export default {
...
+
+			return isSetext
+				? // get only the text from the first line
+					sourceCode.lines[node.position.start.line - 1].trim()
+				: // get the text without the leading # characters
+					sourceCode
+						.getText(node)
+						.slice(node.depth + 1)
+						.trim();
+		}
```
lumirlumir: Could you add logic to handle closed ATX headings and add corresponding test cases?

Currently, `getHeadingText` gives a false negative when using closed-style ATX headings.

For example:

```md
## installation

## installation ##
```

or

```md
## installation

## installation ##########
```

The second one is an invalid case.

https://spec.commonmark.org/0.31.2/#atx-headings

image

src/rules/no-duplicate-headings.js
```diff
@@ -29,46 +30,91 @@ export default {
...
-							.slice(node.depth + 1)
-							.trim();
-
-				if (headings.has(text)) {
+				const headingText = getHeadingText(node);
+
+				if (checkSiblingsOnly) {
+					const currentLevel = node.depth;
+
+					const direction = currentLevel > lastLevel ? 1 : -1;
```
lumirlumir:
```suggestion
					const direction = currentLevel > lastLevel ? 1 : -1;
```

Could you give a name to the magic numbers `1` and `-1`?

Currently, it's somewhat unclear what these numbers are doing.

Maybe we can use a constant like this:

```js
const DOWN = -1;
const UP = 1;
```

Or, we can use more concrete one with symbols:

```js
const DOWN = Symbol('down');
const UP = Symbol('up'):
```

src/rules/no-duplicate-headings.js
```diff
@@ -29,46 +30,91 @@ export default {
...
+					while (lastLevel !== currentLevel) {
+						if (direction > 0) {
```
lumirlumir:
```suggestion
						if (direction > 0) {
```

Like the above comment, could you clarify the conditional statement?

Maybe we can use `if (direction === UP)` or something similar.

src/rules/no-duplicate-headings.js
```diff
@@ -29,46 +30,91 @@ export default {
...
 	},
 
 	create(context) {
-		const headings = new Set();
+		const [{ checkSiblingsOnly }] = context.options;
 		const { sourceCode } = context;
 
+		const headingsByLevel = checkSiblingsOnly
+			? new Array(7).fill(null).map(() => new Set())
+			: [null, new Set()];
```
lumirlumir:
```suggestion
			? new Array(7).fill(null).map(() => new Set())
			: [null, new Set()];
```

Could you use a `Map` instead of an `Array`?

Currently, the leading `null` item in the array is unused and it causes some confusion.

Maybe we can use a `Map` like this:

```js
new Map([
  [1, new Set()],
  [2, new Set()],
  [3, new Set()],
  [4, new Set()],
  [5, new Set()],
  [6, new Set()]
])
```

lumirlumir (changes requested):
Thanks for taking this issue. I've left some comments.

src/rules/no-duplicate-headings.js
xbinaryx: Good catch. Looks like this bug always existed

src/rules/no-duplicate-headings.js
xbinaryx: I updated the code to use a Map instead.

src/rules/no-duplicate-headings.js
xbinaryx: I refactored the code to use a for loop instead. Let me know if it’s clearer now.

nzakas (approved):
LGTM. Would like @lumirlumir to re-review before merging.

PR #398, opened 2025-06-01
feat: add no-reversed-media-syntax rule

### LINKED ISSUE #386: New Rule: no-reversed-links
### Rule details

This rule is triggered when text that appears to be a link is encountered, but where the syntax appears to have been reversed (the [] and () are reversed)

### Example code

```markdown
(Incorrect link syntax)[https://www.example.com/]
```

### Participation

### Additional comments

Reversed links are not rendered as usable links.

xbinaryx:
#### Prerequisites checklist

This PR adds a new rule `no-reversed-media-syntax` to enforce correct link and image syntax in Markdown.

Added the `no-reversed-media-syntax` rule, along with documentation and tests.

#### Related Issues

Fixes #386

### DIFF
--- a/src/rules/no-reversed-media-syntax.js
+++ b/src/rules/no-reversed-media-syntax.js
@@ -31,19 +31,19 @@ const reversedPattern =
 /**
  * Checks if a match is within any of the code spans
  * @param {number} matchIndex The index of the match
- * @param {Array<{start: number, end: number}>} codeSpans Array of code span positions
+ * @param {Array<{startOffset: number, endOffset: number}>} codeSpans Array of code span positions
  * @returns {boolean} True if the match is within a code span
  */
 function isInCodeSpan(matchIndex, codeSpans) {
 	return codeSpans.some(
-		span => matchIndex >= span.start && matchIndex < span.end,
+		span => matchIndex >= span.startOffset && matchIndex < span.endOffset,
 	);
 }
 
 /**
  * Finds all code spans in the paragraph node by traversing its children
  * @param {ParagraphNode} node The paragraph node to search
- * @returns {Array<{start: number, end: number}>} Array of code span positions
+ * @returns {Array<{startOffset: number, endOffset: number}>} Array of code span positions
  */
 function findCodeSpans(node) {
 	const codeSpans = [];
@@ -56,8 +56,8 @@ function findCodeSpans(node) {
 	function traverse(currentNode) {
 		if (currentNode.type === "inlineCode") {
 			codeSpans.push({
-				start: currentNode.position.start.offset,
-				end: currentNode.position.end.offset,
+				startOffset: currentNode.position.start.offset,
+				endOffset: currentNode.position.end.offset,
 			});
 			return;
 		}


tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,278 @@
...
+		"text (text `func()[index]`) text",
+	],
+	invalid: [
+		{
+			code: "()[]",
+			output: "[]()",
+			errors: [
+				{
+					messageId: "reversedSyntax",
+					line: 1,
```
lumirlumir, line 61:
```suggestion
					messageId: "reversedSyntax",
					data: { /* ... */ }
					line: 1,
```

Could you add a `data` field to every other `invalid` test for more precise testing?

src/rules/no-reversed-media-syntax.js
lumirlumir, line 1: Could you take a look at the false negatives in nested `text` nodes?

Currently, nested `text` nodes using emphasis (`*`), strong (`**`), strikethrough (`~`), and inline code blocks using backticks result in false negatives.

```md
(*foo*)[]

(**foo**)[]

(~~foo~~)[]

(`foo`)[]
```

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,278 @@
...
+					column: 23,
+					endLine: 1,
+					endColumn: 33,
+				},
+			],
+		},
+	],
```
lumirlumir, line 417:
```suggestion
		},
		{
			code: "abcd\n  \\a()[]",
			output: "abcd\n  \\a[]()",
			errors: [
				{
					messageId: "reversedSyntax",
					line: 2,
					column: 5,
					endLine: 2,
					endColumn: 9,
				},
			],
		},
	],
```

Could you add a test case for multiline text with indentation and backslashes?

This test case will help prevent common mistakes in Markdown when calculating positions.

lumirlumir (changes requested):
Thanks for your work.

I've left some comments about false negatives and test cases.

src/rules/no-reversed-media-syntax.js
xbinaryx: Hmm these produce 3 text nodes. This should complicate things 😅

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoReversedMediaSyntaxRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const reversedPattern = /(?<!\\)\(((?:\\.|[^()\\])*)\)\[((?:\\.|[^\]\\])*)\]/gu;
```
nzakas: Can you add a comment showing what this regex is meant to match?

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,112 @@
...
+					const matchLength = reversedSyntax.length;
+
+					const {
+						lineOffset: startLineOffset,
+						columnOffset: startColumnOffset,
+					} = findOffsets(text, matchIndex);
+					const {
+						lineOffset: endLineOffset,
+						columnOffset: endColumnOffset,
+					} = findOffsets(text, matchIndex + matchLength);
```
nzakas, line 120: Because the regex only matches single lines, you can just add `matchLength` to `startColumnOffset` to get this location instead of calculating it from scratch.

xbinaryx: The regex pattern can actually match across multiple lines. This would be incorrect in cases where the link spans multiple lines.

docs/rules/no-reversed-media-syntax.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Background
+
+Markdown syntax for links requires the text to be in square brackets `[]` followed by the URL in parentheses `()`. Similarly, images use `![alt](url)`. It's easy to accidentally reverse these brackets, which results in invalid syntax that won't render correctly.
+
+## Rule Details
+
+This rule is triggered when text that appears to be a link or image is encountered, but where the syntax appears to have been reversed (the `[]` and `()` are reversed).
+
+Examples of **incorrect** code for this rule:
```
snitin315:
```suggestion
This rule is triggered when text that appears to be a link or image is encountered, but the syntax seems to have been reversed (the `[]` and `()` are in the wrong order).

Examples of **incorrect** code for this rule:
```

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,382 @@
...
+		"a &rpar; a &lpar; a &rpar;[a]~",
+		"a<pre>&rpar; a &lpar; a &rpar;[a]~</pre>",
+		"\\(foo)[bar]",
+		"(foo\\)[bar]",
+		"(foo)\\[bar]",
+		"(foo)[bar\\]",
+		"text [foo](bar) text [foo](bar) text",
+		"text [foo](bar)[foo](bar) text",
+		"text [foo](bar)[foo](bar)[foo](bar) text",
+		"text (text `func()[index]`) text",
```
snitin315, line 52: Let's add a few more invalid test cases:

`[^1]: (Footnote text)[https://example.com]`

```
(Click here
)[https://example.com]
```

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,382 @@
...
+		dedent`
+		\`\`\`js
+		myObj.getFiles(test)[0];
+		\`\`\`
+		`,
+		"`myobj.getFiles(test)[0]`",
+		"&lpar;reversed&rpar;[link]",
+		"a &rpar; a &lpar; a &rpar;[a]~",
+		"a<pre>&rpar; a &lpar; a &rpar;[a]~</pre>",
+		"\\(foo)[bar]",
```
snitin315, line 45: Why do we consider this valid?

because `\\[foo](bar)` is rendered as link - \\[foo](bar)

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,151 @@
...
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoReversedMediaSyntaxRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/** Matches reversed link/image syntax like (text)[url], ignoring escaped characters like \(text\)[url]. */
+const reversedPattern =
```
snitin315, line 28: We have one false negative:

```md
<!-- eslint markdown/no-reversed-media-syntax: "error" -->

(Test function())[https://docs.com] // No error
```

snitin315: Another observation:

```md
<!-- eslint markdown/no-reversed-media-syntax: "error" -->

(one)[example.com](two)[next.com] // Only two is reported but autofix fixes both one & two


```

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,151 @@
...
+						lineOffset: endLineOffset,
+						columnOffset: endColumnOffset,
+					} = findOffsets(text, matchIndex + matchLength);
+
+					const nodeStartLine = node.position.start.line;
+					const nodeStartColumn = node.position.start.column;
+					const startLine = nodeStartLine + startLineOffset;
+					const endLine = nodeStartLine + endLineOffset;
+					const startColumn =
+						(startLine === nodeStartLine ? nodeStartColumn : 1) +
```
snitin315: Let's avoid magic number if possible.
```suggestion
						(startLine === nodeStartLine ? nodeStartColumn : BASE_COLUMN) +
```

tests/rules/no-reversed-media-syntax.test.js
xbinaryx: This is actually `\(foo)[bar]`. The first \ escapes the second \ so that the actual string contains a single backslash.

src/rules/no-reversed-media-syntax.js
xbinaryx: This is because when two is fixed, one becomes an error so the autofix fixes both.

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,154 @@
...
+			recommended: true,
+			description: "Disallow reversed link and image syntax",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-reversed-media-syntax.md",
+		},
+
+		fixable: "code",
+
+		messages: {
+			reversedSyntax:
+				"Unexpected reversed syntax found. Use [{{label}}]({{url}}) instead.",
```
Tanujkanti4441: Some URLs can be very long i think that would make this error message too long, we can try this:
`"Unexpected reversed syntax found. Use [label](URL) syntax instead."`.

Also, for both the link `[]()` and image syntax ``, we are suggesting the same syntax in message `[]()`. I feel we can suggest their respective syntax?


What do you think?

xbinaryx: > Some URLs can be very long i think that would make this error message too long, we can try this:
"Unexpected reversed syntax found. Use [label](URL) syntax instead.".

Done.

> Also, for both the link []() and image syntax , we are suggesting the same syntax in message [](). I feel we can suggest their respective syntax?

The rule is specifically focused on detecting and fixing the reversed ()[] syntax pattern, which is the same for both links and images. Since we're only highlighting the incorrect pattern and not the specific use case, I think a single error message showing the standard []() syntax is sufficient. What do you think?

Tanujkanti4441: Makes sense to me.

nzakas (approved):
LGTM. Would like another review before merging.

tests/rules/no-reversed-media-syntax.test.js
lumirlumir: @snitin315

I believe this behavior is working as intended. A similar case was previously discussed as a bug in another rule.

Reference: https://github.com/eslint/markdown/pull/348

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,403 @@
...
+					column: 1,
+					endLine: 1,
+					endColumn: 12,
+				},
+			],
+		},
+		{
+			code: "(~~foo~~)[]",
+			output: "[~~foo~~]()",
+			errors: [
```
lumirlumir:
```suggestion
			output: "[~~foo~~]()",			
			language: "markdown/gfm",
			errors: [
```


Could you add `markdown/gfm`, since strikethrough only works in GFM mode?

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -0,0 +1,403 @@
...
+					column: 1,
+					endLine: 1,
+					endColumn: 10,
+				},
+			],
+		},
+		{
+			code: "[^1]: (Footnote text)[https://example.com]",
+			output: "[^1]: [Footnote text](https://example.com)",
+			errors: [
```
lumirlumir:
```suggestion
			output: "[^1]: [Footnote text](https://example.com)",			
			language: "markdown/gfm",
			errors: [
```


Same here.

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,150 @@
...
+
+/** Matches reversed link/image syntax like (text)[url], ignoring escaped characters like \(text\)[url]. */
+const reversedPattern =
+	/(?<!\\)\(((?:\\.|[^()\\]|\([\s\S]*\))*)\)\[((?:\\.|[^\]\\\n])*)\](?!\()/gu;
+const codeSpanPattern = /(?<!\\)`+[^`]*`+/gu;
+
+/**
+ * Checks if a match is within any of the code spans
+ * @param {number} matchIndex The index of the match
+ * @param {Array<{start: number, end: number}>} codeSpans Array of code span positions
```
lumirlumir:
```suggestion
 * @param {Array<Position>} codeSpans Array of code span positions
```

Could you use the `Position` type from `unist` instead?

```js
/** @typedef {import("unist").Position} Position */
```

Currently, the position format doesn't follow the Mdast (Markdown AST) convention, which can cause some confusions.

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,150 @@
...
+function isInCodeSpan(matchIndex, codeSpans) {
+	return codeSpans.some(
+		span => matchIndex >= span.start && matchIndex < span.end,
+	);
+}
+
+/**
+ * Finds all code spans in the text
+ * @param {string} text The text to search
+ * @returns {Array<{start: number, end: number}>} Array of code span positions
```
lumirlumir:
```suggestion
 * @returns {Array<Position>} Array of code span positions
```

Same here.

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/** Matches reversed link/image syntax like (text)[url], ignoring escaped characters like \(text\)[url]. */
+const reversedPattern =
+	/(?<!\\)\(((?:\\.|[^()\\]|\([\s\S]*\))*)\)\[((?:\\.|[^\]\\\n])*)\](?!\()/gu;
+const codeSpanPattern = /(?<!\\)`+[^`]*`+/gu;
```
lumirlumir:
```suggestion
const codeSpanPattern = /(?<!\\)`+[^`]*`+/gu;
```

Could you take a look at this regex? I think this regex is quite fragile.

image

For example, if we add a test case like the one below, it doesn't throw an error. (There might be other fragile cases as well — this is just one example.)

image

---

If our goal is to find the position of every inline code block, it might be more reliable to collect all `InlineCode` node's positions by traversing the children of a specific `Paragraph` node.

In my experiences, regex doesn’t work well when we need accurate position information for a specific node. 😅

lumirlumir (changes requested):

src/rules/no-reversed-media-syntax.js
xbinaryx: We only use the start and end offsets in isInCodeSpan. Do you prefer I return the whole position instead?

lumirlumir:
```suggestion
 * @param {Array<{startOffset: number, endOffset: number}>} codeSpans Array of code span positions
```

@xbinaryx Thinking again based on your comments — instead of using the `Position` type, could you rename `start` and `end` to `startOffset` and `endOffset` for better clarity?

In Mdast, `position.start` and `position.end` include `line`, `column`, and `offset`, so using just `start` and `end` might be confusing I think.

image

---

Others are looks good to me 👍

src/rules/no-reversed-media-syntax.js
```diff
@@ -0,0 +1,163 @@
...
+function isInCodeSpan(matchIndex, codeSpans) {
+	return codeSpans.some(
+		span => matchIndex >= span.start && matchIndex < span.end,
+	);
+}
+
+/**
+ * Finds all code spans in the paragraph node by traversing its children
+ * @param {ParagraphNode} node The paragraph node to search
+ * @returns {Array<{start: number, end: number}>} Array of code span positions
```
lumirlumir:
```suggestion
 * @returns {Array<{startOffset: number, endOffset: number}>} Array of code span positions
```

lumirlumir (approved):
LGTM, Thanks!

Would like @nzakas and @snitin315 to verify the changes before merging.

nzakas (approved):
LGTM. Going to merge so we can do a release.

PR #403, opened 2025-06-02
feat: add no-space-in-emphasis rule

### LINKED ISSUE #378: New Rule: no-space-in-emphasis
### Rule details

This rule is triggered when emphasis markers (bold, italic) are used, but they have spaces between the markers and the text.

### Example code

```markdown
Here is some ** bold ** text.

Here is some * italic * text.

Here is some more __ bold __ text.

Here is some more _ italic _ text.
```

### Participation

### Additional comments

Rationale: Emphasis is only parsed as such when the asterisks/underscores aren't surrounded by spaces. This rule attempts to detect where they were surrounded by spaces, but it appears that emphasized text was intended by the author.

Pixel998:
#### Prerequisites checklist

This PR implements the no-space-in-emphasis rule to disallow spaces around emphasis markers.

Implemented the no-space-in-emphasis rule, along with documentation and tests.

#### Related Issues

Fixes #378

### DIFF
--- a/docs/rules/no-space-in-emphasis.md
+++ b/docs/rules/no-space-in-emphasis.md
@@ -22,6 +22,7 @@ Here is some * italic * text.
 Here is some __ bold __ text.
 Here is some _ italic _ text.
 Here is some *** bold italic *** text.
+Here is some ___ bold italic ___ text.
 ```
 
 Examples of **correct** code for this rule:
@@ -34,6 +35,7 @@ Here is some *italic* text.
 Here is some __bold__ text.
 Here is some _italic_ text.
 Here is some ***bold italic*** text.
+Here is some ___bold italic___ text.
 Here is some **_ bold _** text.
 Here is some _** italic **_ text.
 ```


snitin315:
@Pixel998 Make sure to mark it as ready for review once done, the team doesn't pay attention to draft PRs.

docs/rules/no-space-in-emphasis.md
```diff
@@ -0,0 +1,43 @@
+# no-space-in-emphasis
+
+Disallow spaces around emphasis markers.
+
+## Background
+
+In Markdown, emphasis (bold and italic) is created using asterisks (`*`) or underscores (`_`), and strikethrough is created using tildes (`~`). The emphasis markers must be directly adjacent to the text they're emphasizing, with no spaces between the markers and the text. When spaces are present, the emphasis is not rendered correctly.
```
snitin315:
```suggestion
In Markdown, emphasis (bold and italic) is created using asterisks (`*`) or underscores (`_`), and a strikethrough is created using tildes (`~`). The emphasis markers must be directly adjacent to the text they're emphasizing, with no spaces between the markers and the text. When spaces are present, the emphasis is not rendered correctly.
```

tests/rules/no-space-in-emphasis.test.js
```diff
@@ -0,0 +1,1025 @@
...
+			errors: [
+				{
+					messageId: "spaceInEmphasis",
+					line: 1,
+					column: 8,
+					endLine: 1,
+					endColumn: 20,
+				},
+			],
+		},
```
snitin315: Markdownlint reports two errors in this case, separate for both prefix & suffix highlighting exactly where the whitespace is instead of highlighting the whole text. Should we do the same? It's better in case of long paragraphs.

Screenshot 2025-06-23 at 2 41 54 PM

snitin315:
Following is an edge case, I'm not sure if we want to solve for it in the initial version.

```markdown
**** bold **** // No error, 
```

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,145 @@
...
+import { findOffsets } from "../util.js";
+
+//-----------------------------------------------------------------------------
+// Type Definitions
+//-----------------------------------------------------------------------------
+
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoSpaceInEmphasisRuleDefinition
+ */
```
lumirlumir:
```suggestion
/**
 * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
 * NoSpaceInEmphasisRuleDefinition
 */
```

image

tests/rules/no-space-in-emphasis.test.js
Pixel998: Done

docs/rules/no-unused-definitions.md
```diff
@@ -1,103 +1,103 @@
-# no-unused-definitions
```
lumirlumir: I think this file was modified because it uses CRLF line endings instead of LF.

Can we create a separate PR for that? Changing this file here feels out of scope.

docs/rules/no-space-in-emphasis.md
```diff
@@ -0,0 +1,43 @@
...
+
+```markdown
+<!-- eslint markdown/no-space-in-emphasis: "error" -->
+
+Here is some **bold** text.
+Here is some *italic* text.
+Here is some __bold__ text.
+Here is some _italic_ text.
+Here is some ~~strikethrough~~ text.
+```
```
lumirlumir:
````suggestion
```markdown
<!-- eslint markdown/no-space-in-emphasis: "error" -->

Here is some ** bold ** text.
Here is some * italic * text.
Here is some __ bold __ text.
Here is some _ italic _ text.
Here is some *** bold italic *** text.
Here is some _** bold italic **_ text.
Here is some **_ bold italic _** text.
Here is some ~ strikethrough ~ text.
Here is some ~~ strikethrough ~~ text.
```

Examples of **correct** code for this rule:

```markdown
<!-- eslint markdown/no-space-in-emphasis: "error" -->

Here is some **bold** text.
Here is some *italic* text.
Here is some __bold__ text.
Here is some _italic_ text.
Here is some ***bold italic*** text.
Here is some _**bold italic**_ text.
Here is some **_bold italic_** text.
Here is some ~strikethrough~ text.
Here is some ~~strikethrough~~ text.
```
````

Could you add some more cases?

File: docs/rules/no-space-in-emphasis.md [hunk 6]
```diff
@@ -0,0 +1,43 @@
+# no-space-in-emphasis
+
+Disallow spaces around emphasis markers.
+
+## Background
+
+In Markdown, emphasis (bold and italic) is created using asterisks (`*`) or underscores (`_`), and a strikethrough is created using tildes (`~`). The emphasis markers must be directly adjacent to the text they're emphasizing, with no spaces between the markers and the text. When spaces are present, the emphasis is not rendered correctly.
+
+## Rule Details
+
```
lumirlumir, line 12: Could you add a note that strikethrough is detected only when the `language` mode is `markdown/gfm`? Here is a reference:

https://github.com/eslint/markdown/blob/bc627c81317b80f2b1722b4448d429293d98ddf9/docs/rules/no-duplicate-definitions.md?plain=1#L18-L20

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,228 @@
...
+ * @typedef {[]} NoSpaceInEmphasisOptions
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoSpaceInEmphasisOptions, MessageIds: NoSpaceInEmphasisMessageIds }>} NoSpaceInEmphasisRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const markerPattern = /(?<!\\)(\*\*\*|\*\*|\*|___|__|_|~~|~)/gu;
+const whitespacePattern = /\s/u;
```
lumirlumir: Could you use a more specific whitespace pattern that matches only space (` `) and tab (`\t`) characters? 

According to the CommonMark spec, using `\s` can cause incorrect results when NBSP (U+00A0) or similar characters appear. 

Adding additional test cases for this would be helpful as well.

Ref: https://github.com/eslint/markdown/pull/421

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,228 @@
...
+const emphasisTypes = new Set(["emphasis", "strong", "delete"]);
+
+/**
+ * Finds all emphasis markers in the text
+ * @param {string} text The text to search
+ * @returns {Array<{marker: string, start: number, end: number}>} Array of emphasis markers
```
lumirlumir:
```suggestion
 * @returns {Array<{marker: string, startIndex: number, endIndex: number}>} Array of emphasis markers
```

Could we use a more specific name here? Adding `index` or `offset` would make sense.

Since the Markdown AST already provides `start` and `end` objects with multiple properties, simply using `start` or `end` could confuse users.

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,228 @@
...
+ */
+function findEmphasisMarkers(text) {
+	const markers = [];
+	let match;
+
+	while ((match = markerPattern.exec(text)) !== null) {
+		markers.push({
+			marker: match[1],
+			start: match.index,
+			end: match.index + match[1].length,
```
lumirlumir:
```suggestion
			startIndex: match.index,
			endIndex: match.index + match[1].length,
```

Same here.

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,228 @@
...
+const emphasisTypes = new Set(["emphasis", "strong", "delete"]);
+
+/**
+ * Finds all emphasis markers in the text
+ * @param {string} text The text to search
+ * @returns {Array<{marker: string, start: number, end: number}>} Array of emphasis markers
+ */
+function findEmphasisMarkers(text) {
+	const markers = [];
+	let match;
```
lumirlumir: Could we add explicit types for these variables? Right now they’re typed as `any`.

tests/rules/no-space-in-emphasis.test.js
```diff
@@ -0,0 +1,1234 @@
...
+					column: 16,
+					endLine: 1,
+					endColumn: 20,
+				},
+			],
+		},
+		{
+			code: "Broken ~ strikethrough ~ with spaces",
+			output: "Broken ~strikethrough~ with spaces",
+			errors: [
```
lumirlumir:
```suggestion
			output: "Broken ~strikethrough~ with spaces",
			language: "markdown/commonmark",
			errors: [
```

I think we need to consider further how to handle strikethrough in CommonMark mode.

Currently, the strikethrough syntax still shows an error even when CommonMark mode is enabled.

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,228 @@
...
+ * @typedef {"spaceInEmphasis"} NoSpaceInEmphasisMessageIds
+ * @typedef {[]} NoSpaceInEmphasisOptions
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoSpaceInEmphasisOptions, MessageIds: NoSpaceInEmphasisMessageIds }>} NoSpaceInEmphasisRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const markerPattern = /(?<!\\)(\*\*\*|\*\*|\*|___|__|_|~~|~)/gu;
```
lumirlumir: I think we need to think more about the `markerPattern`.

If markers appear in combinations like `*__` or `__*`, the current rule doesn’t catch them. (The same applies to combinations involving strikethrough.)

For example, the test case below doesn’t throw an error, so the test fails:

```js
		{
			code: "Broken *__ strong and emphasis __* with spaces",
			output: "Broken *__strong and emphasis__* with spaces",
			errors: [
				{
					messageId: "spaceInEmphasis",
					line: 1,
					column: 8,
					endLine: 1,
					endColumn: 13,
				},
				{
					messageId: "spaceInEmphasis",
					line: 1,
					column: 30,
					endLine: 1,
					endColumn: 35,
				},
			],
		},
```

nzakas:
@Pixel998 are you still working on this?

src/rules/no-space-in-emphasis.js
Pixel998: This is working as intended and matches Markdownlint's behavior. I've left a note about this.

tests/rules/no-space-in-emphasis.test.js
Pixel998: I’m not sure there’s a way around this at the moment, but let me know if you have any ideas.

docs/rules/no-space-in-emphasis.md
Pixel998: Let’s address your other comment before moving forward.

tests/rules/no-space-in-emphasis.test.js
lumirlumir: I can think of three possible ways to handle this situation:

1. Remove support for strikethrough syntax:  

    This approach is consistent with Markdownlint, which currently does not account for strikethrough syntax.

    image

3. Add an optional setting for strikethrough handling:  

    Users who want this behavior can enable it as needed.

5. Keep the current implementation and document it as a limitation:  

    This is another option, though I'm not sure it's the best solution.

---

Since ESLint's current implementation doesn't allow rules to check whether commonmark or GFM mode is being used, I think one of these solutions could be reasonable.

I'd appreciate any feedback or suggestions from @eslint/eslint-team before moving forward.

docs/rules/no-space-in-emphasis.md
lumirlumir: Pending until https://github.com/eslint/markdown/pull/403#discussion_r2218418155 is addressed.

src/rules/no-space-in-emphasis.js
lumirlumir: Is there a reason the test isn't passing on my end? Did I miss something?

image

image

---

Or did you mean that not checking `*__` and `__*` is intentional?

src/rules/no-space-in-emphasis.js
lumirlumir, line 1: Currently, we're seeing false negatives on `Heading` and `TableCell` (GFM) nodes.

It would make sense to add `Heading` and `TableCell` to the visitors.

For example:

Image

Image

This behavior aligns with Markdownlint:

Image

Image

lumirlumir (changes requested):

tests/rules/no-space-in-emphasis.test.js
Pixel998: I think an optional setting for strikethrough handling is the most flexible approach. It preserves CommonMark compatibility by default while allowing users using GFM to explicitly enable support. This avoids the rigidity of removing the feature entirely.

src/rules/no-space-in-emphasis.js
Pixel998: Yes, it is intentional. We skip checking `__` since they're inside other emphasis markers (`* *`). I added a note about this—let me know if it's still unclear.

tests/rules/no-space-in-emphasis.test.js
nzakas: I think adding an option is a good idea. :+1:

lumirlumir: Let’s go with the second approach

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,284 @@
...
+						});
+					}
+				}
+			}
+		}
+
+		return {
+			heading: checkEmphasis,
+			paragraph: checkEmphasis,
+			tableCell: checkEmphasis,
```
lumirlumir:
```suggestion
			heading(node) {
				checkEmphasis(node);
			},

			paragraph(node) {
				checkEmphasis(node);
			},

			tableCell(node) {
				checkEmphasis(node);
			},
```

Would it be possible to use the pattern above for consistency across the rules?

Reference:

https://github.com/eslint/markdown/blob/1fdbd62fe62e2c14bfaf53100c6abf7da85edfeb/src/rules/no-bare-urls.js#L157-L167

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,284 @@
...
+
+		messages: {
+			spaceInEmphasis: "Unexpected space around emphasis marker.",
+		},
+
+		schema: [
+			{
+				type: "object",
+				properties: {
+					includeStrikethrough: {
```
lumirlumir:
```suggestion
					checkStrikethrough: {
```

I think `checkStrikethrough` is a more appropriate option name, since many of the rules use `check` for boolean option names.

Here are some references:

https://github.com/eslint/markdown/blob/614ef3e0bf539655170f202467119fbbbf8963f6/src/rules/no-duplicate-headings.js#L64-L66

https://github.com/eslint/markdown/blob/614ef3e0bf539655170f202467119fbbbf8963f6/src/rules/no-empty-definitions.js#L73-L75

https://github.com/eslint/markdown/blob/614ef3e0bf539655170f202467119fbbbf8963f6/src/rules/table-column-count.js#L44-L46

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,284 @@
...
+						type: "boolean",
+					},
+				},
+				additionalProperties: false,
+			},
+		],
+
+		defaultOptions: [
+			{
+				includeStrikethrough: false,
```
lumirlumir:
```suggestion
				checkStrikethrough: false,
```

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,284 @@
...
+		const [{ includeStrikethrough }] = context.options;
+		const markerPattern = createMarkerPattern(includeStrikethrough);
+		const emphasisTypes = createEmphasisTypes(includeStrikethrough);
+
+		/**
+		 * Extracts text from a node, replacing emphasis and HTML nodes with whitespace
+		 * @param {Heading|Paragraph|TableCell} node The node to extract text from
+		 * @returns {string} The extracted text with certain nodes replaced by whitespace
+		 */
+		function extractText(node) {
```
lumirlumir: The current implementation traverses the AST within the specific node, which is inefficient since ESLint already traverses the AST.

It might be better to revisit this approach and use traversal methods instead of performing an additional traversal.

For example, you could do:

- `heading(node)` – mark when entering a heading
- `"heading *"(node)` – capture all descendant nodes of the heading
- `heading:exit(node)` – check the heading once all pieces are known

This way, you can leverage the existing traversal and avoid re-traversing the same node.

The following PR could serve as a good reference for this:
https://github.com/eslint/markdown/pull/498/files

docs/rules/no-space-in-emphasis.md
```diff
@@ -0,0 +1,71 @@
...
+Here is some _italic_ text.
+Here is some ***bold italic*** text.
+Here is some **_ bold _** text.
+Here is some _** italic **_ text.
+```
+
+## Options
+
+The following options are available on this rule:
+
```
lumirlumir, line 46: Could we add a description noting that this option is only useful when using GFM (`language: "markdown/gfm"`) mode?

src/rules/no-space-in-emphasis.js
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * Creates a marker pattern based on whether strikethrough should be included
+ * @param {boolean} includeStrikethrough Whether to include strikethrough markers
+ * @returns {RegExp} The marker pattern
+ */
+function createMarkerPattern(includeStrikethrough) {
+	if (includeStrikethrough) {
+		return /(?<!\\)(\*\*\*|\*\*|\*|___|__|_|~~|~)/gu;
+	}
+	return /(?<!\\)(\*\*\*|\*\*|\*|___|__|_)/gu;
```
lumirlumir:
```suggestion
	if (includeStrikethrough) {
		return /(?<=(?<!\\)(?:\\{2})*)(?<marker>\*\*\*|\*\*|\*|___|__|_|~~|~)/gu;
	}
	return /(?<=(?<!\\)(?:\\{2})*)(?<marker>\*\*\*|\*\*|\*|___|__|_)/gu;
```

The current logic for detecting backslashes is a bit fragile, especially when there are multiple backslashes before the regex.

I’d like to suggest using the pattern above for more reliable detection. It would also be great to add some test cases for it.

You can also reference the ongoing PR here:
https://github.com/eslint/markdown/pull/490/files

---

I would also like to suggest using named capture groups for better readability.

lumirlumir (changes requested):

docs/rules/no-space-in-emphasis.md
```diff
@@ -0,0 +1,75 @@
...
+Examples of **incorrect** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-space-in-emphasis: "error" -->
+
+Here is some ** bold ** text.
+Here is some * italic * text.
+Here is some __ bold __ text.
+Here is some _ italic _ text.
+Here is some *** bold italic *** text.
```
lumirlumir, line 24:
```suggestion
Here is some *** bold italic *** text.
Here is some ___ bold italic ___ text.
```

Could we also add this example here as well?

docs/rules/no-space-in-emphasis.md
```diff
@@ -0,0 +1,75 @@
...
+Examples of **correct** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-space-in-emphasis: "error" -->
+
+Here is some **bold** text.
+Here is some *italic* text.
+Here is some __bold__ text.
+Here is some _italic_ text.
+Here is some ***bold italic*** text.
```
lumirlumir, line 37:
```suggestion
Here is some ***bold italic*** text.
Here is some ___bold italic___ text.
```

Same here.

lumirlumir (dismissed):
I've left a few minor documentation suggestions. otherwise, LGTM.

PR #412, opened 2025-06-10
chore: remove `@types/eslint` and bump mocha

lumirlumir:
#### Prerequisites checklist

In this PR, I've removed `@types/eslint` and bumped the version of `mocha`.

As far as I know, `eslint` now includes types based on `@types/eslint`, so this type package is unnecessary as long as we use the latest `eslint`.

Ref: https://github.com/eslint/eslint/blob/main/lib/types/index.d.ts#L2-L4

I've also bumped the version of Mocha to v11, which has no breaking changes for us, since it only upgrades the Node.js version requirement, which is compatible with `@eslint/markdown`.

Ref: https://github.com/mochajs/mocha/releases/tag/v11.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@eslint/js": "^9.4.0",
     "c8": "^10.1.2",
     "dedent": "^1.5.3",
-    "eslint": "^9.25.1",
+    "eslint": "^9.28.0",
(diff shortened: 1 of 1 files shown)


nzakas (approved):
LGTM. Thanks.

PR #411, opened 2025-06-09
feat: add JSON frontmatter support

Pixel998:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What is the purpose of this pull request?

To add JSON frontmatter support using the VitePress-style format (--- delimiters) 

#### What changes did you make? (Give an overview)

- Added JSON frontmatter parser configuration using --- delimiters
- Added tests
- Added types

#### Related Issues

Closes #404

PR #413, opened 2025-06-10
feat: update `no-multiple-h1` rule to recognize JSON frontmatter

lumirlumir:
#### Prerequisites checklist

In this PR, I've added support for the JSON front matter feature to the `no-multiple-h1` rule, which was newly added in #411.

I've updated the regex pattern to correctly handle the JSON front matter `title` and some edge cases.

#### Related Issues

Prerequisite: https://github.com/eslint/markdown/pull/411

I tried to update the regex as accurately as possible, but there might be some edge cases I’ve missed.

It would be nice if someone could find more—excluding those that cause parsing errors. (I didn’t include cases that cause parsing errors.)


nzakas:
@Pixel998 While I can appreciate what you're saying, we're going to stick with this PR. Thanks for your understanding.

lumirlumir:
Thank you all for clarifying and understanding the situation. I truly realized how important it is to raise an issue and have a conversation before proceeding with specific features. I'll make sure to avoid causing confusion next time.

tests/rules/no-multiple-h1.test.js
```diff
@@ -60,6 +64,56 @@ ruleTester.run("no-multiple-h1", rule, {
 			# Not heading
 			</p>
 		`,
+		{
+			code: dedent`
+			    ---
+				[ "title" ]
+				---
```
lumirlumir, line 71: image

The formatting of the code is consistent in the file view, but the GitHub diff displays it somewhat incorrectly.

PR #417, opened 2025-06-12
chore: replace `npm-prepare` script with npm workspace feature

lumirlumir:
#### Prerequisites checklist

Working in progress.

### DIFF
--- a/npm-prepare.cjs
+++ b/npm-prepare.cjs
@@ -31,3 +31,5 @@ if (!process.env.NO_RECURSIVE_PREPARE) {
 		});
 	}
 }
+
+// TODO: Remove


examples/typescript/package.json
```diff
@@ -1,12 +1,13 @@
 {
   "private": true,
+  "name": "example-typescript",
```
lumirlumir, line 3: The `name` field is necessary. If we omit this field, `npm` cannot hoist packages correctly, which results in missing dependencies after installation.

For example, if we omit this field, `tsc` is not recognized as expected.

examples/typescript/package.json
```diff
@@ -1,12 +1,13 @@
...
   "scripts": {
     "test": "eslint ."
   },
   "devDependencies": {
     "@eslint/js": "^9.7.0",
     "eslint": "^9.15.0",
-    "typescript": "^5.3.3",
+    "typescript": "^5.8.3",
```
lumirlumir, line 10: I've updated the TypeScript version to match the version in the root.

nzakas (dismissed):
I checked this out and tried locally, and everything worked. Nice job!

lumirlumir:
@fasttime Thanks for the suggestion.

I've added some more detailed description to the example, including the point you mentioned. 👍

New commit: https://github.com/eslint/markdown/pull/417/commits/ee91e059bfeb94bca7a3dd0cd66e2975c0541d6c

PR #418, opened 2025-06-12
feat: add no-bare-urls rule

### LINKED ISSUE #399: New Rule: no-literal-urls
### Rule details

This rule warns when GFM autolink literals are used.

### Example code

```markdown
For more info, visit https://www.example.com/ or email user@example.com.
```

### Prior Art

[remark-lint-no-literal-urls](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-literal-urls)
[MD034](https://github.com/DavidAnson/markdownlint/blob/main/doc/md034.md)

### Participation

### Additional comments

GFM autolink literals (just a raw URL) are a feature enabled by GFM. They don’t work everywhere. So, it’s recommended to instead use regular autolinks (`<https://url>`) or full links (`[text](url)`).

xbinaryx:
#### Prerequisites checklist

This PR implements the `no-bare-urls` rule to enforce proper URL formatting in Markdown documents. The rule ensures URLs are either wrapped in angle brackets (autolinks) or used with text (links), rather than being used as bare URLs.

Implemented the `no-bare-urls` rule, along with documentation and tests.

#### Related Issues

Fixes #399

### DIFF
--- a/docs/rules/no-bare-urls.md
+++ b/docs/rules/no-bare-urls.md
@@ -15,7 +15,7 @@ A bare URL appears without angle brackets (`<>`), such as `https://example.com/`
 
 > [!IMPORTANT] <!-- eslint-disable-line -- This should be fixed in https://github.com/eslint/markdown/issues/294 -->
 >
-> This rule relies on bare URLs being link nodes, which requires a GFM-compatible parser (e.g., `language: "markdown/gfm"`).
+> This rule requires `language: "markdown/gfm"`.
 
 This rule flags bare URLs that should be formatted as autolinks or links.
 
@@ -27,6 +27,10 @@ Examples of **incorrect** code for this rule:
 For more info, visit https://www.example.com/
 
 Contact us at user@example.com
+
+[Read the [docs]](https://www.example.com/)
+
+[docs]: https://www.example.com/docs
 ```
 
 Examples of **correct** code for this rule:
@@ -41,6 +45,12 @@ For more info, visit [Example Website](https://www.example.com/)
 Contact us at <user@example.com>
 
 Contact us at [user@example.com](mailto:user@example.com)
+
+Not a clickable link: `https://www.example.com/`
+
+[https://www.example.com/]
+
+[Read the \[docs\]](https://www.example.com/)
 ```
 
 ## When Not to Use It
--- a/docs/rules/no-bare-urls.md
+++ b/docs/rules/no-bare-urls.md
@@ -1,0 +1,53 @@
+# no-bare-urls
+
+Disallow bare URLs.
+
+## Background
+
+In Markdown, URLs are typically formatted in two ways:
+
+1. Autolinks: URLs wrapped in angle brackets, such as `<https://example.com/>`
+2. Links: URLs with descriptive text, such as `[Visit our website](https://example.com/)`
+
+A bare URL appears without angle brackets (`<>`), such as `https://example.com/`. While GitHub Flavored Markdown (GFM) allows bare URLs to be recognized as clickable links, this feature is not supported in all Markdown processors. To ensure compatibility across different processors, use autolinks or links instead.
+
+## Rule Details
+
+> [!IMPORTANT] <!-- eslint-disable-line -- This should be fixed in https://github.com/eslint/markdown/issues/294 -->
+>
+> This rule relies on bare URLs being link nodes, which requires a GFM-compatible parser (e.g., `language: "markdown/gfm"`).
+
+This rule flags bare URLs that should be formatted as autolinks or links.
+
+Examples of **incorrect** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-bare-urls: "error" -->
+
+For more info, visit https://www.example.com/
+
+Contact us at user@example.com
+```
+
+Examples of **correct** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-bare-urls: "error" -->
+
+For more info, visit <https://www.example.com/>
+
+For more info, visit [Example Website](https://www.example.com/)
+
+Contact us at <user@example.com>
+
+Contact us at [user@example.com](mailto:user@example.com)
+```
+
+## When Not to Use It
+
+If you're working in an environment where GFM autolink literals are fully supported and you prefer their simplicity, you can safely disable this rule.
+
+## Prior Art
+
+* [remark-lint-no-literal-urls](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-literal-urls)
+* [MD034 - Bare URL used](https://github.com/DavidAnson/markdownlint/blob/main/doc/md034.md)
--- a/docs/rules/no-bare-urls.md
+++ b/docs/rules/no-bare-urls.md
@@ -28,6 +28,8 @@ For more info, visit https://www.example.com/
 
 Contact us at user@example.com
 
+# https://www.example.com/
+
 [Read the [docs]](https://www.example.com/)
 
 [docs]: https://www.example.com/docs
@@ -45,6 +47,8 @@ For more info, visit [Example Website](https://www.example.com/)
 Contact us at <user@example.com>
 
 Contact us at [user@example.com](mailto:user@example.com)
+
+# <https://www.example.com/>
 
 Not a clickable link: `https://www.example.com/`
 
--- a/src/rules/no-bare-urls.js
+++ b/src/rules/no-bare-urls.js
@@ -7,8 +7,11 @@
 // Type Definitions
 //-----------------------------------------------------------------------------
 
+/** @typedef {import("mdast").Node} Node */
 /** @typedef {import("mdast").Paragraph} ParagraphNode */
+/** @typedef {import("mdast").Heading} HeadingNode */
 /** @typedef {import("mdast").TableCell} TableCellNode */
+/** @typedef {import("mdast").Link} LinkNode */
 /**
  * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
  * NoBareUrlsRuleDefinition
@@ -18,7 +21,7 @@
 // Helpers
 //-----------------------------------------------------------------------------
 
-const htmlTagPattern = /^<([^!>][^/\s>]*)/u;
+const htmlTagNamePattern = /^<([^!>][^/\s>]*)/u;
 
 /**
  * Parses an HTML tag to extract its name and closing status
@@ -26,7 +29,7 @@ const htmlTagPattern = /^<([^!>][^/\s>]*)/u;
  * @returns {{ name: string; isClosing: boolean; } | null} Object containing tag name and closing status, or null if not a valid tag
  */
 function parseHtmlTag(tagText) {
-	const match = tagText.match(htmlTagPattern);
+	const match = tagText.match(htmlTagNamePattern);
 	if (match) {
 		const tagName = match[1].toLowerCase();
 		const isClosing = tagName.startsWith("/");
@@ -64,56 +67,79 @@ export default {
 
 	create(context) {
 		const { sourceCode } = context;
+		/** @type {Array<LinkNode>} */
 		const bareUrls = [];
 
 		/**
 		 * Finds bare URLs in markdown nodes while handling HTML tags.
 		 * When an HTML tag is found, it looks for its closing tag and skips all nodes
 		 * between them to prevent checking for bare URLs inside HTML content.
-		 * @param {ParagraphNode|TableCellNode} node The node to process
+		 * @param {ParagraphNode|HeadingNode|TableCellNode} node The node to process
 		 * @returns {void}
 		 */
 		function findBareUrls(node) {
-			if (!node.children) {
-				return;
-			}
+			/**
+			 * Recursively traverses the AST to find bare URLs, skipping over HTML blocks.
+			 * @param {Node} currentNode The current AST node being traversed.
+			 * @returns {void}
+			 */
+			function traverse(currentNode) {
+				if (
+					"children" in currentNode &&
+					Array.isArray(currentNode.children)
+				) {
+					for (let i = 0; i < currentNode.children.length; i++) {
+						const child = currentNode.children[i];
 
-			for (let i = 0; i < node.children.length; i++) {
-				const child = node.children[i];
+						if (child.type === "html") {
+							const tagInfo = parseHtmlTag(
+								sourceCode.getText(child),
+							);
 
-				if (child.type === "html") {
-					const tagInfo = parseHtmlTag(sourceCode.getText(child));
-
-					if (tagInfo && !tagInfo.isClosing) {
-						for (let j = i + 1; j < node.children.length; j++) {
-							const nextChild = node.children[j];
-							if (nextChild.type === "html") {
-								const closingTagInfo = parseHtmlTag(
-									sourceCode.getText(nextChild),
-								);
-								if (
-									closingTagInfo?.name === tagInfo.name &&
-									closingTagInfo?.isClosing
+							if (tagInfo && !tagInfo.isClosing) {
+								for (
+									let j = i + 1;
+									j < currentNode.children.length;
+									j++
 								) {
-									i = j;
-									break;
+									const nextChild = currentNode.children[j];
+									if (nextChild.type === "html") {
+										const closingTagInfo = parseHtmlTag(
+											sourceCode.getText(nextChild),
+										);
+										if (
+											closingTagInfo?.name ===
+												tagInfo.name &&
+											closingTagInfo?.isClosing
+										) {
+											i = j;
+											break;
+										}
+									}
 								}
+								continue;
 							}
 						}
-					}
-				} else if (child.type === "link") {
-					const text = sourceCode.getText(child);
-					const { url } = child;
 
-					if (
-						text === url ||
-						url === `http://${text}` ||
-						url === `mailto:${text}`
-					) {
-						bareUrls.push(child);
+						if (child.type === "link") {
+							const text = sourceCode.getText(child);
+							const { url } = child;
+
+							if (
+								text === url ||
+								url === `http://${text}` ||
+								url === `mailto:${text}`
+							) {
+								bareUrls.push(child);
+							}
+						}
+
+						traverse(child);
 					}
 				}
 			}
+
+			traverse(node);
 		}
 
 		return {
@@ -131,6 +157,10 @@ export default {
 			},
 
 			paragraph(node) {
+				findBareUrls(node);
+			},
+
+			heading(node) {
 				findBareUrls(node);
 			},
 
--- a/src/rules/no-bare-urls.js
+++ b/src/rules/no-bare-urls.js
@@ -1,0 +1,142 @@
+/**
+ * @fileoverview Rule to prevent bare URLs in Markdown.
+ * @author xbinaryx
+ */
+
+//-----------------------------------------------------------------------------
+// Type Definitions
+//-----------------------------------------------------------------------------
+
+/** @typedef {import("mdast").Paragraph} ParagraphNode */
+/** @typedef {import("mdast").TableCell} TableCellNode */
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoBareUrlsRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const htmlTagPattern = /^<([^!>][^/\s>]*)/u;
+
+/**
+ * Parses an HTML tag to extract its name and closing status
+ * @param {string} tagText The HTML tag text to parse
+ * @returns {{ name: string; isClosing: boolean; } | null} Object containing tag name and closing status, or null if not a valid tag
+ */
+function parseHtmlTag(tagText) {
+	const match = tagText.match(htmlTagPattern);
+	if (match) {
+		const tagName = match[1].toLowerCase();
+		const isClosing = tagName.startsWith("/");
+
+		return {
+			name: isClosing ? tagName.slice(1) : tagName,
+			isClosing,
+		};
+	}
+
+	return null;
+}
+
+//-----------------------------------------------------------------------------
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+/** @type {NoBareUrlsRuleDefinition} */
+export default {
+	meta: {
+		type: "problem",
+
+		docs: {
+			description: "Disallow bare URLs",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-bare-urls.md",
+		},
+
+		fixable: "code",
+
+		messages: {
+			bareUrl:
+				"Unexpected bare URL. Use autolink (<URL>) or link ([text](URL)) instead.",
+		},
+	},
+
+	create(context) {
+		const { sourceCode } = context;
+		const bareUrls = [];
+
+		/**
+		 * Finds bare URLs in markdown nodes while handling HTML tags.
+		 * When an HTML tag is found, it looks for its closing tag and skips all nodes
+		 * between them to prevent checking for bare URLs inside HTML content.
+		 * @param {ParagraphNode|TableCellNode} node The node to process
+		 * @returns {void}
+		 */
+		function findBareUrls(node) {
+			if (!node.children) {
+				return;
+			}
+
+			for (let i = 0; i < node.children.length; i++) {
+				const child = node.children[i];
+
+				if (child.type === "html") {
+					const tagInfo = parseHtmlTag(sourceCode.getText(child));
+
+					if (tagInfo && !tagInfo.isClosing) {
+						for (let j = i + 1; j < node.children.length; j++) {
+							const nextChild = node.children[j];
+							if (nextChild.type === "html") {
+								const closingTagInfo = parseHtmlTag(
+									sourceCode.getText(nextChild),
+								);
+								if (
+									closingTagInfo?.name === tagInfo.name &&
+									closingTagInfo?.isClosing
+								) {
+									i = j;
+									break;
+								}
+							}
+						}
+					}
+				} else if (child.type === "link") {
+					const text = sourceCode.getText(child);
+					const { url } = child;
+
+					if (
+						text === url ||
+						url === `http://${text}` ||
+						url === `mailto:${text}`
+					) {
+						bareUrls.push(child);
+					}
+				}
+			}
+		}
+
+		return {
+			"root:exit"() {
+				for (const bareUrl of bareUrls) {
+					context.report({
+						node: bareUrl,
+						messageId: "bareUrl",
+						fix(fixer) {
+							const text = sourceCode.getText(bareUrl);
+							return fixer.replaceText(bareUrl, `<${text}>`);
+						},
+					});
+				}
+			},
+
+			paragraph(node) {
+				findBareUrls(node);
+			},
+
+			tableCell(node) {
+				findBareUrls(node);
+			},
+		};
+	},
+};
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ export default defineConfig([
 | :- | :- | :-: |
 | [`fenced-code-language`](./docs/rules/fenced-code-language.md) | Require languages for fenced code blocks | yes |
 | [`heading-increment`](./docs/rules/heading-increment.md) | Enforce heading levels increment by one | yes |
+| [`no-bare-urls`](./docs/rules/no-bare-urls.md) | Disallow bare URLs | no |
 | [`no-duplicate-definitions`](./docs/rules/no-duplicate-definitions.md) | Disallow duplicate definitions | yes |
 | [`no-duplicate-headings`](./docs/rules/no-duplicate-headings.md) | Disallow duplicate headings in the same document | no |
 | [`no-empty-definitions`](./docs/rules/no-empty-definitions.md) | Disallow empty definitions | yes |
--- a/tests/rules/no-bare-urls.test.js
+++ b/tests/rules/no-bare-urls.test.js
@@ -249,6 +249,19 @@ ruleTester.run("no-bare-urls", rule, {
 			],
 		},
 		{
+			code: "Text <a>https://example.com</a> https://example.com",
+			output: "Text <a>https://example.com</a> <https://example.com>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 33,
+					endLine: 1,
+					endColumn: 52,
+				},
+			],
+		},
+		{
 			code: "<br> Another violation: https://example.com. <br>",
 			output: "<br> Another violation: <https://example.com>. <br>",
 			errors: [
--- a/tests/rules/no-bare-urls.test.js
+++ b/tests/rules/no-bare-urls.test.js
@@ -1,0 +1,309 @@
+/**
+ * @fileoverview Tests for no-bare-urls rule.
+ * @author xbinaryx
+ */
+
+//------------------------------------------------------------------------------
+// Imports
+//------------------------------------------------------------------------------
+
+import rule from "../../src/rules/no-bare-urls.js";
+import markdown from "../../src/index.js";
+import { RuleTester } from "eslint";
+import dedent from "dedent";
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+	plugins: {
+		markdown,
+	},
+	language: "markdown/gfm",
+});
+
+ruleTester.run("no-bare-urls", rule, {
+	valid: [
+		"<https://www.google.com/>",
+		"<user@example.com>",
+		"[text [undefined] text](https://example.com).",
+		"[is-a-valid]: https://example.com",
+		'<a href="https://example.com">link</a>.',
+		'<a href="https://example.com">https://example.com</a>',
+		'<a href="https://example.com">Text https://example.com</a>',
+		"This is not a bare [link]( https://example.com )",
+		"Nor is [link](https://example.com/path-with(parens)).",
+		"Or <https://example.com/path-with(parens)>.",
+		dedent`
+        <element-name first-attribute=" https://example.com/first " second-attribute=" https://example.com/second ">
+            Text
+        </element-name>`,
+		dedent`
+        <element-name
+            first-attribute=" https://example.com/first "
+            second-attribute=" https://example.com/second "></element-name>`,
+		"Not <code>https://example.com</code> bare.",
+		"Not <pre>https://example.com</pre> bare.",
+		dedent`
+        <p>
+        Not bare due to being in an HTML block:
+        https://example.com
+        <code>https://example.com</code>
+        <pre>https://example.com</pre>
+        </p>`,
+		dedent`
+        <div>
+        https://example.com
+        </div>`,
+		dedent`
+        <div>
+        https://example.com
+
+        </div>`,
+		"Text [link to https://example.com site](https://example.com) text.",
+		"Image ![for https://example.com site](https://example.com) text.",
+		"Links with spaces inside angle brackets are okay: [blue jay](<https://en.wikipedia.org/wiki/Blue jay>)",
+		"[link \\[is-not-a-valid\\] link](https://example.com)",
+		"text <https://example.com/dir/dir>",
+		dedent`
+        \`\`\`text
+        Code https://example.com/code?type=fence code
+        \`\`\``,
+		"       Code https://example.com/code?type=indent code",
+	],
+	invalid: [
+		{
+			code: "https://www.example.com/",
+			output: "<https://www.example.com/>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 25,
+				},
+			],
+		},
+		{
+			code: "user@example.com",
+			output: "<user@example.com>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 17,
+				},
+			],
+		},
+		{
+			code: "text https://www.google.com/",
+			output: "text <https://www.google.com/>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 6,
+					endLine: 1,
+					endColumn: 29,
+				},
+			],
+		},
+		{
+			code: "hTtPs://gOoGlE.cOm/",
+			output: "<hTtPs://gOoGlE.cOm/>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 20,
+				},
+			],
+		},
+		{
+			code: "(https://example.com)",
+			output: "(<https://example.com>)",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 2,
+					endLine: 1,
+					endColumn: 21,
+				},
+			],
+		},
+		{
+			code: dedent`
+            | Link                 |
+            |----------------------|
+            | https://example.com/ |`,
+			output: dedent`
+            | Link                 |
+            |----------------------|
+            | <https://example.com/> |`,
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 3,
+					column: 3,
+					endLine: 3,
+					endColumn: 23,
+				},
+			],
+		},
+		{
+			code: dedent`
+            [link that [is-a-valid] link](https://example.com)
+            
+            [is-a-valid]: https://example.com
+            `,
+			output: dedent`
+            [link that [is-a-valid] link](<https://example.com>)
+            
+            [is-a-valid]: https://example.com
+            `,
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 31,
+					endLine: 1,
+					endColumn: 50,
+				},
+			],
+		},
+		{
+			code: "Text <https://example.com/brackets> text https://example.com/bare text",
+			output: "Text <https://example.com/brackets> text <https://example.com/bare> text",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 42,
+					endLine: 1,
+					endColumn: 66,
+				},
+			],
+		},
+		{
+			code: "> https://example.com/",
+			output: "> <https://example.com/>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 3,
+					endLine: 1,
+					endColumn: 23,
+				},
+			],
+		},
+		{
+			code: "Text https://example.com/first more text https://example.com/second",
+			output: "Text <https://example.com/first> more text <https://example.com/second>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 6,
+					endLine: 1,
+					endColumn: 31,
+				},
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 42,
+					endLine: 1,
+					endColumn: 68,
+				},
+			],
+		},
+		{
+			code: dedent`
+            Text https://example.com/first
+            more text https://example.com/second`,
+			output: dedent`
+            Text <https://example.com/first>
+            more text <https://example.com/second>`,
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 6,
+					endLine: 1,
+					endColumn: 31,
+				},
+				{
+					messageId: "bareUrl",
+					line: 2,
+					column: 11,
+					endLine: 2,
+					endColumn: 37,
+				},
+			],
+		},
+		{
+			code: "<br> Another violation: https://example.com. <br>",
+			output: "<br> Another violation: <https://example.com>. <br>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 25,
+					endLine: 1,
+					endColumn: 44,
+				},
+			],
+		},
+		{
+			code: dedent`
+            <div>
+            
+            https://example.com
+            </div>`,
+			output: dedent`
+            <div>
+            
+            <https://example.com>
+            </div>`,
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 3,
+					column: 1,
+					endLine: 3,
+					endColumn: 20,
+				},
+			],
+		},
+		{
+			code: dedent`
+            <div>
+            
+            https://example.com
+
+            </div>`,
+			output: dedent`
+            <div>
+            
+            <https://example.com>
+
+            </div>`,
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 3,
+					column: 1,
+					endLine: 3,
+					endColumn: 20,
+				},
+			],
+		},
+	],
+});
--- a/tests/rules/no-bare-urls.test.js
+++ b/tests/rules/no-bare-urls.test.js
@@ -139,6 +139,19 @@ ruleTester.run("no-bare-urls", rule, {
 			],
 		},
 		{
+			code: "# https://www.example.com/",
+			output: "# <https://www.example.com/>",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 3,
+					endLine: 1,
+					endColumn: 27,
+				},
+			],
+		},
+		{
 			code: dedent`
             | Link                 |
             |----------------------|
@@ -315,6 +328,19 @@ ruleTester.run("no-bare-urls", rule, {
 					column: 1,
 					endLine: 3,
 					endColumn: 20,
+				},
+			],
+		},
+		{
+			code: "text **<a>https://example.com</a>** *https://example.com*",
+			output: "text **<a>https://example.com</a>** *<https://example.com>*",
+			errors: [
+				{
+					messageId: "bareUrl",
+					line: 1,
+					column: 38,
+					endLine: 1,
+					endColumn: 57,
 				},
 			],
 		},


docs/rules/no-bare-urls.md
```diff
@@ -0,0 +1,53 @@
...
+Disallow bare URLs.
+
+## Background
+
+In Markdown, URLs are typically formatted in two ways:
+
+1. Autolinks: URLs wrapped in angle brackets, such as `<https://example.com/>`
+2. Links: URLs with descriptive text, such as `[Visit our website](https://example.com/)`
+
+A bare URL appears without angle brackets (`<>`), such as `https://example.com/`. While GitHub Flavored Markdown (GFM) allows bare URLs to be recognized as clickable links, this feature is not supported in all Markdown processors. To ensure compatibility across different processors, use autolinks or links instead.
```
nzakas, line 12: Can you link "autolinks" and "links" in the last sentence to somewhere? I'm not sure people will understand what these mean without additional context.

docs/rules/no-bare-urls.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Rule Details
+
+> [!IMPORTANT] <!-- eslint-disable-line -- This should be fixed in https://github.com/eslint/markdown/issues/294 -->
+>
+> This rule relies on bare URLs being link nodes, which requires a GFM-compatible parser (e.g., `language: "markdown/gfm"`).
```
nzakas: We typically don't mention nodes in documentation because users don't need to understand ASTs. If you're saying this rule requires GFM, we should say that:

```suggestion
> This rule requires `language: "markdown/gfm"`.
```

And a followup question: It seems like this should apply regardless of the Markdown format. I know it's easier to detect in GFM, but it seems relatively easy to also pick up in CommonMark. I think the user expectation is that it would work in both. What do you think?

src/rules/no-bare-urls.js
```diff
@@ -0,0 +1,142 @@
...
+		const bareUrls = [];
+
+		/**
+		 * Finds bare URLs in markdown nodes while handling HTML tags.
+		 * When an HTML tag is found, it looks for its closing tag and skips all nodes
+		 * between them to prevent checking for bare URLs inside HTML content.
+		 * @param {ParagraphNode|TableCellNode} node The node to process
+		 * @returns {void}
+		 */
+		function findBareUrls(node) {
```
nzakas, line 80: I'm not sure I understand the approach you're taking here. It seems like you're doing extra traversals that may not be necessary.

To start, you could look for `link` in the visitor object and just check if the first and last characters are `<` and `>`. That would completely cover GFM.

You could use `html` and `html:exit` to ensure that you're not flagging anything that appears inside of an HTML block.

You could also use `text` and run a regex to find the rest of the URLs for CommonMark.

Is there something I'm missing about the approach you chose?

docs/rules/no-bare-urls.md
xbinaryx: Want me to add links to the CommonMark spec for "autolinks" and "links"? I tried to explain them in the numbered points above, but the spec links might be helpful too.

docs/rules/no-bare-urls.md
xbinaryx: > We typically don't mention nodes in documentation because users don't need to understand ASTs. If you're saying this rule requires GFM, we should say that:

SGTM. I tried to word it similarly to how it's done in [table-column-count.md](https://github.com/eslint/markdown/blob/8b77e6993ed045482a966470d906d9636b1b4abc/docs/rules/table-column-count.md?plain=1#L13).

> And a followup question: It seems like this should apply regardless of the Markdown format. I know it's easier to detect in GFM, but it seems relatively easy to also pick up in CommonMark. I think the user expectation is that it would work in both. What do you think?

I tried that approach first, but it got complicated. We'd need to write regex that matches URLs and email addresses according to the spec, and then detect if these aren't inside HTML, code spans, etc. I decided to follow how remarklint and markdownlint implemented this rule instead.

src/rules/no-bare-urls.js
xbinaryx: This is because links inside HTML produce an AST that looks like [this](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG57XFxuICAgIFxcXCJrZXkxXFxcIjogW3RydWUsIGZhbHNlLCBudWxsXSxcXG4gICAgXFxcImtleTJcXFwiOiB7XFxuICAgICAgICBcXFwia2V5M1xcXCI6IFsxLCAyLCBcXFwiM1xcXCIsIDFlMTAsIDFlLTNdXFxuICAgIH1cXG59XCIsXCJtYXJrZG93blwiOlwiPGE%2BaHR0cHM6Ly9leGFtcGxlLmNvbTwvYT5cIixcImNzc1wiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIENTUyBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG4gKiB0aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcbiAqXFxuICogVGhlIHRhYnMgYXJlOlxcbiAqXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICpcXG4gKiBZb3UgY2FuIGNoYW5nZSB0aGUgd2F5IHRoYXQgdGhlIENTUyBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkNTU1xcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuQGltcG9ydCB1cmwoJ2h0dHBzOi8vZm9udHMuZ29vZ2xlYXBpcy5jb20vY3NzMj9mYW1pbHk9Um9ib3RvOndnaHRANDAwOzcwMCZkaXNwbGF5PXN3YXAnKTtcXG5cXG5ib2R5IHtcXG5cXHRmb250LWZhbWlseTogc2Fucy1zZXJpZjtcXG59XFxuXFxuaDEge1xcblxcdGNvbG9yOiAjMzMzO1xcbn1cXG5cXG5wIHtcXG5cXHRtYXJnaW46IDFlbSAwO1xcbn1cIixcImh0bWxcIjpcIjwhRE9DVFlQRSBodG1sPlxcbjwhLS1cXG5UeXBlIG9yIHBhc3RlIHNvbWUgSFRNTCBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG50aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcblxcblRoZSB0YWJzIGFyZTpcXG5cXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0Llxcbi0tPlxcblxcbjxodG1sIGxhbmc9XFxcImVuXFxcIj5cXG4gICAgPGhlYWQ%2BXFxuICAgICAgICA8bWV0YSBjaGFyc2V0PVxcXCJVVEYtOFxcXCI%2BXFxuICAgICAgICA8dGl0bGU%2BSFRNTDwvdGl0bGU%2BXFxuICAgIDwvaGVhZD5cXG4gICAgPGJvZHk%2BXFxuICAgICAgICA8cD5UZXh0PC9wPlxcbiAgICA8L2JvZHk%2BXFxuPC9odG1sPlwifSxcImxhbmd1YWdlXCI6XCJtYXJrZG93blwiLFwianNPcHRpb25zXCI6e1wicGFyc2VyXCI6XCJlc3ByZWVcIixcInNvdXJjZVR5cGVcIjpcIm1vZHVsZVwiLFwiZXNWZXJzaW9uXCI6XCJsYXRlc3RcIixcImlzSlNYXCI6dHJ1ZX0sXCJqc29uT3B0aW9uc1wiOntcImpzb25Nb2RlXCI6XCJqc29uY1wiLFwiYWxsb3dUcmFpbGluZ0NvbW1hc1wiOmZhbHNlfSxcImNzc09wdGlvbnNcIjp7XCJjc3NNb2RlXCI6XCJjc3NcIixcInRvbGVyYW50XCI6ZmFsc2V9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJnZm1cIixcIm1hcmtkb3duRnJvbnRtYXR0ZXJcIjpcIm9mZlwifSxcIndyYXBcIjp0cnVlLFwidmlld01vZGVzXCI6e1wiYXN0Vmlld1wiOlwidHJlZVwiLFwic2NvcGVWaWV3XCI6XCJmbGF0XCIsXCJwYXRoVmlld1wiOlwiZ3JhcGhcIn0sXCJwYXRoSW5kZXhcIjp7XCJpbmRleFwiOjAsXCJpbmRleGVzXCI6MX0sXCJlc3F1ZXJ5U2VsZWN0b3JcIjp7XCJzZWxlY3RvclwiOlwiXCJ9fSxcInZlcnNpb25cIjowfSI%3D). So we cant use `html` and `html:exit`

docs/rules/no-bare-urls.md
nzakas: Ah good call, I think I skimmed over that part. No need to add the URLs then.

src/rules/no-bare-urls.js
nzakas: Interesting problem! I think you can still use the visitor methods and then use `html + link` (sibling operator) to ignore certain links?

docs/rules/no-bare-urls.md
```diff
@@ -0,0 +1,53 @@
...
+<!-- eslint markdown/no-bare-urls: "error" -->
+
+For more info, visit <https://www.example.com/>
+
+For more info, visit [Example Website](https://www.example.com/)
+
+Contact us at <user@example.com>
+
+Contact us at [user@example.com](mailto:user@example.com)
+```
```
lumirlumir: Could you add more correct and incorrect examples listed in the Markdownlint rule?

image

It might help users transition to this rule more smoothly.

src/rules/no-bare-urls.js
xbinaryx: With the next-sibling operator (`+`), this is a false positive:
```markdown
<a>Text https://example.com</a>
```

With the subsequent-sibling operator (`~`), this is false negative:
```markdown
<span>Text</span> https://example.com
```

nzakas: Hmm okay. I still feel like there has to be a more efficient way of doing this but I can't think of anything else off the top of my head.

nzakas (approved):

nzakas:
ping @lumirlumir

docs/rules/no-bare-urls.md
lumirlumir: I would like to give 👍 to what @nzakas said.

Since it might be more complicated to implement—especially if we consider CommonMark autolinks—we can use the same approach for both CommonMark and GFM autolinks.

For example, if we implement this rule as it is now, and other users later request that it also apply to CommonMark, the rule might lack extensibility, since it’s currently based on an approach that only works in GFM mode.

So, since GFM is a superset of CommonMark, I think we could set the standard based on CommonMark. Then, we can extend the basic logic to support GFM mode as well.

src/rules/no-bare-urls.js
lumirlumir: I've looked into the [CommonMark Spec for HTML Blocks](https://spec.commonmark.org/0.31.2/#html-block), and it's definitely an interesting topic to think about.

According to the spec, the `<a>` tag cannot create an HTML block if it is on a single line, so the AST ends up being a bit unexpected—different from what we might intuitively expect.

Please take a look at the example code below:

[AST for this code](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG57XFxuICAgIFxcXCJrZXkxXFxcIjogW3RydWUsIGZhbHNlLCBudWxsXSxcXG4gICAgXFxcImtleTJcXFwiOiB7XFxuICAgICAgICBcXFwia2V5M1xcXCI6IFsxLCAyLCBcXFwiM1xcXCIsIDFlMTAsIDFlLTNdXFxuICAgIH1cXG59XCIsXCJtYXJrZG93blwiOlwiPGE%2BVGV4dCBodHRwczovL2V4YW1wbGUuY29tPC9hPlwiLFwiY3NzXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgQ1NTIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgQ1NTIGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiQ1NTXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKi9cXG5cXG5AaW1wb3J0IHVybCgnaHR0cHM6Ly9mb250cy5nb29nbGVhcGlzLmNvbS9jc3MyP2ZhbWlseT1Sb2JvdG86d2dodEA0MDA7NzAwJmRpc3BsYXk9c3dhcCcpO1xcblxcbmJvZHkge1xcblxcdGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xcbn1cXG5cXG5oMSB7XFxuXFx0Y29sb3I6ICMzMzM7XFxufVxcblxcbnAge1xcblxcdG1hcmdpbjogMWVtIDA7XFxufVwiLFwiaHRtbFwiOlwiPCFET0NUWVBFIGh0bWw%2BXFxuPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbnRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuXFxuVGhlIHRhYnMgYXJlOlxcblxcbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG5jYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuLS0%2BXFxuXFxuPGh0bWwgbGFuZz1cXFwiZW5cXFwiPlxcbiAgICA8aGVhZD5cXG4gICAgICAgIDxtZXRhIGNoYXJzZXQ9XFxcIlVURi04XFxcIj5cXG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cXG4gICAgPC9oZWFkPlxcbiAgICA8Ym9keT5cXG4gICAgICAgIDxwPlRleHQ8L3A%2BXFxuICAgIDwvYm9keT5cXG48L2h0bWw%2BXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb25jXCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D)

```md
<a>Text https://example.com</a>
```

We can interpret this code in two ways.

## 1. Following the CommonMark spec

CommonMark recognizes the whole code as a `Paragraph` node. So the `<a>` tag is interpreted as inline HTML (not a HTML block), and `Text https://example.com` as `Text` and a `Link` node.

So, we can treat `Text https://example.com` as a paragraph (or text).

If we determine this to be a paragraph (or text), it should trigger an error, because the CommonMark spec says that `Text https://example.com` is not an HTML block.

But the downside of this approach is that it’s different from what we might intuitively expect.

## 2. Following our intuition

I think this approach is what @xbinaryx might have expected.

Although the spec says it is a paragraph (or text), we intuitively know that the link inside an `<a>` tag should not trigger an error.

Also, Markdownlint follows this approach.

image

---

In conclusion, the second approach is more complicated and harder to implement, but I think it is worth following, since it aligns with our intuition and is the approach that Markdownlint uses.

I might be wrong, but it would be nice to consider this comment before deciding on the final approach for this rule.

src/rules/no-bare-urls.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoBareUrlsRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const htmlTagPattern = /^<([^!>][^/\s>]*)/u;
```
lumirlumir:
```suggestion
const htmlTagPattern = /^<([^!>][^/\s>]*)/u;
```

Could you make the `htmlTagPattern` more detailed?

Considering the CommonMark spec, we can use ` ` (space) and `\t` as whitespace characters instead of `\s`.

References:

* [https://spec.commonmark.org/0.31.2/#html-block](https://spec.commonmark.org/0.31.2/#html-block)
* [https://github.com/eslint/markdown/pull/421](https://github.com/eslint/markdown/pull/421)

Currently, `\s` also matches NBSP (`\u00a0`) and similar characters, which are not considered standard whitespace in CommonMark.

```js
console.log(/\s/.test('\u00a0')); // true
```

lumirlumir (changes requested):

docs/rules/no-bare-urls.md
lumirlumir: Maybe we can take a look at the GFM spec for the regex.

https://github.github.com/gfm/#autolinks-extension-

nzakas: I think it's okay to release the first version of this rule that works only with GFM and then follow up with CommonMark later. Rules don't have to be perfect initially and sometimes it's helpful to gather feedback from users when deciding future direction.

src/rules/no-bare-urls.js
nzakas: As with the previous comment, I think it's fine to start with one approach for the initial release of the rule. We can always change it in the future if necessary.

docs/rules/no-bare-urls.md
lumirlumir: Sounds good to me 👍

src/rules/no-bare-urls.js
xbinaryx: You're right about CommonMark's whitespace rules for Markdown, but since we're parsing HTML tags specifically, we need to account for all valid HTML whitespace. With just space/tab matching, this valid HTML would incorrectly flag as a bare URL:

```html
<a
href="#">https://eslint.org/</a>
```

lumirlumir: @xbinaryx Thanks for the explanation.

What do you think about using `(?:\r\n|[ \t\n\r])` or something similar instead of `\s`? I think it’s a bit more accurate, and I’d love to hear your thoughts.

FYI, with NBSP, it results in a different [AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG4vLyBoaVxcblxcbi8qIGhpICovXFxuXFxue1xcbiAgICBcXFwia2V5MVxcXCI6IFt0cnVlLCBmYWxzZSwgbnVsbF0sXFxuICAgIFxcXCJrZXkyXFxcIjoge1xcbiAgICAgICAgXFxcImtleTNcXFwiOiBbMSwgMiwgXFxcIjNcXFwiLCAxZTEwLCAxZS0zXVxcbiAgICB9XFxufVwiLFwibWFya2Rvd25cIjpcIjxhoFxcbmhyZWY9XFxcIiNcXFwiPmh0dHBzOi8vZXNsaW50Lm9yZy88L2E%2BO1wiLFwiY3NzXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgQ1NTIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgQ1NTIGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiQ1NTXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKi9cXG5cXG5AaW1wb3J0IHVybCgnaHR0cHM6Ly9mb250cy5nb29nbGVhcGlzLmNvbS9jc3MyP2ZhbWlseT1Sb2JvdG86d2dodEA0MDA7NzAwJmRpc3BsYXk9c3dhcCcpO1xcblxcbmJvZHkge1xcblxcdGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xcbn1cXG5cXG5oMSB7XFxuXFx0Y29sb3I6ICMzMzM7XFxufVxcblxcbnAge1xcblxcdG1hcmdpbjogMWVtIDA7XFxufVwiLFwiaHRtbFwiOlwiPCFET0NUWVBFIGh0bWw%2BXFxuPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbnRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuXFxuVGhlIHRhYnMgYXJlOlxcblxcbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG5jYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuLS0%2BXFxuXFxuPGh0bWwgbGFuZz1cXFwiZW5cXFwiPlxcbiAgICA8aGVhZD5cXG4gICAgICAgIDxtZXRhIGNoYXJzZXQ9XFxcIlVURi04XFxcIj5cXG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cXG4gICAgPC9oZWFkPlxcbiAgICA8Ym9keT5cXG4gICAgICAgIDxwPlRleHQ8L3A%2BXFxuICAgIDwvYm9keT5cXG48L2h0bWw%2BXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb25jXCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D).

image

src/rules/no-bare-urls.js
```diff
@@ -0,0 +1,142 @@
...
+				}
+			},
+
+			paragraph(node) {
+				findBareUrls(node);
+			},
+
+			tableCell(node) {
+				findBareUrls(node);
+			},
```
lumirlumir:
```suggestion
			paragraph(node) {
				findBareUrls(node);
			},
			
			heading(node) {
			    findBareUrls(node);
			},

			tableCell(node) {
				findBareUrls(node);
			},
```

Could you add `heading` to the code and documentation example?

Since a `Heading` node doesn’t yield a `Paragraph` node, I think we need to check it as well.

image

image

src/rules/no-bare-urls.js
```diff
@@ -0,0 +1,142 @@
...
+		 * between them to prevent checking for bare URLs inside HTML content.
+		 * @param {ParagraphNode|TableCellNode} node The node to process
+		 * @returns {void}
+		 */
+		function findBareUrls(node) {
+			if (!node.children) {
+				return;
+			}
+
+			for (let i = 0; i < node.children.length; i++) {
```
lumirlumir:
```suggestion
			for (let i = 0; i < node.children.length; i++) {
```

Interestingly, `Html` and `Link` nodes can appear deeper than just 1-level depth.

[Here's the AST for reference.](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG4vLyBoaVxcblxcbi8qIGhpICovXFxuXFxue1xcbiAgICBcXFwia2V5MVxcXCI6IFt0cnVlLCBmYWxzZSwgbnVsbF0sXFxuICAgIFxcXCJrZXkyXFxcIjoge1xcbiAgICAgICAgXFxcImtleTNcXFwiOiBbMSwgMiwgXFxcIjNcXFwiLCAxZTEwLCAxZS0zXVxcbiAgICB9XFxufVwiLFwibWFya2Rvd25cIjpcImhpICoqPGE%2BaGk8L2E%2BKiogKmh0dHBzOi8vd3d3Lmdvb2dsZS5jb20qXCIsXCJjc3NcIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBDU1MgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKlxcbiAqIFRoZSB0YWJzIGFyZTpcXG4gKlxcbiAqIC0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG4gKiAgIGJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIHRoZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUuIFlvdVxcbiAqICAgY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxcbiAqXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBDU1MgY29kZSBpcyBpbnRlcnByZXRlZFxcbiAqIGJ5IGNsaWNraW5nIFxcXCJDU1NcXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqL1xcblxcbkBpbXBvcnQgdXJsKCdodHRwczovL2ZvbnRzLmdvb2dsZWFwaXMuY29tL2NzczI%2FZmFtaWx5PVJvYm90bzp3Z2h0QDQwMDs3MDAmZGlzcGxheT1zd2FwJyk7XFxuXFxuYm9keSB7XFxuXFx0Zm9udC1mYW1pbHk6IHNhbnMtc2VyaWY7XFxufVxcblxcbmgxIHtcXG5cXHRjb2xvcjogIzMzMztcXG59XFxuXFxucCB7XFxuXFx0bWFyZ2luOiAxZW0gMDtcXG59XCIsXCJodG1sXCI6XCI8IURPQ1RZUEUgaHRtbD5cXG48IS0tXFxuVHlwZSBvciBwYXN0ZSBzb21lIEhUTUwgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG5cXG5UaGUgdGFicyBhcmU6XFxuXFxuLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbmJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIHRoZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUuIFlvdVxcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4tLT5cXG5cXG48aHRtbCBsYW5nPVxcXCJlblxcXCI%2BXFxuICAgIDxoZWFkPlxcbiAgICAgICAgPG1ldGEgY2hhcnNldD1cXFwiVVRGLThcXFwiPlxcbiAgICAgICAgPHRpdGxlPkhUTUw8L3RpdGxlPlxcbiAgICA8L2hlYWQ%2BXFxuICAgIDxib2R5PlxcbiAgICAgICAgPHA%2BVGV4dDwvcD5cXG4gICAgPC9ib2R5PlxcbjwvaHRtbD5cIn0sXCJsYW5ndWFnZVwiOlwibWFya2Rvd25cIixcImpzT3B0aW9uc1wiOntcInBhcnNlclwiOlwiZXNwcmVlXCIsXCJzb3VyY2VUeXBlXCI6XCJtb2R1bGVcIixcImVzVmVyc2lvblwiOlwibGF0ZXN0XCIsXCJpc0pTWFwiOnRydWV9LFwianNvbk9wdGlvbnNcIjp7XCJqc29uTW9kZVwiOlwianNvbmNcIixcImFsbG93VHJhaWxpbmdDb21tYXNcIjpmYWxzZX0sXCJjc3NPcHRpb25zXCI6e1wiY3NzTW9kZVwiOlwiY3NzXCIsXCJ0b2xlcmFudFwiOmZhbHNlfSxcIm1hcmtkb3duT3B0aW9uc1wiOntcIm1hcmtkb3duTW9kZVwiOlwiZ2ZtXCIsXCJtYXJrZG93bkZyb250bWF0dGVyXCI6XCJvZmZcIn0sXCJ3cmFwXCI6dHJ1ZSxcInZpZXdNb2Rlc1wiOntcImFzdFZpZXdcIjpcInRyZWVcIixcInNjb3BlVmlld1wiOlwiZmxhdFwiLFwicGF0aFZpZXdcIjpcImdyYXBoXCJ9LFwicGF0aEluZGV4XCI6e1wiaW5kZXhcIjowLFwiaW5kZXhlc1wiOjF9LFwiZXNxdWVyeVNlbGVjdG9yXCI6e1wic2VsZWN0b3JcIjpcIlwifX0sXCJ2ZXJzaW9uXCI6MH0i)

image

So, I think it would be better if we traverse every child node recursively, like in the code you used [here](https://github.com/eslint/markdown/pull/398/commits/3bc7fc72c83856a479e4d1b46d94131e455334bb).

src/rules/no-bare-urls.js
```diff
@@ -0,0 +1,142 @@
...
+
+		messages: {
+			bareUrl:
+				"Unexpected bare URL. Use autolink (<URL>) or link ([text](URL)) instead.",
+		},
+	},
+
+	create(context) {
+		const { sourceCode } = context;
+		const bareUrls = [];
```
lumirlumir:
```suggestion
		const { sourceCode } = context;
		/** @type {...} */
		const bareUrls = [];
```

Could you add a type definition for `bareUrls`? It would be clearer if the variable had an explicit type.

lumirlumir (changes requested):

src/rules/no-bare-urls.js
xbinaryx: > What do you think about using (?:\r\n|[ \t\n\r]) or something similar instead of \s? I think it’s a bit more accurate, and I’d love to hear your thoughts.

Could you share an example where `\s` would incorrectly match (or miss) something compared to `(?:\r\n|[ \t\n\r])`?

For context, this same pattern is used in [Markdownlint’s HTML tag parsing](https://github.com/DavidAnson/markdownlint/blob/224987d727ebada36752074d2466a061c9e03ff0/helpers/micromark-helpers.cjs#L242).

> FYI, with NBSP, it results in a different [AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG4vLyBoaVxcblxcbi8qIGhpICovXFxuXFxue1xcbiAgICBcXFwia2V5MVxcXCI6IFt0cnVlLCBmYWxzZSwgbnVsbF0sXFxuICAgIFxcXCJrZXkyXFxcIjoge1xcbiAgICAgICAgXFxcImtleTNcXFwiOiBbMSwgMiwgXFxcIjNcXFwiLCAxZTEwLCAxZS0zXVxcbiAgICB9XFxufVwiLFwibWFya2Rvd25cIjpcIjxhoFxcbmhyZWY9XFxcIiNcXFwiPmh0dHBzOi8vZXNsaW50Lm9yZy88L2E%2BO1wiLFwiY3NzXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgQ1NTIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgQ1NTIGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiQ1NTXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKi9cXG5cXG5AaW1wb3J0IHVybCgnaHR0cHM6Ly9mb250cy5nb29nbGVhcGlzLmNvbS9jc3MyP2ZhbWlseT1Sb2JvdG86d2dodEA0MDA7NzAwJmRpc3BsYXk9c3dhcCcpO1xcblxcbmJvZHkge1xcblxcdGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xcbn1cXG5cXG5oMSB7XFxuXFx0Y29sb3I6ICMzMzM7XFxufVxcblxcbnAge1xcblxcdG1hcmdpbjogMWVtIDA7XFxufVwiLFwiaHRtbFwiOlwiPCFET0NUWVBFIGh0bWw%2BXFxuPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbnRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuXFxuVGhlIHRhYnMgYXJlOlxcblxcbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG5jYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuLS0%2BXFxuXFxuPGh0bWwgbGFuZz1cXFwiZW5cXFwiPlxcbiAgICA8aGVhZD5cXG4gICAgICAgIDxtZXRhIGNoYXJzZXQ9XFxcIlVURi04XFxcIj5cXG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cXG4gICAgPC9oZWFkPlxcbiAgICA8Ym9keT5cXG4gICAgICAgIDxwPlRleHQ8L3A%2BXFxuICAgIDwvYm9keT5cXG48L2h0bWw%2BXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb25jXCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D).

The rule should flag the URL in this context since this as a malformed tag.

lumirlumir: Thanks for sharing the context.

Let's go with your suggestion 👍

nzakas (approved):
LGTM. Thanks!

PR #432, opened 2025-06-22
chore: fix `tsconfig.json` to include `.js` and `.ts` files in `src` dir

lumirlumir:
#### Prerequisites checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "files": ["src/index.js"],
+  "include": ["src/**/*.js", "src/**/*.ts"],
   "compilerOptions": {
     "declaration": true,
     "allowJs": true,


tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "files": ["src/index.js"],
+  "include": ["src/**/*.js", "src/**/*.ts"],
```
lumirlumir, line 2: If we use the `files` field, glob patterns are not supported.

snitin315:
So the package export's  aren't valid anymore:

https://github.com/eslint/markdown/blob/589323ff9f589e4fb7503a768ba91bf96b5d0ff4/package.json#L11-L24

lumirlumir:
@snitin315 Thanks for sharing the context.

However, the changes you mentioned originate from #383, not from this PR. (I believe the latest changes from #383 are not yet reflected in your local environment.)

Even if we revert the changes in this PR, the `dist` folder remains the same — the differences shown in your screenshot are due to #383.

So the package exports aren't valid anymore:

That’s exactly why #383 was marked as a breaking change — the file structure in the dist folder was changed in that PR.

snitin315 (approved):

PR #433, opened 2025-06-23
feat: add no-reference-like-urls rule

### LINKED ISSUE #419: New Rule: no-reference-like-url
### Rule details

Warn when URLs are also defined identifiers.

### Example code

```markdown
[Mercury](mercury) is the first planet from the sun.

[mercury]: https://example.com/mercury/
```

### Prior Art

```markdown
[remark-lint-no-reference-like-url](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-reference-like-url)
```

### Participation

### Additional comments

While full URLs for definition identifiers are okay ([https://example.com]: https://example.com), and what looks like an identifier could be an actual URL ([text](alpha)), the more common case is that, assuming a definition [alpha]: https://example.com, then a link [text](alpha) should instead have been [text][alpha].

TKDev7:
#### Prerequisites checklist

Add a new rule, no-reference-like-url, to disallow inline links or images when the URL matches a defined reference identifier.

Implemented the no-reference-like-url rule with autofix, added documentation, and included tests.

#### Related Issues

Fixes #419


docs/rules/no-reference-like-url.md
```diff
@@ -0,0 +1,57 @@
+# no-reference-like-url
+
+Disallow URLs that match defined reference identifiers.
+
+## Background
+
+In Markdown, you can create links using either inline syntax `[text](url)` or reference syntax `[text][id]` with a separate definition `[id]: url`. This rule encourages reference syntax when a link's URL matches an existing reference identifier.
```
snitin315:
```suggestion
In Markdown, you can create links using either inline syntax `[text](url)` or reference syntax `[text][id]` with a separate definition `[id]: url`. This rule encourages the use of reference syntax when a link's URL matches an existing reference identifier.
```

docs/rules/no-reference-like-url.md
```diff
@@ -0,0 +1,57 @@
...
+[mercury]: https://example.com/mercury/
+```
+
+```markdown
+<!-- eslint markdown/no-reference-like-url: "error" -->
+
+![**Venus** is a planet](venus).
+
+[venus]: https://example.com/venus.jpg
+```
```
snitin315: I think we can combine both examples in a single code block.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,99 @@
...
+			description:
+				"Disallow URLs that match defined reference identifiers",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-reference-like-url.md",
+		},
+
+		fixable: "code",
+
+		messages: {
+			referenceLikeUrl:
+				"Unexpected resource {{type}} with URL that matches a definition identifier. Use [text][id] syntax instead.",
```
snitin315: Let's also include the syntax used in error message `[type](url)` for link and `type` for images.
```suggestion
				"Unexpected resource {{type}} ('[type](url)') with URL that matches a definition identifier. Use '[text][id]' syntax instead.",
```

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,99 @@
...
+									1;
+								const endOffset =
+									node.position.start.offset +
+									closeParenIndex +
+									1;
+
+								return fixer.replaceTextRange(
+									[startOffset, endOffset],
+									`[${node.url}]`,
+								);
```
snitin315: The fixer also removes the link title. For example:

```markdown
[**Mercury**](mercury "Mercury Planet") is the first planet from the sun.

[mercury]: https://example.com/mercury/
```

the above code is get's auto fixed to:

```markdown
[**Mercury**][mercury] is the first planet from the sun.

[mercury]: https://example.com/mercury/
```

Instead of:

```markdown
```markdown
[**Mercury**](mercury) is the first planet from the sun.

[mercury]: https://example.com/mercury/ "Mercury Planet"
```

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,295 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+	plugins: {
+		markdown,
+	},
+	language: "markdown/commonmark",
+});
+
+ruleTester.run("no-reference-like-url", rule, {
```
snitin315: Let's add  valid/invalid test cases for:

- Links/images with titles.
- Identifiers with whitespace or special characters.
- Link with whitespaces:
```md
[Click Me
](test)

[test]: https://abc.com
```

src/rules/no-reference-like-url.js
TKDev7: Are you saying we shouldn't report links with titles?

snitin315: The rule reports correctly, but the autofix is incorrect.

TKDev7: Based on your example, it seems the correct autofix behavior would be to remove the inline title and move it to the reference definition. However, I'm not entirely comfortable with this approach.

I see two alternative solutions we could consider:

1. Don't flag links that contain titles
2. Remove the title during autofix and rely on the user having already defined it in the reference

I lean toward option 1 (skipping links with titles) as it's more predictable and preserves the author's original formatting choices. What are your thoughts on these alternatives?

snitin315: What does remarklint do in this case?

snitin315: Another approach could be to just report this error & not apply autofix in this case (when node.title is present).

TKDev7: > What does remarklint do in this case?

It flags links that contain titles. 

> Another approach could be to just report this error & not apply autofix in this case (when node.title is present).

I chose to implement this approach.

snitin315 (approved):
LGTM, thanks. Leaving it open for 2nd review.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,104 @@
...
+//-----------------------------------------------------------------------------
+
+/** @typedef {import("mdast").Node} Node */
+/** @typedef {import("mdast").Link} LinkNode */
+/** @typedef {import("mdast").Image} ImageNode */
+/** @typedef {import("mdast").Definition} DefinitionNode */
+/**
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
+ * NoReferenceLikeUrlRuleDefinition
+ */
```
lumirlumir:
```suggestion
/** @typedef {import("mdast").Node} Node */
/** @typedef {import("mdast").Link} LinkNode */
/** @typedef {import("mdast").Image} ImageNode */
/** @typedef {import("mdast").Definition} DefinitionNode */
/**
 * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>}
 * NoReferenceLikeUrlRuleDefinition
 */
```

image

lumirlumir (changes requested):

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,104 @@
...
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+/** @type {NoReferenceLikeUrlRuleDefinition} */
+export default {
+	meta: {
+		type: "problem",
+
+		docs: {
+			description:
```
lumirlumir:
```suggestion
		docs: {
		    recommended: true,
			description:
```

I think this rule could be added to the recommended set. 

I’d like to hear the @eslint/eslint-team’s thoughts.

JoshuaKGoldberg: 👍 agreed, I don't see any reason not to.

src/rules/no-reference-like-url.js
JoshuaKGoldberg: +1 on option 2:

* still reporting
* not auto-fixing when there's no way to do so without losing information.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,104 @@
...
+									`[${node.url}]`,
+								);
+							},
+						});
+					}
+				}
+			},
+
+			definition(node) {
+				definitionIdentifiers.add(node.identifier.toLowerCase());
```
lumirlumir:
```suggestion
				definitionIdentifiers.add(node.identifier);
```

Could you remove `toLowerCase()`? Mdast already normalizes the identifier value internally, so there’s no need to re‑normalize it.

image

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,104 @@
...
+		const { sourceCode } = context;
+		/** @type {Set<string>} */
+		const definitionIdentifiers = new Set();
+		/** @type {Array<Link|Image>} */
+		const resources = [];
+
+		return {
+			"root:exit"() {
+				for (const node of resources) {
+					const url = node.url.toLowerCase();
```
lumirlumir:
```suggestion
					const url = normalizeIdentifier(node.url).toLowerCase();
```

Mdast internally uses `normalizeIdentifier` function for Unicode-aware case folding. If we simply use `toLowerCase()`, it could result in unintended false negatives or positives.

I think we can use the [`mdast-util-normalize-identifier`](https://github.com/micromark/micromark/tree/main/packages/micromark-util-normalize-identifier) package, which is used by Mdast internally.

Ref: https://github.com/syntax-tree/mdast-util-from-markdown/blob/main/dev/lib/index.js#L656-L658

Could you apply it and add some test cases for error-prone Unicode characters?

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,461 @@
...
+			![Mercury]\(mercury)
+
+			[mercury]: https://example.com/mercury/
+		`,
+		dedent`
+			[foo bar](foo bar)
+			![foo bar](foo bar)
+
+			[foo bar]: https://example.com/foo-bar
+		`,
```
lumirlumir:
```suggestion
		`,
		{
			code: dedent`
				<http://foo>

				[http://foo]: hi
			`,
			language: "markdown/gfm",
		},
		{
			code: dedent`
				http://foo

				[http://foo]: hi
			`,
			language: "markdown/gfm",
		},
```

Currently, we have false positives when autolink literals are used in GFM.

Could you take a look at it?

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,461 @@
...
+				{
+					messageId: "referenceLikeUrl",
+					data: { type: "image", prefix: "!" },
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 16,
+				},
+			],
+		},
```
lumirlumir:
```suggestion
		},
		{
			code: dedent`
				![Venus](VENUS)

				[venus]: https://example.com/venus.jpg
			`,
			output: dedent`
				![Venus][VENUS]

				[venus]: https://example.com/venus.jpg
			`,
			errors: [
				{
					messageId: "referenceLikeUrl",
					data: { type: "image", prefix: "!" },
					line: 1,
					column: 1,
					endLine: 1,
					endColumn: 16,
				},
			],
		},
		{
			code: dedent`
				![Venus](VeNuS)

				[venus]: https://example.com/venus.jpg
			`,
			output: dedent`
				![Venus][VeNuS]

				[venus]: https://example.com/venus.jpg
			`,
			errors: [
				{
					messageId: "referenceLikeUrl",
					data: { type: "image", prefix: "!" },
					line: 1,
					column: 1,
					endLine: 1,
					endColumn: 16,
				},
			],
		},
```

Could you add these test cases?

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,461 @@
...
+					messageId: "referenceLikeUrl",
+					data: { type: "link", prefix: "" },
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 19,
+				},
+			],
+		},
+		{
```
lumirlumir:
```suggestion
		},
		{
			code: dedent`
				[Mercury](MeRcUrY)

				[mercury]: https://example.com/mercury
			`,
			output: dedent`
				[Mercury][MeRcUrY]

				[mercury]: https://example.com/mercury
			`,
			errors: [
				{
					messageId: "referenceLikeUrl",
					data: { type: "link", prefix: "" },
					line: 1,
					column: 1,
					endLine: 1,
					endColumn: 19,
				},
			],
		},
		{
```

Could you add this test case?

docs/rules/no-reference-like-url.md
lumirlumir, line 1: I think it would be good to include a description stating that autofix is not performed when links or images have a title.

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,461 @@
...
+			code: dedent`
+				[Mercury](mercury "")
+
+				[mercury]: https://example.com/mercury
+			`,
+			output: dedent`
+				[Mercury][mercury]
+
+				[mercury]: https://example.com/mercury
+			`,
```
lumirlumir: I think that auto-fixing only when the title is empty is not consistent behavior.

Examples like in the screenshot below don’t perform auto-fixing:

image

Is there any reason to perform auto-fix only when the title is empty? I think it can bring some confusion to users.

tests/rules/no-reference-like-url.test.js
lumirlumir:
```suggestion
		`,
		dedent`
		        [link](<uri>)

   		        [uri]: https://example.com/uri
		`,
		dedent`
			![image](<uri>)

			[uri]: https://example.com/uri
		`,
```

Another false positives occur when the URI is wrapped in brackets.

lumirlumir (changes requested):

tests/rules/no-reference-like-url.test.js
TKDev7: In the AST, links with no titles or empty ones both have null as the value for title. Therefore, I think it's safe to remove empty titles. What do you think?

tests/rules/no-reference-like-url.test.js
TKDev7: Links can be wrapped in angle brackets (<>), but these are stripped in the AST representation. If we want to handle this, I believe the only solution is regex-based parsing. However, since this seems like an edge case, I’m not sure it’s worth the effort.

tests/rules/no-reference-like-url.test.js
lumirlumir: Ah, I missed that case. Thank you for pointing it out.

In this case, I think it would be safe.

Instead, would it be possible to add a brief comment about this in the section where the logic is implemented?

tests/rules/no-reference-like-url.test.js
lumirlumir: Since the [AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiZm9vID8gYmFyIDogYmF6ID09PSBxdXggPyBxdXh4IDogZm9vYmFyO1wiLFwianNvblwiOlwiKGEgPT09IDEgPyB0cnVlIDogZmFsc2UpXFxuICA%2FIHRydWUgXFxuICA6IGZhbHNlXCIsXCJtYXJrZG93blwiOlwiW2hpXVs8aGk%2BXVxcblxcbls8aGk%2BXTogaGlcIixcImNzc1wiOlwiXCIsXCJodG1sXCI6XCI8IS0tXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb241XCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwiaHRtbE9wdGlvbnNcIjp7XCJ0ZW1wbGF0ZUVuZ2luZVN5bnRheFwiOlwibm9uZVwiLFwiZnJvbnRtYXR0ZXJcIjpmYWxzZX0sXCJ3cmFwXCI6dHJ1ZSxcInZpZXdNb2Rlc1wiOntcImFzdFZpZXdcIjpcInRyZWVcIixcInNjb3BlVmlld1wiOlwiZmxhdFwiLFwicGF0aFZpZXdcIjpcImdyYXBoXCJ9LFwicGF0aEluZGV4XCI6e1wiaW5kZXhcIjowLFwiaW5kZXhlc1wiOjJ9LFwiZXNxdWVyeVNlbGVjdG9yXCI6e1wic2VsZWN0b3JcIjpcIlwifX0sXCJ2ZXJzaW9uXCI6MH0i) below produces a correct `LinkReference` and `Definition` pair, I think it's worth the effort.

Image

On the other hand, the following [AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiZm9vID8gYmFyIDogYmF6ID09PSBxdXggPyBxdXh4IDogZm9vYmFyO1wiLFwianNvblwiOlwiKGEgPT09IDEgPyB0cnVlIDogZmFsc2UpXFxuICA%2FIHRydWUgXFxuICA6IGZhbHNlXCIsXCJtYXJrZG93blwiOlwiW2hpXVs8aGk%2BXVxcblxcbltoaV06IGhpXCIsXCJjc3NcIjpcIlwiLFwiaHRtbFwiOlwiPCEtLVwifSxcImxhbmd1YWdlXCI6XCJtYXJrZG93blwiLFwianNPcHRpb25zXCI6e1wicGFyc2VyXCI6XCJlc3ByZWVcIixcInNvdXJjZVR5cGVcIjpcIm1vZHVsZVwiLFwiZXNWZXJzaW9uXCI6XCJsYXRlc3RcIixcImlzSlNYXCI6dHJ1ZX0sXCJqc29uT3B0aW9uc1wiOntcImpzb25Nb2RlXCI6XCJqc29uNVwiLFwiYWxsb3dUcmFpbGluZ0NvbW1hc1wiOmZhbHNlfSxcImNzc09wdGlvbnNcIjp7XCJjc3NNb2RlXCI6XCJjc3NcIixcInRvbGVyYW50XCI6ZmFsc2V9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJnZm1cIixcIm1hcmtkb3duRnJvbnRtYXR0ZXJcIjpcIm9mZlwifSxcImh0bWxPcHRpb25zXCI6e1widGVtcGxhdGVFbmdpbmVTeW50YXhcIjpcIm5vbmVcIixcImZyb250bWF0dGVyXCI6ZmFsc2V9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoyfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D) does not yield a correct `LinkReference` and `Definition` pair:

Image

---

> If we want to handle this, I believe the only solution would be regex-based parsing.

I think this approach makes sense, since many other Markdown rules already use regex-based parsing.

TKDev7: I think there might be a misunderstanding here. `[Mercury](<mercury>)` will be autofixed to `[Mercury][mercury]`, not `[Mercury][<mercury>]`. This behavior aligns with the [spec](https://spec.commonmark.org/0.31.2/#inline-link) which states:

> The link’s URI consists of the link destination, excluding enclosing <...> if present ...

Based on my interpretation of the spec, the cases above are not false positives. What do you think?

lumirlumir: [Based on my understanding](https://github.com/eslint/markdown/issues/419#issuecomment-2969374348), I believe the purpose of this rule is to prevent error-prone syntax that users often find confusing.

This rule assumes that an identifier-like keyword in the `Link` node syntax is likely a mistake, and converts it to `LinkReference` syntax.

So, I believe automatically removing `<>` is an unintended behavior.

> `[Mercury](<mercury>)` is automatically fixed to `[Mercury][mercury]`, not `[Mercury][<mercury>]`.

I think this behavior is incorrect, and it should actually be fixed to `[Mercury][<mercury>]`.

My first example demonstrated this as the expected behavior:

Image

Also, the test case below does not throw an error, which seems to be a false negative:

image

> This behavior aligns with the [spec](https://spec.commonmark.org/0.31.2/#inline-link) which states:

I don’t think this is related to the Link spec, since this rule is specifically for checking error-prone syntax.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,190 @@
...
+ * @typedef {[]} NoReferenceLikeUrlOptions
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoReferenceLikeUrlOptions, MessageIds: NoReferenceLikeUrlMessageIds }>} NoReferenceLikeUrlRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+// Pattern to match both inline links: [text](url) and images: ![alt](url), with optional title
+const linkOrImagePattern =
```
lumirlumir: Would it be possible to use named capture groups with the `?<name>` syntax for clearer regex handling? The current regex is quite long, so it's a bit difficult to understand what the logic is doing.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,190 @@
...
+ * @typedef {"referenceLikeUrl"} NoReferenceLikeUrlMessageIds
+ * @typedef {[]} NoReferenceLikeUrlOptions
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoReferenceLikeUrlOptions, MessageIds: NoReferenceLikeUrlMessageIds }>} NoReferenceLikeUrlRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+// Pattern to match both inline links: [text](url) and images: ![alt](url), with optional title
```
lumirlumir:
```suggestion
/** Pattern to match both inline links: `[text](url)` and images: `![alt](url)`, with optional title */
```

Nitpick. I think a JSDoc comment would be more appropriate here.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,190 @@
...
+	);
+}
+
+/**
+ * Finds all code span and HTML block ranges in the AST.
+ * @param {Root} node The root AST node
+ * @returns {Array<{startOffset: number, endOffset: number}>} Array of skip ranges
+ */
+function findSkipRanges(node) {
+	const skipRanges = [];
```
lumirlumir: Can we add a type annotation here?

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,190 @@
...
+		const { sourceCode } = context;
+		/** @type {Set<string>} */
+		const definitionIdentifiers = new Set();
+
+		return {
+			definition(node) {
+				definitionIdentifiers.add(node.identifier);
+			},
+
+			"root:exit"(node) {
```
lumirlumir: Checking the entire `Root` node doesn’t seem necessary to me.

If we look at the whole document, there are many nodes that should be ignored.  
Currently, line 59 only ignores `inlineCode` and `html` nodes, as shown below:

```js
if (currentNode.type === "inlineCode" || currentNode.type === "html") 
```

However, if we consider the whole document, the list of nodes to ignore grows significantly, since there are several nodes where link syntax can’t be used. For example, `Code` nodes.

It would make more sense to focus on just `Paragraph`, `Heading`, and `TableCell` nodes instead of the entire document.

---

It would also be nice to add tests for `Heading` and `TableCell` nodes.

lumirlumir (changes requested):

src/rules/no-reference-like-url.js
TKDev7: Errors should be reported only after all definition nodes have been collected, since definition nodes may appear in any order within the document. With your approach, this would result in a false negative:

```markdown
[Mercury](mercury)

```

lumirlumir: Ah, you're right. We need to report the nodes after the `definition` nodes have been collected.

However, what's in my mind is that looking at the whole document could result in false positives, since there are many nodes that should be ignored and some nodes can't contain `Link` or `*Reference` node types.

This approach could also be costly, as it would require matching the entire document with regex. I think we should narrow down the scope.

---

I think we can collect `Paragraph`, `Heading`, and `TableCell` nodes as we enter them (using `paragraph(node)`, `heading(node)`, and `tableCell(node)`), and then report only on those nodes when we exit the root node using `root:exit()`.

The location and text information can be retrieved from the nodes we've collected as we enter them.

src/rules/no-reference-like-url.js
lumirlumir, line 1: Could we update the rule name to use the plural form? I think naming it `no-reference-links-urls` with an `s` at the end would be more consistent, since other rules like `no-bare-urls`, `no-empty-images`, and `no-empty-links` also use the plural form. (I missed it in the initial review.)

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Checks if a given index is within any skip range.
+ * @param {number} index The index to check
+ * @param {Array<{startOffset: number, endOffset: number}>} skipRanges The skip ranges
+ * @returns {boolean} True if index is in a skip range
+ */
+function isInSkipRange(index, skipRanges) {
+	return skipRanges.some(
+		range => index >= range.startOffset && index < range.endOffset,
```
lumirlumir:
```suggestion
		range => range.startOffset <= index && index < range.endOffset,
```

I think the form above would make the inequality easier to read.

src/rules/no-reference-like-url.js
```diff
@@ -0,0 +1,216 @@
...
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoReferenceLikeUrlOptions, MessageIds: NoReferenceLikeUrlMessageIds }>} NoReferenceLikeUrlRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/** Pattern to match both inline links: `[text](url)` and images: `![alt](url)`, with optional title */
+const linkOrImagePattern =
+	/(?<!\\)(?<imageBang>!)?\[(?<label>(?:\\.|[^()\\]|\([\s\S]*\))*?)\]\((?<destination>(?:<[^>]*>)|(?:[^ \t)]+))(?:[ \t]+(?<title>"[^"]*"|'[^']*'|\([^)]*\)))?\)(?!\()/gu;
```
lumirlumir:
```suggestion
	/(?<!(?<!\\)\\)(?<imageBang>!)?\[(?<label>(?:\\.|[^()\\]|\([\s\S]*\))*?)\]\((?<destination>(?:<[^>]*>)|(?:[^ \t)]+))(?:[ \t]+(?<title>"[^"]*"|'[^']*'|\([^)]*\)))?\)(?!\()/gu;
```

Would it be possible to update the negative lookbehind in the regex? Right now, we're seeing false negatives when consecutive backslashes appear.

Adding a test case like the one below could help highlight the current issue.

```js
		{
			code: dedent`
				\\\\\![Mercury](mercury)

				[mercury]: https://example.com/mercury/
			`,
			output: dedent`
				\\\\\![Mercury][mercury]

				[mercury]: https://example.com/mercury/
			`,
			errors: [
				{
					data: { type: "image", prefix: "!" },
					line: 1,
					column: 6,
					endLine: 1,
					endColumn: 25,
				},
			],
		},
```

Just a heads up, consecutive `\\\\` in JavaScript code is interpreted as `\\`, just like in Markdown. This means the `\` is treated as a regular character, not as an escape character.

Image

You can find more detailed concerns about this behavior here: https://github.com/eslint/markdown/pull/452#discussion_r2202739441

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,936 @@
...
+		`,
+		dedent`
+			\[Mercury](mercury)
+			[Mercury\](mercury)
+			![Mercury\](mercury)
+			[Mercury]\(mercury)
+			![Mercury]\(mercury)
+
+			[mercury]: https://example.com/mercury/
+		`,
```
lumirlumir: I think this test case should be reported, since `\[` in JavaScript code is treated the same as a simple `[` character.

Image

Also, the AST looks just like a combination of `Link` and `Image` nodes, so it need to be reported.

Image

tests/rules/no-reference-like-url.test.js
```diff
@@ -0,0 +1,936 @@
...
+		dedent`
```
lumirlumir:
```suggestion
		dedent`
			\\[Mercury](mercury)
			[Mercury\\](mercury)
			![Mercury\\](mercury)
			[Mercury]\\(mercury)
			![Mercury]\\(mercury)

			[mercury]: https://example.com/mercury/
		`,
		dedent`
```

Adding a test case to confirm that the rule correctly ignores backslash escape characters would be helpful here.

tests/rules/no-reference-like-url.test.js
TKDev7: Yes, but this isn’t a plain string—we’re using `dedent`, which treats the template literal literally and preserves escape sequences. So ``console.log(dedent`\[Mercury](mercury)`);`` prints `\[Mercury](mercury)`.

tests/rules/no-reference-like-url.test.js
TKDev7: This is invalid. https://github.com/eslint/markdown/pull/433#discussion_r2239852154

tests/rules/no-reference-like-url.test.js
lumirlumir: Ah, thank you for pointing that out. Sometimes the internal behavior of `dedent` is a bit confusing.

Yes, let's keep it this way.

tests/rules/no-reference-like-urls.test.js
```diff
@@ -0,0 +1,980 @@
...
+		`,
+		dedent`
+			[foo bar](foo bar)
+			![foo bar](foo bar)
+
+			[foo bar]: https://example.com/foo-bar
+		`,
+		dedent`
+		    [link](<uri>)
+   		    [uri]: https://example.com/uri
```
lumirlumir:
```suggestion
		    [link](<uri>)
		    
   		    [uri]: https://example.com/uri
```

I think we need a blank line here, since the AST doesn't recognize this as a valid `Link` and `Definition` pair.

[AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLy8gZmlyc3QgbGluZVxcbi8vIHNlY29uZCBsaW5lXCIsXCJqc29uXCI6XCIoYSA9PT0gMSA%2FIHRydWUgOiBmYWxzZSlcXG4gID8gdHJ1ZSBcXG4gIDogZmFsc2VcIixcIm1hcmtkb3duXCI6XCJbbGlua10oPHVyaT4pXFxuW3VyaV06IGh0dHBzOi8vZXhhbXBsZS5jb20vdXJpXCIsXCJjc3NcIjpcIlwiLFwiaHRtbFwiOlwiPCEtLVwifSxcImxhbmd1YWdlXCI6XCJtYXJrZG93blwiLFwianNPcHRpb25zXCI6e1wicGFyc2VyXCI6XCJlc3ByZWVcIixcInNvdXJjZVR5cGVcIjpcIm1vZHVsZVwiLFwiZXNWZXJzaW9uXCI6XCJsYXRlc3RcIixcImlzSlNYXCI6dHJ1ZX0sXCJqc29uT3B0aW9uc1wiOntcImpzb25Nb2RlXCI6XCJqc29uNVwiLFwiYWxsb3dUcmFpbGluZ0NvbW1hc1wiOmZhbHNlfSxcImNzc09wdGlvbnNcIjp7XCJjc3NNb2RlXCI6XCJjc3NcIixcInRvbGVyYW50XCI6ZmFsc2V9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJnZm1cIixcIm1hcmtkb3duRnJvbnRtYXR0ZXJcIjpcIm9mZlwifSxcImh0bWxPcHRpb25zXCI6e1widGVtcGxhdGVFbmdpbmVTeW50YXhcIjpcIm5vbmVcIixcImZyb250bWF0dGVyXCI6ZmFsc2V9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D)

Image

tests/rules/no-reference-like-urls.test.js
```diff
@@ -0,0 +1,980 @@
...
+		`,
+		dedent`
+			![image](<uri>)
+			[uri]: https://example.com/uri
```
lumirlumir:
```suggestion
			![image](<uri>)
			
			[uri]: https://example.com/uri
```

Same with https://github.com/eslint/markdown/pull/433/files#r2250920796

tests/rules/no-reference-like-urls.test.js
```diff
@@ -0,0 +1,980 @@
...
+					endLine: 4,
+					endColumn: 32,
+				},
+			],
+		},
+		// This test case is skipped for non-Node environments like Bun
+		...(typeof process !== "undefined" &&
+		process.release &&
+		process.release.name === "node" &&
+		(!process.versions || !process.versions.bun)
```
lumirlumir: As we did in https://github.com/eslint/markdown/pull/488#discussion_r2242154517, would it be possible to simplify this a bit using optional chaining?

lumirlumir (changes requested):

snitin315 (commented):
The following case doesn't report any errors, is this expected behavior?

```md
<!-- eslint markdown/no-reference-like-urls: "error" -->

[Mercury](  mercury) is the first planet from the sun.

[mercury]: https://example.com/mercury/

```

src/rules/no-reference-like-urls.js
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		if ("children" in currentNode && Array.isArray(currentNode.children)) {
+			currentNode.children.forEach(traverse);
+		}
+	}
+
+	traverse(node);
+	return skipRanges;
+}
```
snitin315: Do we really need this util? I think we can remove it because ESLint itself traverses the tree no need of additional traversing. Please refer https://github.com/eslint/markdown/pull/498/files

snitin315 (changes requested):

src/rules/no-reference-like-urls.js
lumirlumir: We could also use the `sourceCode.getRange(node)` method and the `SourceRange` type to make the rule feel more native.

PR #434, opened 2025-06-23
feat: add checkMissingCells option to table-column-count

### LINKED ISSUE #422: Rule Change: Extend table-column-count to flag rows with fewer cells than header

table-column-count

Generate more warnings

A new option

### Example code

```markdown
| Header | Header | Header |
| ------ | ------ | ------ |
| Cell   | Cell   |        |  <!-- Currently VALID: fewer cells than header -->
```

The row with fewer cells is considered valid.

The rule will flag rows with fewer cells than the header.

### Participation

### Additional comments

This change mirrors Markdownlint's `MD056` rule behavior, which flags both excess and missing cells. As for the option's name, I suggest `checkMissingCells`.

TKDev7:
#### Prerequisites checklist

This pull request introduces a new option, checkMissingCells, to the table-column-count rule. The purpose of this addition is to provide users with more granular control over how the rule handles tables with missing cells, allowing for stricter or more flexible enforcement as needed.

- Added the checkMissingCells option to table-column-count
- Updated rule logic, docs, and tests to support the new option

#### Related Issues

Fixes #422

### DIFF
--- a/docs/rules/table-column-count.md
+++ b/docs/rules/table-column-count.md
@@ -43,7 +43,7 @@ Examples of **correct** code for this rule:
 <!-- Rows with fewer cells are valid because they render correctly and no data is lost -->
 | Header | Header | Header |
 | ------ | ------ | ------ |
-| Cell   | Cell   |
+| Cell   | Cell   |        |
 
 <!-- Table with some empty cells (VALID for this rule) -->
 <!-- Missing cells are treated as empty and don't cause rendering issues -->
@@ -70,9 +70,10 @@ Examples of **incorrect** code when configured as `"table-column-count": ["error
 <!-- eslint markdown/table-column-count: ["error", { checkMissingCells: true }] -->
 
 <!-- Data row with fewer cells than header -->
-| Header | Header | Header |
-| ------ | ------ | ------ |
-| Cell   | Cell   |
+| Col A | Col B | Col C |
+| ----- | ----- | ----- |
+| 1     |       | 3     |
+| 4     | 5     |
 ```
 
 ## When Not To Use It


src/rules/table-column-count.js
```diff
@@ -30,10 +31,28 @@ export default {
 		messages: {
 			inconsistentColumnCount:
 				"Table column count mismatch (Expected: {{expectedCells}}, Actual: {{actualCells}}), extra data starting here will be ignored.",
+			missingCells:
+				"Table column count mismatch (Expected: {{expectedCells}}, Actual: {{actualCells}}), row will be missing data.",
```
snitin315: We can't be sure.

```suggestion
				"Table column count mismatch (Expected: {{expectedCells}}, Actual: {{actualCells}}), row might be missing data.",
```

src/rules/table-column-count.js
```diff
@@ -65,6 +84,29 @@ export default {
...
+							currentRow.children[actualCellsLength - 1];
+						const rowEnd = currentRow.position.end;
+
+						context.report({
+							loc: {
+								start: {
+									column:
+										lastCellNode.position.end.column - 1,
+									line: lastCellNode.position.end.line,
+								},
```
snitin315: The locations seems incorrect when we have whitespaces:

Screenshot 2025-06-23 at 10 08 58 AM

This is reporting error on column `16` but it should report on `11`

TKDev7: This behavior is also present in Markdownlint. If you check the AST, the whitespace affects the previous cell's position, which explains the reported location.

snitin315: Oh! Got it.

snitin315 (approved):
LGTM, would like another review before merge.

src/rules/table-column-count.js
```diff
@@ -8,7 +8,8 @@
 //-----------------------------------------------------------------------------
 
 /**
- * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: []; }>} TableColumnCountRuleDefinition
+ * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ checkMissingCells?: boolean; }]; }>}
```
lumirlumir:
```suggestion
 * @typedef {import("../types.ts").MarkdownRuleDefinition<{ RuleOptions: [{ checkMissingCells?: boolean; }]; }>}
```

image

docs/rules/table-column-count.md
```diff
@@ -43,7 +43,7 @@ Examples of **correct** code for this rule:
 <!-- Rows with fewer cells are valid because they render correctly and no data is lost -->
 | Header | Header | Header |
 | ------ | ------ | ------ |
-| Cell   | Cell   |        |
+| Cell   | Cell   |
```
lumirlumir:
```suggestion
| Cell   | Cell   |        |
```

I think the original intention of this example is to show that an empty cell is okay. The following example already covers the change you made.

image


Could you revert the example?

docs/rules/table-column-count.md
```diff
@@ -58,6 +58,23 @@ Examples of **correct** code for this rule:
 | Single Cell   |
 ```
 
+## Options
+
+The following options are available on this rule:
+
+* `checkMissingCells: boolean` - When set to `true`, the rule will also flag rows that have fewer cells than the header row. Default is `false`.
```
lumirlumir:
```suggestion
* `checkMissingCells: boolean` - When set to `true`, the rule will also flag rows that have fewer cells than the header row. (default: `false`)
```

Could you use `(default: false)` instead of `Default is false`, as discussed in https://github.com/eslint/markdown/pull/447#discussion\_r2171927150?

docs/rules/table-column-count.md
```diff
@@ -58,6 +58,23 @@ Examples of **correct** code for this rule:
...
+
+Examples of **incorrect** code for this rule with `checkMissingCells: true`:
```
lumirlumir:
```suggestion
Examples of **incorrect** code when configured as `"table-column-count": ["error", { checkMissingCells: true }]`:
```

image

Could you follow the convention used in the other documentation when describing this option?

src/rules/table-column-count.js
```diff
@@ -30,10 +33,28 @@ export default {
 		messages: {
 			inconsistentColumnCount:
```
lumirlumir:
```suggestion
			extraCells:
```

I think we can rename it to `extraCells` here. Previously, `inconsistentColumnCount` made sense, but now we have two messages with conflicting meanings.

src/rules/table-column-count.js
```diff
@@ -65,6 +86,29 @@ export default {
...
+									line: lastCellNode.position.end.line,
+								},
+								end: rowEnd,
+							},
+							messageId: "missingCells",
+							data: {
+								actualCells: String(actualCellsLength),
+								expectedCells: String(expectedCellsLength),
+							},
+						});
```
lumirlumir: I think we can refactor this code slightly, as shown in the screenshot below, because `lastActualCellNode` and `lastCellNode` serve the same purpose, and `rowEnd` is used only once.

image

lumirlumir (changes requested):

docs/rules/table-column-count.md
TKDev7: Its description says: `<!-- Data row with fewer cells than header (VALID for this rule) -->`. Without this change, the number of cells would have matched the header.

src/rules/table-column-count.js
TKDev7: At first, I wanted to do that, but I thought it might be a breaking change. However, since the next version is a major release, I think we can go ahead with it.

nzakas:
@TKDev7 can you take a look at the merge conflict?

@lumirlumir this looks ready for you to re-review.

docs/rules/table-column-count.md
lumirlumir: I think we need to keep both cases in the document as shown below:

Image

If your suggestion applies, I believe updating the Markdown comment would make things clearer than changing the example.

---

Also, I think the `Options` section need to be updated to follow the example shown here.

TKDev7: If we modify the example under Options to:
```markdown
| Header | Header | Header |
| ------ | ------ | ------ |
| Cell   | Cell   |        |
```
then it no longer reports an error. This table contains an empty cell, not a missing cell.

lumirlumir: Ah, I meant that the example under `Options` is currently using a modified version of the example below, like this:

image

My suggestion is to update it to use an example like this:

image

That way, the example could be revised as follows:

```diff
<!-- Data row with fewer cells than header -->
- | Header | Header | Header |
- | ------ | ------ | ------ |
- | Cell   | Cell   |
+ | Col A | Col B | Col C |
+ | ----- | ----- | ----- |
+ | 1     |       | 3     |
+ | 4     | 5     |
```

By using the same example both with and without the option, I think it makes it much clearer for users to understand how the option works.

PR #451, opened 2025-06-27
chore: Update bug-report.yml

nzakas:
#### Prerequisites checklist

Adds a StackBlitz template URL to the bug report issue template.


lumirlumir (approved):
LGTM, Thanks!

Just a small suggestion from my side:

Currently, the version in StackBlitz is set to `6.6.0` and `9.28.0`. Would it be nicer to make this version to simply just `latest`? The versions can keep change, and if there is major version update, I think we have to update the versions manually.

image

PR #450, opened 2025-06-27
feat: add `checkClosedHeadings` option in no-missing-atx-heading-space

SwetaTanwar:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What is the purpose of this pull request?
 Add new option `checkClosedHeadings`.

#### What changes did you make? (Give an overview)

This PR supports the following options:

* `checkClosedHeadings: boolean` -
    When `true`, this option also checks for missing spaces before the closing `#` characters in ATX-style headings. (default: `false`)

    Examples of **incorrect** code when configured as `"no-missing-atx-heading-space": ["error", { checkClosedHeadings: true }]`:

    ```markdown
    <!-- eslint markdown/no-missing-atx-heading-space: ["error", { checkClosedHeadings: true }] -->
    
    # Heading 1#
    ## Heading 2##
    ### Heading 3###
    
    ```

#### Related Issues

Fix #409 

#### Is there anything you'd like reviewers to focus on?
No

PR #452, opened 2025-06-27
feat: add checkClosedHeadings option to no-missing-atx-heading-space

### LINKED ISSUE #409: Rule Change: Add check for missing spaces before closing hashes in ATX headings

no-missing-atx-heading-space

Generate more warnings

A new default behavior

### Example code

```markdown
#Heading 1#
```

Currently, the rule only reports an error for `#Heading` (missing space after opening hash) but ignores `1#` (missing space before closing hash).

After the change, the rule will report errors for missing space before closing hash

### Participation

### Additional comments

This behavior aligns with Markdownlint's implementation (rule MD018), which also checks for spaces both before and after hash sequences in ATX headings.

xbinaryx:
#### Prerequisites checklist

This pull request introduces a new option, checkClosedHeadings, to the no-missing-atx-heading-space rule. The purpose is to enhance the rule’s ability to detect missing spaces before closing hash sequences in ATX-style Markdown headings, in addition to the existing check for missing spaces after opening hashes.

- Added a checkClosedHeadings option to the no-missing-atx-heading-space rule.
- Updated the rule’s logic to support this additional check for both heading and paragraph nodes.

#### Related Issues

Fixes #409

### DIFF
--- a/src/rules/no-missing-atx-heading-space.js
+++ b/src/rules/no-missing-atx-heading-space.js
@@ -19,7 +19,8 @@
 //-----------------------------------------------------------------------------
 
 const leadingAtxHeadingHashPattern = /^(#{1,6})(?:[^# \t]|$)/u;
-const trailingAtxHeadingHashPattern = /(?<![ \t])([ \t]*)(?<!\\)(#+)([ \t]*)$/u;
+const trailingAtxHeadingHashPattern =
+	/(?<![ \t])([ \t]*)(?<!(?<!\\)\\)(#+)([ \t]*)$/u;
 const newLinePattern = /\r?\n/u;
 
 /**
--- a/tests/rules/no-missing-atx-heading-space.test.js
+++ b/tests/rules/no-missing-atx-heading-space.test.js
@@ -760,6 +760,21 @@ const invalidTests = [
 			},
 		],
 	},
+	{
+		code: "# Heading 1\\\\#",
+		output: "# Heading 1\\\\ #",
+		options: [{ checkClosedHeadings: true }],
+		errors: [
+			{
+				messageId: "missingSpace",
+				data: { position: "before" },
+				line: 1,
+				column: 13,
+				endLine: 1,
+				endColumn: 15,
+			},
+		],
+	},
 	// Note: Closing ATX patterns (e.g., missing space before trailing #) are reported in the next iteration and will be fixed. This behavior is consistent with Markdownlint.
 	{
 		code: "#Heading 1#",


nzakas:
@xbinaryx can you take a look at the merge conflict?

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -504,13 +608,425 @@ const invalidTests = [
...
 			{
 				messageId: "missingSpace",
+				data: { position: "after" },
 				column: 1,
 				line: 1,
 				endLine: 1,
 				endColumn: 4,
 			},
 		],
 	},
```
snitin315, line 642: Can we add following test case:

```md
# heading [#123](issue-link)#
```

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -504,13 +608,425 @@ const invalidTests = [
...
+			{
+				messageId: "missingSpace",
+				data: { position: "before" },
+				line: 1,
+				column: 10,
+				endLine: 1,
+				endColumn: 12,
+			},
+		],
+	},
```
snitin315: Should we report 2 errors here? Markdownlint only reports one for the prefix `#` at first.

src/rules/no-missing-atx-heading-space.js
```diff
@@ -96,8 +166,47 @@ export default {
...
+											" ",
+										);
+									},
+								});
+							}
+						}
+					}
+
+					offset += line.length + 1;
+				});
```
snitin315: I think we can also remove all this code if we just report one error like markdownlint for cases like `#heading#`.  So basically don't check for closed `#` until it's a valid heading node.

snitin315: Something similar to https://github.com/eslint/markdown/pull/450/files#r2173156648. WDYT?

snitin315 (dismissed):

tests/rules/no-missing-atx-heading-space.test.js
xbinaryx: There are two violations here: a missing space after the opening hash and a missing space before the closing hash. Reporting two errors makes sense, and this way, autofix can fix them at once.

docs/rules/no-missing-atx-heading-space.md
```diff
@@ -38,13 +38,40 @@ Examples of **correct** code for this rule:
 This is a paragraph with a #hashtag, not a heading.
 ```
 
+## Options
```
lumirlumir, line 41: Could we add an "end of the line" description, such as "using hash (#) characters at the beginning or end of the line," to the `Rule Details` section?

Image

src/rules/no-missing-atx-heading-space.js
```diff
@@ -10,17 +10,46 @@
...
  * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoMissingAtxHeadingSpaceOptions, MessageIds: NoMissingAtxHeadingSpaceMessageIds }>} NoMissingAtxHeadingSpaceRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
 // Helpers
 //-----------------------------------------------------------------------------
 
-const headingPattern = /^(#{1,6})(?:[^# \t]|$)/u;
+const atxHeadingPattern = /^(#{1,6})(?:[^# \t]|$)/u;
+const atxHeadingClosedPattern = /([ \t]*)(?<!\\)(#+)([ \t]*)$/u;
```
lumirlumir:
```suggestion
const leadingAtxHeadingHashPattern = /^(#{1,6})(?:[^# \t]|$)/u;
const trailingAtxHeadingHashPattern = /([ \t]*)(?<!\\)(#+)([ \t]*)$/u;
```

I think we could rename the variables to follow the naming convention below, or something similar, since the pre-implemented rule uses these names:

https://github.com/eslint/markdown/blob/56eb0ee4d1ffd81fe4eb7cec9dda27e35939842f/src/rules/no-duplicate-headings.js#L39-L40

lumirlumir (changes requested):

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -114,6 +114,84 @@ const validHeadings = [
...
+	{
+		code: "<h1>#valid heading#</h1>",
+		options: [{ checkClosedHeadings: true }],
+	},
+	{
+		code: '"#something#"',
+		options: [{ checkClosedHeadings: true }],
+	},
+	{
+		code: "    #Heading 1#",
```
lumirlumir:
```suggestion
		code: "    #Heading 1#", // Recognized as an indented code block.
```

Could you add this comment?

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -114,6 +114,84 @@ const validHeadings = [
...
+	{
+		code: "[#370#](https://github.com/eslint/markdown/issues/370)",
+		options: [{ checkClosedHeadings: true }],
+	},
+	{
+		code: "<h1>#valid heading#</h1>",
+		options: [{ checkClosedHeadings: true }],
+	},
+	{
+		code: '"#something#"',
```
lumirlumir, line 178: Can I ask why this code is considered valid?

Markdownlint reports this line as an error:

Image

Image

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -114,6 +114,84 @@ const validHeadings = [
...
+		code: dedent`# Regular heading
+		## Closed heading ##
+		### Another regular heading`,
+		options: [{ checkClosedHeadings: true }],
+	},
+	{
+		code: "# Heading with extra spaces  #",
+		options: [{ checkClosedHeadings: true }],
+	},
 ];
```
lumirlumir, line 219:
```suggestion
	},
	{
		code: "#Setext Heading\n===",
		options: [{ checkClosedHeadings: true }],
	},
	{
		code: "Setext Heading#\n===",
		options: [{ checkClosedHeadings: true }],
	},
	{
		code: "#Setext Heading#\n===",
		options: [{ checkClosedHeadings: true }],
	},
	{
		code: "#Setext Heading\n---",
		options: [{ checkClosedHeadings: true }],
	},
	{
		code: "Setext Heading#\n---",
		options: [{ checkClosedHeadings: true }],
	},
	{
		code: "#Setext Heading#\n---",
		options: [{ checkClosedHeadings: true }],
	},
];
```

Can we add these test cases?

src/rules/no-missing-atx-heading-space.js
lumirlumir: We have a false negative when a backslash is escaped with another backslash.

Image

Image

src/rules/no-missing-atx-heading-space.js
lumirlumir: Lines L170-L206 perform the same function as L104-L133.

It might be a good idea to refactor this code so it's only used once.

tests/rules/no-missing-atx-heading-space.test.js
xbinaryx: This is a string: `"#something#"`

src/rules/no-missing-atx-heading-space.js
xbinaryx: Markdownlint also doesn't report an error. It appears that a backslash does not act as an escape character for another backslash when it precedes the closing hash. This behavior is also visible in the AST.

tests/rules/no-missing-atx-heading-space.test.js
lumirlumir: ping @snitin315

src/rules/no-missing-atx-heading-space.js
```diff
@@ -10,17 +10,46 @@
...
+ * @returns {{ closingHashIdx: number, beforeHashIdx: number, endIdx: number } | null} The positions of the closing hashes in the heading, or null if no missing space is found.
+ */
+function findMissingSpaceBeforeClosingHash(text) {
+	const match = trailingAtxHeadingHashPattern.exec(text);
+
+	if (match) {
+		const [, closingSequenceSpaces, closingSequence, trailingSpaces] =
+			match;
+
+		if (!closingSequenceSpaces.length) {
```
lumirlumir:
```suggestion
		if (closingSequenceSpaces.length === 0) {
```

Just a small nitpick—using `=== 0` would make the intention of this code clearer.

tests/rules/no-missing-atx-heading-space.test.js
```diff
@@ -504,13 +632,432 @@ const invalidTests = [
...
+			{
+				messageId: "missingSpace",
+				data: { position: "after" },
+				line: 1,
+				column: 1,
+				endLine: 1,
+				endColumn: 3,
+			},
+		],
+	},
```
lumirlumir, line 793: Would it be possible to add a note saying that closing ATX patterns are reported in the next iteration and will be fixed, and that this behavior is consistent with Markdownlint?

lumirlumir (changes requested):
I’ve left a few comments about some nitpicks, Others are LGTM.

src/rules/no-missing-atx-heading-space.js
lumirlumir: > Markdownlint also doesn't report an error.

I believe Markdownlint misses this edge case as well.

---

For example, the JavaScript code `# Heading 1\\\\#` becomes `# Heading 1\\#` in the Markdown document.

Then, in rendering, `# Heading 1\\#` turns into `# Heading 1\#`, since the double backslash is escaped and the single backslash acts as a regular character, not as an escape character.

So, the `\` is treated as a normal character here, and I think it should be reported.

[AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogaGlcXG4gKi9cXG5oaVwiLFwianNvblwiOlwiKGEgPT09IDEgPyB0cnVlIDogZmFsc2UpXFxuICA%2FIHRydWUgXFxuICA6IGZhbHNlXCIsXCJtYXJrZG93blwiOlwiIyBIZWFkaW5nIDFcXFxcXFxcXCNcIixcImNzc1wiOlwiXCIsXCJodG1sXCI6XCI8IS0tXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb241XCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwiaHRtbE9wdGlvbnNcIjp7XCJ0ZW1wbGF0ZUVuZ2luZVN5bnRheFwiOlwibm9uZVwiLFwiZnJvbnRtYXR0ZXJcIjpmYWxzZX0sXCJ3cmFwXCI6dHJ1ZSxcInZpZXdNb2Rlc1wiOntcImFzdFZpZXdcIjpcInRyZWVcIixcInNjb3BlVmlld1wiOlwiZmxhdFwiLFwicGF0aFZpZXdcIjpcImdyYXBoXCJ9LFwicGF0aEluZGV4XCI6e1wiaW5kZXhcIjowLFwiaW5kZXhlc1wiOjF9LFwiZXNxdWVyeVNlbGVjdG9yXCI6e1wic2VsZWN0b3JcIjpcIlwifX0sXCJ2ZXJzaW9uXCI6MH0i)

image

image

---

It might be helpful to tweak the regex a bit, perhaps by adjusting the negative lookbehind if needed.

tests/rules/no-missing-atx-heading-space.test.js
xbinaryx: @lumirlumir I took @snitin315's approach since you 👍'd their [comment](https://github.com/eslint/markdown/pull/452#discussion_r2190698494). Did I interpret that right?

src/rules/no-missing-atx-heading-space.js
xbinaryx: So basically report if we have more than one backslash?

src/rules/no-missing-atx-heading-space.js
```diff
@@ -10,17 +10,46 @@
...
  * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoMissingAtxHeadingSpaceOptions, MessageIds: NoMissingAtxHeadingSpaceMessageIds }>} NoMissingAtxHeadingSpaceRuleDefinition
  */
 
 //-----------------------------------------------------------------------------
 // Helpers
 //-----------------------------------------------------------------------------
 
-const headingPattern = /^(#{1,6})(?:[^# \t]|$)/u;
+const leadingAtxHeadingHashPattern = /^(#{1,6})(?:[^# \t]|$)/u;
+const trailingAtxHeadingHashPattern = /(?<![ \t])([ \t]*)(?<!\\)(#+)([ \t]*)$/u;
```
snitin315: Regex are error prone, we are recently seeing many bugs and misses with regex. I see the approach in https://github.com/eslint/markdown/pull/450 is regex free can you evaluate that if that would work for all the test cases you have added?

src/rules/no-missing-atx-heading-space.js
lumirlumir: Ah, I meant we only need to report cases where the backslash is treated as a regular character.

The example I gave above is one such case where the backslash is recognized as a normal character.

(Backslash is treated as a normal character by escaping it with another backslash.)

---

Also, as mentioned in this comment (https://github.com/eslint/markdown/pull/452#discussion_r2227251285), if there's another way to achieve the same behavior without using regex, that would be a great option as well.

tests/rules/no-missing-atx-heading-space.test.js
lumirlumir: Ah, sorry for the misunderstanding. I thought this comment was still unresolved.

Thanks for pointing that out 👍

src/rules/no-missing-atx-heading-space.js
xbinaryx: The approach in #450 uses `indexOf` to find the hash, which is error-prone. For example, it retrieves the wrong hash in `# Heading #1#`.

lumirlumir (approved):
LGTM, thanks.

Would like @snitin315 and @nzakas to verify the changes before merging.

nzakas (approved):
LGTM. Thanks!

PR #453, opened 2025-06-30
refactor: add write-types-export script

JoshuaKGoldberg:
Throwing up as a reference for one strategy to solve the types re-export issue.


lumirlumir:
Hi @JoshuaKGoldberg 😄

I've taken another shot at this and wanted to share a different approach.

Here's what I did:

- First, I created an empty `types.js` file at `src/types.js`:

```js
export default {};
```

- Then, I added `export * from "./types.js";` to the end of `index.js`:

```diff
// ...

export default plugin;
export { MarkdownSourceCode };
+ export * from "./types.js";
```

---

This approach seems to work well, since it allows all types from `src/types.ts` to be imported directly from the `@eslint/markdown` package, rather than from `@eslint/markdown/types`.

image

---

What do you think of this approach? I'd love to hear your thoughts 🚀

JoshuaKGoldberg:
Ha nice that's smart! I like it. I'd separately thought of proposing converting just `index.js` to `index.ts` so we can use the `export` syntax (not yet available in `.js` files: https://github.com/microsoft/TypeScript/issues/60831).

JoshuaKGoldberg:
That is a great question, and I don't know how or where to push it 😄. Mentioned in: https://github.com/eslint/eslint/issues/19920

lumirlumir:
Ah, thank you! If it's alright, let's wait a bit for the team's decision. 😄

After that, I think we can open a separate issue for it.

lumirlumir:
Hi, @JoshuaKGoldberg

I've created an issue in https://github.com/eslint/json/issues/130, and it's been decided to postpone this change until ESLint v10. After v10, we can drop CJS module support, since Node.js will natively support `require(ESM)`. This should make it much easier to adopt the change across all repositories. 👍

PR #460, opened 2025-07-14
fix: no-reversed-media-syntax add missing offset for code spans

DMartens:
#### Prerequisites checklist

Fix a bug in `no-reversed-media-syntax-inline` for ignoring text inside code blocks.

Add the offset of the paragraph when comparing whether the matched text is inside code spans.
This is required as the regular expression match index is relative to the paragraph and the offsets of the code span is absolute.
They only match up if the paragraph containing the code span is the first node.

#### Related Issues
I hope it is okay not create an issue for such a small change/fix.

-

### DIFF
--- a/tests/rules/no-reversed-media-syntax.test.js
+++ b/tests/rules/no-reversed-media-syntax.test.js
@@ -40,13 +40,9 @@ ruleTester.run("no-reversed-media-syntax", rule, {
 		`,
 		"`myobj.getFiles(test)[0]`",
 		dedent`
-		# Heading,
-		\`myobj.getFiles(test)[0]\`,
-		`,
-		dedent`
-		foo
-		# Heading,
-		\`myobj.getFiles(test)[0]\`,
+		Some long text about foo
+		# Split,
+		\`myobj.getFiles(test)[0]\`
 		`,
 		"&lpar;reversed&rpar;[link]",
 		"a &rpar; a &lpar; a &rpar;[a]~",


tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -39,6 +39,15 @@ ruleTester.run("no-reversed-media-syntax", rule, {
...
 		"`myobj.getFiles(test)[0]`",
+		dedent`
+		# Heading,
+		\`myobj.getFiles(test)[0]\`,
+		`,
+		dedent`
+		foo
+		# Heading,
+		\`myobj.getFiles(test)[0]\`,
+		`,
```
lumirlumir: Good catch, thanks 👍

Interestingly, the current test cases don’t detect the changed logic properly, since the `matchIndex` is ignored by the range of the `InlineCode` node by chance!

Image

So even if we revert this change, the test cases still pass 😄

---

```suggestion
		dedent`
		# Heading
		\`(test)[0]\`
		`,
		dedent`
		foo
		# Heading
		\`(test)[0]\`
		`,
```

Would it be possible to use the above test cases instead? If we modify them, the tests wouldn’t pass if we reverted this change.

Also, it’s not a technical issue, but I think the comma (`,`) was added by mistake. If it wasn’t intentional, it would be great if you could remove it.

lumirlumir (changes requested):

tests/rules/no-reversed-media-syntax.test.js
DMartens: You are right that it did not actuall test the change :sweat_smile: .
I removed the unnecessary first test case and added more text in the next.

lumirlumir (approved):
LGTM, thanks!

nzakas (approved):
LGTM. It's okay to not create an issue, but in the future, please use the same format for the PR description as the bug template (What did you do, what happened, what did you expect to happen, etc.). It makes it a lot easier to understand what you're trying to fix.

PR #461, opened 2025-07-14
fix: make `meta` property optional in `Block` type

lumirlumir:
#### Prerequisites checklist

In this PR, I've corrected the `Block` type used by `processor`.

Currently, the `Block` type extends the `Node` type and overrides the `meta` property.

However, when we look at its origin in the `processor.js` file, the `Block` type always uses information from the `Code` node, which is a more specific type than `Node` (The `Code` type extends `Node` and some other types, so it includes all properties from the `Node` type).

The `Code` node has its own *optional* `meta` property, and the issue arises when we override the `meta` property in the `Block` type.

image

In the logic of `processor.js`, when we push information to the `blocks` variable, we never re-declare the `meta` information. The `meta` information comes from the `Code` node itself, so the `meta` property should be optional, not required.

As a result, this causes a type error:

image

So, I've made the `meta` property optional.

FYI, if we set the `meta` property as optional, it becomes the same property that's defined in the `Code` node. So I've removed it and tweaked the type a bit.

I've corrected the `Block` type used by `processor`.

### DIFF
--- a/tests/types/types.test.ts
+++ b/tests/types/types.test.ts
@@ -13,8 +13,8 @@ import { Position } from "unist";
 import type {
 	// Nodes (abstract)
 	Node,
-	Data,
 	Parent,
+	CodeData,
 	// Nodes
 	Blockquote,
 	Break,
@@ -47,6 +47,7 @@ import type {
 } from "mdast";
 
 // Test that `Block` extends `Code` and `BlockBase` correctly
+// `meta` property is optional and not required in `Block`
 const validBlock: Block = {
 	// `Code` properties
 	type: "code",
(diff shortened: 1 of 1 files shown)


src/processor.js
```diff
@@ -95,7 +95,7 @@ const leadingWhitespaceRegex = /^[>\s]*/u;
 /**
  * Gets the offset for the first column of the node's first line in the
  * original source text.
- * @param {Node} node A Markdown code block AST node.
+ * @param {Code} node A Markdown code block AST node.
```
lumirlumir, line 98: I've also updated the `Node` types to use the more specific `Code` node type.

src/types.ts
```diff
@@ -74,9 +74,7 @@ export interface BlockBase {
 	rangeMap: RangeMap[];
 }
 
-export interface Block extends Node, BlockBase {
-	meta: string | null;
-}
+export type Block = Code & BlockBase;
```
lumirlumir, line 77: It no longer overrides or declares new properties—I've used an intersection type instead.

PR #469, opened 2025-07-16
fix: handle multiline setext headings in `no-duplicate-headings`

lumirlumir:
#### Prerequisites checklist

Hi,

In this PR, I've fixed the handling of multiline setext headings in the `no-duplicate-headings` rule.

The current logic cannot recognize multiline setext headings, which has resulted in false positives, as shown in the examples below:

```md
Heading
Hi
===

Heading
Bye
===
```

According to the [CommonMark Specification](https://spec.commonmark.org/0.31.2/#setext-heading), setext headings can span more than two lines of text.

For example, the following setext heading is valid:

[AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogaGlcXG4gKi9cXG5oaVwiLFwianNvblwiOlwiKGEgPT09IDEgPyB0cnVlIDogZmFsc2UpXFxuICA%2FIHRydWUgXFxuICA6IGZhbHNlXCIsXCJtYXJrZG93blwiOlwiSGVsbG9cXG5Xb3JsZFxcbiFcXG49PT1cIixcImNzc1wiOlwiXCIsXCJodG1sXCI6XCI8IS0tXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb241XCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImdmbVwiLFwibWFya2Rvd25Gcm9udG1hdHRlclwiOlwib2ZmXCJ9LFwiaHRtbE9wdGlvbnNcIjp7XCJ0ZW1wbGF0ZUVuZ2luZVN5bnRheFwiOlwibm9uZVwiLFwiZnJvbnRtYXR0ZXJcIjpmYWxzZX0sXCJ3cmFwXCI6dHJ1ZSxcInZpZXdNb2Rlc1wiOntcImFzdFZpZXdcIjpcInRyZWVcIixcInNjb3BlVmlld1wiOlwiZmxhdFwiLFwicGF0aFZpZXdcIjpcImdyYXBoXCJ9LFwicGF0aEluZGV4XCI6e1wiaW5kZXhcIjowLFwiaW5kZXhlc1wiOjF9LFwiZXNxdWVyeVNlbGVjdG9yXCI6e1wic2VsZWN0b3JcIjpcIlwifX0sXCJ2ZXJzaW9uXCI6MH0i)

```md
Hello
World
!
===
```

---

I took a slightly different approach to fix this bug, as handling multiline setext headings can be tricky, especially with whitespace. I’ve included detailed comments about this below.

---

I’ve also made sure that all previous tests pass, so there wouldn’t be any concerns about backward compatibility.

### DIFF
--- a/src/rules/no-duplicate-headings.js
+++ b/src/rules/no-duplicate-headings.js
@@ -93,18 +93,13 @@ export default {
 
 			"heading *"({ type, value }) {
 				if (value) {
-					headingChildrenSequence += JSON.stringify({
-						type,
-						value,
-					});
+					headingChildrenSequence += `[${type},${value}]`; // We use a custom sequence representation to keep track of heading children.
 
 					if (type !== "html") {
 						headingText += value;
 					}
 				} else {
-					headingChildrenSequence += JSON.stringify({
-						type,
-					});
+					headingChildrenSequence += `[${type}]`;
 				}
 			},
 


package.json
```diff
@@ -90,10 +90,12 @@
   "dependencies": {
     "@eslint/core": "^0.15.0",
     "@eslint/plugin-kit": "^0.3.1",
+    "fast-deep-equal": "^3.1.3",
```
lumirlumir: The `fast-deep-equal` package is already used in [eslint](https://github.com/eslint/eslint/blob/main/package.json#L130), so it won't impact real users' dependencies.

---

This package is necessary because the `children` property of certain nodes can be deeply nested.

---

Ref: https://github.com/eslint/markdown/pull/469/files#r2215574938

The reason why I used `fast-deep-equal` package:

Simply comparing a `Heading` node’s text using the `toString` method would break the current behavior (which would be a breaking change), since the current rule does not report the following example:

```md
# Heading 1

# Heading *1*
```

package.json
```diff
@@ -90,10 +90,12 @@
...
     "github-slugger": "^2.0.0",
     "mdast-util-from-markdown": "^2.0.2",
     "mdast-util-frontmatter": "^2.0.1",
     "mdast-util-gfm": "^3.0.0",
+    "mdast-util-to-string": "^4.0.0",
```
lumirlumir: It won’t increase the number of dependencies for users, since this package is already a sub-dependency of the Mdast ecosystem.

Ref: https://npmgraph.js.org/?q=%40eslint%2Fmarkdown#select=exact%3Amdast-util-to-string%404.0.0

Image

src/rules/no-duplicate-headings.js
```diff
@@ -147,16 +122,20 @@ export default {
 					currentLevelHeadings = headingsByLevel.get(currentLevel);
 				}
 
-				if (currentLevelHeadings.has(headingText)) {
+				if (
```
lumirlumir: The most notable change in this PR is here.

Now, instead of comparing heading text directly through manual processing, we compare the `Heading` node's `children` AST directly for more robust and accurate equality.

---

Since handling multiline setext headings should also account for whitespaces, manually processing the text would introduce too many edge cases.

That's why I chose not to compare heading text, and instead opted to compare the ASTs directly.

Here are some test cases that demonstrate these edge cases:

Image

Image

Image

src/rules/no-duplicate-headings.js
```diff
@@ -3,41 +3,51 @@
...
- * ```
- *
- * @see https://spec.commonmark.org/0.31.2/#example-76
+ * Removes `position` and `data` properties from a node, recursively for all children.
+ * @param {Node} node The node to process.
+ * @returns {NodeWithoutPositionAndData} A new node without `position` and `data` properties.
  */
-const trailingAtxHeadingHashPattern = /(?<![ \t])[ \t]+#+[ \t]*$/u;
-const leadingAtxHeadingHashPattern = /^#{1,6}[ \t]+/u;
+function removePositionAndDataFromNode(node) {
```
lumirlumir: It would be best to remove the `position` and `data` properties, since `position` differs for each node and `data` isn’t needed when comparing the structure of the Markdown AST.

nzakas (commented):
Thanks for looking at this. The way this is currently implemented, we're actually traversing the AST inside of the `heading()` function, which is inefficient because we're already traversing the AST with ESLint. Additionally, cloning the AST to do a compare is additional overhead that I don't think we need.

I'd suggest revisiting the approach here to use traversal methods instead of doing another traversal.

For instance, you can do:

- `heading(node)` - mark when you're entering a heading
- `heading *(node)` - capture all of the nodes that are descendants of the heading
- `heading:exit(node)` - check the heading now that we know all the pieces

that way, you're working with the already-existing traversal and I don't think we need the new dependencies.

src/rules/no-duplicate-headings.js
```diff
@@ -146,8 +89,27 @@ export default {
 					lastLevel = currentLevel;
 					currentLevelHeadings = headingsByLevel.get(currentLevel);
 				}
+			},
+
+			"heading *"({ type, value }) {
+				if (value) {
+					headingChildrenSequence += JSON.stringify({
```
lumirlumir: I've used `JSON.stringify` to represent the sequence of a `Heading` node's children.

I wanted to find a simple and reliable way to show the sequence of heading children, and I thought `JSON.stringify` would be a safe option.

If there are any other suggestions or better approaches, I'd be happy to hear them.

lumirlumir (commented):
@nzakas Thank you!

I hadn't thought of that approach before, and I found it really impressive.

I've updated this PR to follow your suggestion.

src/rules/no-duplicate-headings.js
snitin315: JSON.stringify can be slow. I think we can we use a utility to create a canonical string:

```js
function getCanonicalHeadingText(node) {
	let text = "";
	for (const child of node.children) {
		if (child.type === "text" || child.type === "inlineCode") {
			text += child.value;
		}
		// You can add more node types here if needed
	}
	return text.toLowerCase().trim();
}
```

WDYT?

lumirlumir: The custom stringify approach sounds good to me 👍 I'll follow your direction with a few adjustments.

lumirlumir: I've added a new commit https://github.com/eslint/markdown/pull/469/commits/5aad726b3c2736d1662380fb3593480da8fefa1e.

I replaced `JSON.stringify` with a custom string representation for the heading children sequence.

I wrapped the string in `[]` to treat it as a unit when adding child nodes to the sequence. Without the brackets, the `value` could interfere with duplicate heading checks and result in false positives or negatives.

PR #487, opened 2025-07-24
fix: fenced-code-language highlights only relevant parts

### LINKED ISSUE #476: Rule Change: Highlight fences only in fenced-code-language rule reports

fenced-code-language

Highlight fences only in fenced-code-language rule report

Other

### Example code

\

\```js
const message = "Hello, world!";
console.log(message);
\```

Currently, when a violation occurs (e.g., missing language or disallowed language), the rule reports the entire code block.

After this change, the rule will only highlight the relevant parts of the code block for a violation.
- For a **missing language** error, it will highlight only the **opening fences**.
- For a **disallowed language** error, it will highlight the **opening fence and the specified language**.

### Participation

### Additional comments

This change will make the reported locations more precise and similar to how markdownlint highlight issues, making it easier for developers to pinpoint the exact source of the problem without distracting from the entire code content.

TKDev7:
#### Prerequisites checklist

This pull request aims to enhance the clarity and precision of error reporting for the fenced-code-language rule. Previously, rule violations would highlight the entire code block, making it difficult to pinpoint the exact issue. This change ensures that only the relevant parts of the code are highlighted, improving the developer experience.

I've modified the fenced-code-language rule to provide more targeted highlighting for violations:

- For missing language errors, only the opening fences (e.g., ```) of the code block are now highlighted.
- For disallowed language errors, it will highlight the opening fence and the specified language.

#### Related Issues

Fixes #476

### DIFF
--- a/src/rules/fenced-code-language.js
+++ b/src/rules/fenced-code-language.js
@@ -81,7 +81,16 @@ export default {
 					}
 
 					context.report({
-						loc: node.position,
+						loc: {
+							start: node.position.start,
+							end: {
+								line: node.position.start.line,
+								column:
+									sourceCode.lines[
+										node.position.start.line - 1
+									].length + 1,
+							},
+						},
 						messageId: "missingLanguage",
 					});
 
@@ -90,7 +99,16 @@ export default {
 
 				if (required.size && !required.has(node.lang)) {
 					context.report({
-						loc: node.position,
+						loc: {
+							start: node.position.start,
+							end: {
+								line: node.position.start.line,
+								column:
+									sourceCode.lines[
+										node.position.start.line - 1
+									].length + 1,
+							},
+						},
 						messageId: "disallowedLanguage",
 						data: {
 							lang: node.lang,
--- a/src/rules/fenced-code-language.js
+++ b/src/rules/fenced-code-language.js
@@ -98,15 +98,19 @@ export default {
 				}
 
 				if (required.size && !required.has(node.lang)) {
+					const lineText =
+						sourceCode.lines[node.position.start.line - 1];
+					const langIndex = lineText.indexOf(node.lang);
+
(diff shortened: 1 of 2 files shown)


src/rules/fenced-code-language.js
```diff
@@ -90,7 +99,16 @@ export default {
...
 					context.report({
-						loc: node.position,
+						loc: {
+							start: node.position.start,
+							end: {
+								line: node.position.start.line,
+								column:
+									sourceCode.lines[
+										node.position.start.line - 1
+									].length + 1,
```
JoshuaKGoldberg: [Bug] Any meta text attached to the code block will also be captured by this. But meta text isn't what's relevant.

````md
```js foo
````

````md
```js filename="src/index.js"
````

Proposal: how about limiting the range to just the <code>\`\`\`</code>, as well as the language if disallowed?

tests/rules/fenced-code-language.test.js
JoshuaKGoldberg, line 1: [Testing] Regardless of whether my other comment proposal is taken on, I think testing for the meta case should be added. The rule's decision to report or not report on meta text is going to now be intentional behavior.

JoshuaKGoldberg (changes requested):
A solid start! 🙌

src/rules/fenced-code-language.js
TKDev7: I went back and forth on this and decided to follow Markdownlint by highlighting the entire info string.

JoshuaKGoldberg: Other than keeping to Markdownlint's precedent, was there a motivation for this way?

TKDev7: That was the main reason. However, I decided to go with your approach to avoid capturing meta text.

PR #455, opened 2025-07-02
feat: add allow options to no-empty-definitions

TKDev7:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What is the purpose of this pull request?

This pull request adds support for the `allowDefinitions` option to the no-empty-definitions rule, allowing users to specify definition identifiers that should be intentionally allowed to be empty.  It also adds an `allowFootnoteDefinitions` option to let users specify footnote identifiers that should be allowed to be empty.

#### What changes did you make? (Give an overview)

- Modified the rule logic to skip reporting empty definitions and footnote definitions whose identifiers are included in the respective allow lists.
- Added tests to verify the new options work as intended.
- Improved documentation to describe the new options and provide usage examples.

#### Related Issues

Fixes #436

PR #488, opened 2025-07-27
fix: normalize definition identifiers in no-*-definitions rules

### LINKED ISSUE #474: Rule Change: Normalize definition identifiers in no-*-definition

no-*-definition

Generate fewer warnings

A new default behavior

### Example code

```markdown
<!-- eslint markdown/no-unused-definitions: ["error", { allowDefinitions: ["MERCURY"] }] -->

[MERCURY]: https://example.com/mercury/
```

It reports a false positive, flagging the definition as unused, because the identifier in the allow list ("MERCURY") does not match the normalized identifier used by mdast.

It will correctly recognize that "MERCURY" (regardless of case or whitespace differences) matches the normalized identifier, and will not report the definition as unused.

### Participation

### Additional comments

We can use [mdast-util-normalize-identifier](https://github.com/micromark/micromark/tree/main/packages/micromark-util-normalize-identifier), which is used by Mdast internally. This normalization issue affects no-empty-definitions, no-duplicate-definitions, and no-unused-definitions.

TKDev7:
#### Prerequisites checklist

This PR fixes a false positive issue in the `no-unused-definitions`, `no-duplicate-definitions`, and `no-empty-definitions` rules where the `allowDefinitions` and `allowFootnoteDefinitions` options were not properly handling case-insensitive and whitespace-normalized identifier matching. The rules were doing exact string comparisons instead of using the same normalization algorithm that mdast uses internally, causing definitions to be incorrectly flagged as violations when they should have been allowed.

- Updated allow list processing to normalize identifiers in the `allowDefinitions` and `allowFootnoteDefinitions` arrays using `normalizeIdentifier` function from `micromark-util-normalize-identifier`
- Added test cases to verify that case-insensitive and whitespace-normalized matching works correctly

#### Related Issues

Fixes #474


src/rules/no-empty-definitions.js
```diff
@@ -88,9 +94,15 @@ export default {
...
 	create(context) {
-		const allowDefinitions = new Set(context.options[0].allowDefinitions);
+		const allowDefinitions = new Set(
+			context.options[0].allowDefinitions.map(id =>
+				normalizeIdentifier(id).toLowerCase(),
+			),
+		);
 		const allowFootnoteDefinitions = new Set(
-			context.options[0].allowFootnoteDefinitions,
+			context.options[0].allowFootnoteDefinitions.map(id =>
```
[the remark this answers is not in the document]
lumirlumir: Let’s keep the optional chaining in the code.

The options will be injected, so removing it wouldn’t be a big issue, but we’ve usually used optional chaining for options throughout the rules when they’re consumed by a set.

src/rules/no-duplicate-definitions.js
```diff
@@ -67,9 +73,15 @@ export default {
 	},
 
 	create(context) {
-		const allowDefinitions = new Set(context.options[0]?.allowDefinitions);
+		const allowDefinitions = new Set(
+			context.options[0].allowDefinitions.map(id =>
```
lumirlumir:
```suggestion
			context.options[0].allowDefinitions.map(identifier =>
```

I think using `identifier` here would be more consistent with the rest of the codebase.

(The same applies to the other code as well.)

tests/rules/no-duplicate-definitions.test.js
```diff
@@ -76,6 +76,127 @@ ruleTester.run("no-duplicate-definitions", rule, {
...
+`,
+			options: [
+				{
+					allowFootnoteDefinitions: ["mercury"],
+				},
+			],
+		},
+		{
+			code: `
+[^   mercury   ]: Hello, Mercury!
```
lumirlumir: Trailing or leading spaces are not recognized as valid footnote definitions:

[AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogaGlcXG4gKi9cXG5oaVwiLFwianNvblwiOlwiKGEgPT09IDEgPyB0cnVlIDogZmFsc2UpXFxuICA%2FIHRydWUgXFxuICA6IGZhbHNlXCIsXCJtYXJrZG93blwiOlwiW14gICBtZXJjdXJ5ICAgXTogSGVsbG8sIE1lcmN1cnkhXCIsXCJjc3NcIjpcIlwiLFwiaHRtbFwiOlwiPCEtLVwifSxcImxhbmd1YWdlXCI6XCJtYXJrZG93blwiLFwianNPcHRpb25zXCI6e1wicGFyc2VyXCI6XCJlc3ByZWVcIixcInNvdXJjZVR5cGVcIjpcIm1vZHVsZVwiLFwiZXNWZXJzaW9uXCI6XCJsYXRlc3RcIixcImlzSlNYXCI6dHJ1ZX0sXCJqc29uT3B0aW9uc1wiOntcImpzb25Nb2RlXCI6XCJqc29uNVwiLFwiYWxsb3dUcmFpbGluZ0NvbW1hc1wiOmZhbHNlfSxcImNzc09wdGlvbnNcIjp7XCJjc3NNb2RlXCI6XCJjc3NcIixcInRvbGVyYW50XCI6ZmFsc2V9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJnZm1cIixcIm1hcmtkb3duRnJvbnRtYXR0ZXJcIjpcIm9mZlwifSxcImh0bWxPcHRpb25zXCI6e1widGVtcGxhdGVFbmdpbmVTeW50YXhcIjpcIm5vbmVcIixcImZyb250bWF0dGVyXCI6ZmFsc2V9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D)

Image

We can either remove this test case or replace it with another valid one.

tests/rules/no-duplicate-definitions.test.js
```diff
@@ -76,6 +76,127 @@ ruleTester.run("no-duplicate-definitions", rule, {
...
+			options: [
+				{
+					allowFootnoteDefinitions: ["mercury"],
+				},
+			],
+		},
+		{
+			code: `
+[^mercury]: Hello, Mercury!
+[^   mercury   ]: Hello, Venus!
```
lumirlumir: Same with https://github.com/eslint/markdown/pull/488/files#r2242147164.

tests/rules/no-duplicate-definitions.test.js
```diff
@@ -76,6 +76,127 @@ ruleTester.run("no-duplicate-definitions", rule, {
...
+				{
+					allowFootnoteDefinitions: ["   mercury   "],
+				},
+			],
+		},
+		// This test case is skipped for non-Node environments like Bun
+		...(typeof process !== "undefined" &&
+		process.release &&
+		process.release.name === "node" &&
+		(!process.versions || !process.versions.bun)
```
lumirlumir: Can we simplify this code a bit with optional chaining?

tests/rules/no-unused-definitions.test.js
```diff
@@ -119,6 +119,118 @@ Mercury[^mercury]
...
+				{
+					allowFootnoteDefinitions: ["   mercury   "],
+				},
+			],
+		},
+		// This test case is skipped for non-Node environments like Bun
+		...(typeof process !== "undefined" &&
+		process.release &&
+		process.release.name === "node" &&
+		(!process.versions || !process.versions.bun)
```
lumirlumir: Same with https://github.com/eslint/markdown/pull/488/files#r2242154517

tests/rules/no-empty-definitions.test.js
```diff
@@ -67,13 +67,96 @@ ruleTester.run("no-empty-definitions", rule, {
...
+		},
+		{
+			code: "[^note]:",
+			options: [{ allowFootnoteDefinitions: ["   note   "] }],
 		},
+		// This test case is skipped for non-Node environments like Bun
+		...(typeof process !== "undefined" &&
+		process.release &&
+		process.release.name === "node" &&
+		(!process.versions || !process.versions.bun)
```
lumirlumir: Same with https://github.com/eslint/markdown/pull/488/files#r2242154517

tests/rules/no-empty-definitions.test.js
```diff
@@ -67,13 +67,96 @@ ruleTester.run("no-empty-definitions", rule, {
 		},
 		{
 			code: "[^note]:",
-			options: [
-				{
-					checkFootnoteDefinitions: true,
```
lumirlumir: Let's keep the `checkFootnoteDefinitions: true`.

tests/rules/no-empty-definitions.test.js
```diff
@@ -67,13 +67,96 @@ ruleTester.run("no-empty-definitions", rule, {
...
+		{
+			code: "[^note]:",
+			options: [{ allowFootnoteDefinitions: ["NOTE"] }],
+		},
+		{
+			code: "[^NOTE]:",
+			options: [{ allowFootnoteDefinitions: ["note"] }],
+		},
+		{
+			code: "[^   note   ]:",
```
lumirlumir: Same with https://github.com/eslint/markdown/pull/488/files#r2242147164

tests/rules/no-unused-definitions.test.js
```diff
@@ -119,6 +119,118 @@ Mercury[^mercury]
...
+		{
+			code: "[^MERCURY]: Hello, Mercury!",
+			options: [
+				{
+					allowFootnoteDefinitions: ["mercury"],
+				},
+			],
+		},
+		{
+			code: "[^   mercury   ]: Hello, Mercury!",
```
lumirlumir: Same with https://github.com/eslint/markdown/pull/488/files#r2242147164

lumirlumir (changes requested):

src/rules/no-empty-definitions.js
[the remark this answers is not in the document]
TKDev7: I've removed them following nazaka's suggestion from the other PR (https://github.com/eslint/markdown/pull/455#discussion_r2180239449). Let me know if you'd still like to keep the optional chaining.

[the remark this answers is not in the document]
nzakas: There's no need for optional chaining. We don't need to agree with Copilot's comments. It's just there as an extra set of eyes.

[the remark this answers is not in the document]
lumirlumir: Alright, let's go ahead and remove optional chaining from this section.

lumirlumir (approved):
LGTM, thanks.

lumirlumir (approved):
Re-approving.

package.json
```diff
@@ -97,7 +97,8 @@
     "mdast-util-frontmatter": "^2.0.1",
     "mdast-util-gfm": "^3.1.0",
     "micromark-extension-frontmatter": "^2.0.0",
-    "micromark-extension-gfm": "^3.0.0"
+    "micromark-extension-gfm": "^3.0.0",
+    "micromark-util-normalize-identifier": "^2.0.1"
```
snitin315, line 101: Adding this package will also [install `micromark-util-symbol` ](https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-normalize-identifier/package.json#L43) which increases the total package size to 53kb.

Screenshot 2025-08-09 at 1 09 31 PM

snitin315: And if we see the implementation, it is just using [one value (`values.space`)](https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-normalize-identifier/dev/index.js#L26) which we can [simply hardcode in our codebase](https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-symbol/lib/values.js#L13). I don't think it's worth adding the addition 53kb to the installation size.

I think we can just move [the small utility](https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-normalize-identifier/dev/index.js#L22) to our codebase along with the MIT license.

snitin315 (changes requested):

package.json
TKDev7: Sounds good. Should I simply copy the function into our util.js file and add the MIT license at the top, or would you prefer I create a separate file for it to keep the license isolated?

snitin315: You can create a separate file and add MIT license at the top, don't combine it with utils.js. But you can import & re-export from utils.js.

lumirlumir: Personally, I think it's fine to keep `micromark-util-normalize-indentifier` as an external package.

Since `micromark-util-normalize-indentifier` is closely integrated with [`mdast-util-from-markdown`](https://github.com/eslint/markdown/blob/main/package.json#L96), which is our direct dependency and main parser, adding it as a dependency shouldn't increase the bundle size.

https://github.com/eslint/markdown/blob/1b30119e6627b92ebba55ef64e10c3de644f24f5/package.json#L96

https://npmgraph.js.org/?q=mdast-util-from-markdown#select=exact%3Amicromark-util-normalize-identifier%402.0.1

image

Modern package managers can dedupe dependencies, and bundlers can automatically remove unused code paths if needed, so I think adding this dependency would not affect the size.

I would love to hear your thoughts on this, @snitin315 😄

snitin315: I feel if we can remove the depency we should

lumirlumir: I'm not sure this approach would actually offer any advantages. Do you have any examples where it could be beneficial?

If `micromark-util-normalize-indentifier` and `micromark-util-symbol` weren't already included as sub-dependencies, we could save package size as you suggested. However, since they are brought in by `mdast-util-from-markdown`, they're already part of our dependency tree.

I looked into it further, but I couldn't find a way to reduce the package size by internalizing them. In fact, doing so would only increase the size of `@eslint/markdown` by 53KB due to duplicating the same logic.

Package managers like npm, yarn, and pnpm will deduplicate these dependencies regardless of how many times they're installed, and even bundlers will avoid duplicating the logic if it's coming from the same package.

Additionally, internalizing these modules would prevent us from benefiting from patches or new features, and it would increase our overall maintenance costs.

snitin315: > I looked into it further, but I couldn't find a way to reduce the package size by internalizing them. In fact, doing so would only increase the size of @eslint/markdown by 53KB due to duplicating the same logic.

We don't have to internalise the complete 2 packages. And I'm not suggesting that we should, we just need 4 lines of code that we can use.

snitin315: But yeah, I actually misread the previous comment. If this is deduped we can keep the logic as is.

PR #468, opened 2025-07-16
fix: detect errors after comments in no-multiple-h1 and require-alt-text

TKDev7:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### What is the purpose of this pull request?

This pull request fixes a bug where the `no-multiple-h1` and `require-alt-text` rules failed to report errors for headings and images that appear after a closed HTML comment in Markdown files. 

#### What changes did you make? (Give an overview)

- Added a utility function to strip HTML comments from Markdown content before running rule checks.
- Updated the `no-multiple-h1` rule to ensure that `<h1>` tags after HTML comments are correctly detected and reported.
- Updated the `require-alt-text` rule to ensure that `` tags after HTML comments are correctly detected and reported.
- Added tests to confirm that both rules now report errors for content following closed HTML comments.

#### Related Issues

Fixes #464

PR #491, opened 2025-07-31
fix: improve no-html rule's tag location reporting

### LINKED ISSUE #478: Rule Change: Improve no-html rule's location reporting for HTML tags

no-html

Other

Other

### Example code

```markdown
<div id="foo">Some content</div>
```

Currently, the rule calculates the end of the reported location for an HTML tag using `start.column + match[0].length + 1`. This typically reports only the tag name itself and an extra character, missing any attributes or trailing spaces within the opening tag. For `<div id="foo">`, it reports `<div `

After this change, the rule will report the entire opening HTML tag, including any attributes and spaces, as long as it's on a single line. If the opening tag spans multiple lines, only the first line of the tag will be reported. This aligns with how markdownlint reports similar issues.

- For `<div id="foo">`, it will report the full `<div id="foo">`.
- For `<p\n  data-attribute="value">`, it will report only `<p`.

### Participation

### Additional comments

_No response_

TKDev7:
#### Prerequisites checklist

This pull request aims to enhance the clarity and precision of error reporting for the `no-html` rule. Previously, rule violations for HTML tags would highlight only a partial segment, specifically the tag name and an extra character.

I've modified the `no-html` rule's reporting logic to report the entire opening HTML tag, including any attributes and spaces, as long as it's on a single line. If the opening tag spans multiple lines, only the first line of the tag will be reported. 

- For `<div id="foo">`, it will report the full `<div id="foo">`.
- For `<p\n  data-attribute="value">`, it will report only `<p`.

#### Related Issues

Fixes #478


nzakas (approved):
LGTM. Thanks! Please take a look at the Copilot comments.

src/rules/no-html.js
lumirlumir, line 1: As we discussed in https://github.com/eslint/markdown/pull/488#discussion_r2248121620, can we remove optional chaining here on line 71?

```diff
- const allowed = new Set(context.options[0]?.allowed);
+ const allowed = new Set(context.options[0].allowed);
```

src/rules/no-html.js
```diff
@@ -24,7 +24,7 @@ import { findOffsets } from "../util.js";
 // Helpers
 //-----------------------------------------------------------------------------
 
-const htmlTagPattern = /<([a-z0-9]+(?:-[a-z0-9]+)*)/giu;
+const htmlTagPattern = /<([a-z0-9]+(?:-[a-z0-9]+)*)[^>]*>/giu;
```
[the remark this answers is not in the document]
lumirlumir: I think the issue here is not what Copilot mentioned.

The behavior is correct:

Image

---

However, the problem is that the position is reported incorrectly when `>` is escaped or used inside an attribute value:

Image

Image

src/rules/no-html.js
```diff
@@ -84,9 +85,16 @@ export default {
 						line: node.position.start.line + lineOffset,
 						column: node.position.start.column + columnOffset,
 					};
+
+					const firstLineEnd = fullMatch.indexOf("\n");
+					const endColumn =
+						firstLineEnd === -1
+							? start.column + fullMatch.length
+							: start.column + firstLineEnd;
```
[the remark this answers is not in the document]
lumirlumir: I think `firstNewlineIndex` makes sense here.

src/rules/no-html.js
```diff
@@ -84,9 +85,16 @@ export default {
 						line: node.position.start.line + lineOffset,
 						column: node.position.start.column + columnOffset,
 					};
+
+					const firstLineEnd = fullMatch.indexOf("\n");
```
lumirlumir: According to the HTML and Markdown specifications, LF, CR, and CRLF can all be used for line endings.

https://infra.spec.whatwg.org/#example-code-point-notation-hard-to-render

Image

https://spec.commonmark.org/0.31.2/#line-ending

Image

---

So, the following code reports incorrect position information:

Image

lumirlumir (changes requested):

src/rules/no-html.js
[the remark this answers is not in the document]
TKDev7: This first edge case is difficult to solve using our current regex approach. It's a good candidate for the ultrahtml refactor, and I've opened #496 to track it.

The second case looks correct ([AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG57XFxuICAgIFxcXCJrZXkxXFxcIjogW3RydWUsIGZhbHNlLCBudWxsXSxcXG4gICAgXFxcImtleTJcXFwiOiB7XFxuICAgICAgICBcXFwia2V5M1xcXCI6IFsxLCAyLCBcXFwiM1xcXCIsIDFlMTAsIDFlLTNdXFxuICAgIH1cXG59XCIsXCJtYXJrZG93blwiOlwiPGlucHV0IHBsYWNlaG9sZGVyPVxcXFw%2BIGNsYXNzPWFiYz5cIixcImNzc1wiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIENTUyBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG4gKiB0aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcbiAqXFxuICogVGhlIHRhYnMgYXJlOlxcbiAqXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICpcXG4gKiBZb3UgY2FuIGNoYW5nZSB0aGUgd2F5IHRoYXQgdGhlIENTUyBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkNTU1xcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuQGltcG9ydCB1cmwoJ2h0dHBzOi8vZm9udHMuZ29vZ2xlYXBpcy5jb20vY3NzMj9mYW1pbHk9Um9ib3RvOndnaHRANDAwOzcwMCZkaXNwbGF5PXN3YXAnKTtcXG5cXG5ib2R5IHtcXG5cXHRmb250LWZhbWlseTogc2Fucy1zZXJpZjtcXG59XFxuXFxuaDEge1xcblxcdGNvbG9yOiAjMzMzO1xcbn1cXG5cXG5wIHtcXG5cXHRtYXJnaW46IDFlbSAwO1xcbn1cIixcImh0bWxcIjpcIjwhRE9DVFlQRSBodG1sPlxcbjwhLS1cXG5UeXBlIG9yIHBhc3RlIHNvbWUgSFRNTCBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG50aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcblxcblRoZSB0YWJzIGFyZTpcXG5cXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0Llxcbi0tPlxcblxcbjxodG1sIGxhbmc9XFxcImVuXFxcIj5cXG4gICAgPGhlYWQ%2BXFxuICAgICAgICA8bWV0YSBjaGFyc2V0PVxcXCJVVEYtOFxcXCI%2BXFxuICAgICAgICA8dGl0bGU%2BSFRNTDwvdGl0bGU%2BXFxuICAgIDwvaGVhZD5cXG4gICAgPGJvZHk%2BXFxuICAgICAgICA8cD5UZXh0PC9wPlxcbiAgICA8L2JvZHk%2BXFxuPC9odG1sPlwifSxcImxhbmd1YWdlXCI6XCJtYXJrZG93blwiLFwianNPcHRpb25zXCI6e1wicGFyc2VyXCI6XCJlc3ByZWVcIixcInNvdXJjZVR5cGVcIjpcIm1vZHVsZVwiLFwiZXNWZXJzaW9uXCI6XCJsYXRlc3RcIixcImlzSlNYXCI6dHJ1ZX0sXCJqc29uT3B0aW9uc1wiOntcImpzb25Nb2RlXCI6XCJqc29uY1wiLFwiYWxsb3dUcmFpbGluZ0NvbW1hc1wiOmZhbHNlfSxcImNzc09wdGlvbnNcIjp7XCJjc3NNb2RlXCI6XCJjc3NcIixcInRvbGVyYW50XCI6ZmFsc2V9LFwibWFya2Rvd25PcHRpb25zXCI6e1wibWFya2Rvd25Nb2RlXCI6XCJjb21tb25tYXJrXCIsXCJtYXJrZG93bkZyb250bWF0dGVyXCI6XCJvZmZcIn0sXCJodG1sT3B0aW9uc1wiOntcInRlbXBsYXRlRW5naW5lU3ludGF4XCI6XCJub25lXCIsXCJmcm9udG1hdHRlclwiOmZhbHNlfSxcIndyYXBcIjp0cnVlLFwidmlld01vZGVzXCI6e1wiYXN0Vmlld1wiOlwidHJlZVwiLFwic2NvcGVWaWV3XCI6XCJmbGF0XCIsXCJwYXRoVmlld1wiOlwiZ3JhcGhcIn0sXCJwYXRoSW5kZXhcIjp7XCJpbmRleFwiOjAsXCJpbmRleGVzXCI6MX0sXCJlc3F1ZXJ5U2VsZWN0b3JcIjp7XCJzZWxlY3RvclwiOlwiXCJ9fSxcInZlcnNpb25cIjowfSI%3D))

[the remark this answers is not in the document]
lumirlumir: As mentioned in https://github.com/eslint/markdown/pull/468#issuecomment-3103042304, I don’t think it would be ideal to wait for the HTML handling logic.

However, I’m okay to keep this PR on hold if needed. I’d appreciate hearing the team’s thoughts on this. @eslint/eslint-team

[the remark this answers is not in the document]
TKDev7: @lumirlumir Since #489 might take time, I've pushed a regex fix for > in attributes. PTAL

tests/rules/no-html.test.js
```diff
@@ -130,5 +130,187 @@ ruleTester.run("no-html", rule, {
...
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 7,
+					data: {
+						name: "input",
+					},
+				},
+			],
+		},
```
snitin315, line 406: Can you add a test case which actually goes to the next line, for example:

```js
code: `
     <input 
         placeholder="enter name"
         name="First-name"
      />
`
```

To assert that this should not have endLine as `1`.

snitin315 (approved):
LGTM, thanks!

src/rules/no-html.js
```diff
@@ -24,7 +24,9 @@ import { findOffsets } from "../util.js";
 // Helpers
 //-----------------------------------------------------------------------------
 
-const htmlTagPattern = /<([a-z0-9]+(?:-[a-z0-9]+)*)/giu;
+const htmlTagPattern =
```
lumirlumir, line 27: According to the spec, we’re seeing some false positives.

I’m not looking for 100% spec-compliant behavior with this regex, but the sequence below can be used and I think it should be supported.

image

---

```md
<input placeholder=I'm>

<input placeholder=double"quote>
```

image

src/rules/no-html.js
```diff
@@ -24,7 +24,9 @@ import { findOffsets } from "../util.js";
...
+	/<([a-z0-9]+(?:-[a-z0-9]+)*)(?:\s+(?:[^>"']|"[^"]*"|'[^']*')*)?>/giu;
+const nextLinesPattern = /[\r\n][\s\S]*$/u;
```
lumirlumir:
```suggestion
const lineEndingPattern = /\n|\r|\r\n/u;
```

I think it's clearer to use the regex above and the variable name, since the ESLint plugin ecosystem typically uses `lineEndingPattern` for this.

Ref: https://github.com/eslint/rewrite/blob/main/packages/plugin-kit/src/source-code.js#L248

src/rules/no-html.js
```diff
@@ -84,9 +87,17 @@ export default {
 						line: node.position.start.line + lineOffset,
 						column: node.position.start.column + columnOffset,
 					};
+
+					const firstNewlineIndex =
+						fullMatch.search(nextLinesPattern);
+					const endColumn =
+						firstNewlineIndex === -1
+							? start.column + fullMatch.length
+							: start.column + firstNewlineIndex;
```
lumirlumir:
```suggestion
					const firstNewlineIndex =
						fullMatch.match(lineEndingPattern)?.index;
					const endColumn =
						firstNewlineIndex === undefined
							? start.column + fullMatch.length
							: start.column + firstNewlineIndex;
```

Also, I think we can update it like this to make our intentions clearer with a more explicit regex pattern.

src/rules/no-html.js
TKDev7: These are not valid HTML nodes. ([AST](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLyoqXFxuICogVHlwZSBvciBwYXN0ZSBzb21lIEphdmFTY3JpcHQgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cXG4gKiBcXG4gKiBUaGUgdGhyZWUgdGFicyBhcmU6XFxuICogXFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxcbiAqICAgYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuICogICBjYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxcbiAqICAgdXNlZnVsIHRvIHVuZGVyc3RhbmQgaG93IHZhcmlhYmxlcyBhcmUgZGVmaW5lZCBhbmRcXG4gKiAgIHdoZXJlIHRoZXkgYXJlIHVzZWQuXFxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcXG4gKiAgIGNhbiBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdGhlIGNvZGUgaXMgZXhlY3V0ZWQuXFxuICogXFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiSmF2YVNjcmlwdFxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbiAqIG9wdGlvbnMuXFxuICovXFxuXFxuaW1wb3J0IGpzIGZyb20gXFxcIkBlc2xpbnQvanNcXFwiO1xcblxcbmZ1bmN0aW9uIGdldENvbmZpZygpIHtcXG4gICAgcmV0dXJuIHtcXG4gICAgICAgIHJ1bGVzOiB7XFxuICAgICAgICAgICAgXFxcInByZWZlci1jb25zdFxcXCI6IFxcXCJ3YXJuXFxcIlxcbiAgICAgICAgfVxcbiAgICB9O1xcbn1cXG5cXG5leHBvcnQgZGVmYXVsdCBbXFxuICAgIC4uLmpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXFxuICAgIGdldENvbmZpZygpXFxuXTtcIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG57XFxuICAgIFxcXCJrZXkxXFxcIjogW3RydWUsIGZhbHNlLCBudWxsXSxcXG4gICAgXFxcImtleTJcXFwiOiB7XFxuICAgICAgICBcXFwia2V5M1xcXCI6IFsxLCAyLCBcXFwiM1xcXCIsIDFlMTAsIDFlLTNdXFxuICAgIH1cXG59XCIsXCJtYXJrZG93blwiOlwiPGlucHV0IHBsYWNlaG9sZGVyPUknbT5cXG5cXG48aW5wdXQgcGxhY2Vob2xkZXI9ZG91YmxlXFxcInF1b3RlPlwiLFwiY3NzXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgQ1NTIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgQ1NTIGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiQ1NTXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKi9cXG5cXG5AaW1wb3J0IHVybCgnaHR0cHM6Ly9mb250cy5nb29nbGVhcGlzLmNvbS9jc3MyP2ZhbWlseT1Sb2JvdG86d2dodEA0MDA7NzAwJmRpc3BsYXk9c3dhcCcpO1xcblxcbmJvZHkge1xcblxcdGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xcbn1cXG5cXG5oMSB7XFxuXFx0Y29sb3I6ICMzMzM7XFxufVxcblxcbnAge1xcblxcdG1hcmdpbjogMWVtIDA7XFxufVwiLFwiaHRtbFwiOlwiPCFET0NUWVBFIGh0bWw%2BXFxuPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbnRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuXFxuVGhlIHRhYnMgYXJlOlxcblxcbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG5jYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuLS0%2BXFxuXFxuPGh0bWwgbGFuZz1cXFwiZW5cXFwiPlxcbiAgICA8aGVhZD5cXG4gICAgICAgIDxtZXRhIGNoYXJzZXQ9XFxcIlVURi04XFxcIj5cXG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cXG4gICAgPC9oZWFkPlxcbiAgICA8Ym9keT5cXG4gICAgICAgIDxwPlRleHQ8L3A%2BXFxuICAgIDwvYm9keT5cXG48L2h0bWw%2BXCJ9LFwibGFuZ3VhZ2VcIjpcIm1hcmtkb3duXCIsXCJqc09wdGlvbnNcIjp7XCJwYXJzZXJcIjpcImVzcHJlZVwiLFwic291cmNlVHlwZVwiOlwibW9kdWxlXCIsXCJlc1ZlcnNpb25cIjpcImxhdGVzdFwiLFwiaXNKU1hcIjp0cnVlfSxcImpzb25PcHRpb25zXCI6e1wianNvbk1vZGVcIjpcImpzb25jXCIsXCJhbGxvd1RyYWlsaW5nQ29tbWFzXCI6ZmFsc2V9LFwiY3NzT3B0aW9uc1wiOntcImNzc01vZGVcIjpcImNzc1wiLFwidG9sZXJhbnRcIjpmYWxzZX0sXCJtYXJrZG93bk9wdGlvbnNcIjp7XCJtYXJrZG93bk1vZGVcIjpcImNvbW1vbm1hcmtcIixcIm1hcmtkb3duRnJvbnRtYXR0ZXJcIjpcIm9mZlwifSxcImh0bWxPcHRpb25zXCI6e1widGVtcGxhdGVFbmdpbmVTeW50YXhcIjpcIm5vbmVcIixcImZyb250bWF0dGVyXCI6ZmFsc2V9LFwid3JhcFwiOnRydWUsXCJ2aWV3TW9kZXNcIjp7XCJhc3RWaWV3XCI6XCJ0cmVlXCIsXCJzY29wZVZpZXdcIjpcImZsYXRcIixcInBhdGhWaWV3XCI6XCJncmFwaFwifSxcInBhdGhJbmRleFwiOntcImluZGV4XCI6MCxcImluZGV4ZXNcIjoxfSxcImVzcXVlcnlTZWxlY3RvclwiOntcInNlbGVjdG9yXCI6XCJcIn19LFwidmVyc2lvblwiOjB9Ig%3D%3D))

src/rules/no-html.js
TKDev7: I can rename the variable, but could you double-check your regex? I based mine on the one from [Markdownlint](https://github.com/DavidAnson/markdownlint/blob/224987d727ebada36752074d2466a061c9e03ff0/helpers/shared.cjs#L16).

src/rules/no-html.js
TKDev7: I can implement your suggestion, but could you help me understand why it’s better? I think `search()` with a `-1` check is a bit more straightforward and avoids the extra object from `match()`.

lumirlumir: That makes sense to me. I think I can go with your suggestion.

src/rules/no-html.js
lumirlumir: This is because using a consistent line ending regex pattern across the markdown repository will make future refactoring easier.

I'm preparing for https://github.com/eslint/markdown/pull/493, and in this PR, I'd like to consolidate all the line ending regex patterns into a single one for a cleaner codebase.

I'm not saying your regex is incorrect. If you'd prefer to keep it as is, just renaming the variable name is fine.

nzakas:
@TKDev7 can you take a look at the conflicts?

PR #493, opened 2025-08-04
fix: add CR to the line ending pattern

lumirlumir:
## Prerequisites checklist

Working in progress.

According to the Markdown specification, LF, CR, and CRLF can all be used for line endings.

https://spec.commonmark.org/0.31.2/#line-ending

Image

## Related Issues

Ref: https://github.com/eslint/markdown/pull/491#discussion_r2251033615

### DIFF
--- a/src/language/markdown-source-code.js
+++ b/src/language/markdown-source-code.js
@@ -167,7 +167,7 @@ export class MarkdownSourceCode extends TextSourceCodeBase {
 	 * @param {Root} options.ast The root AST node.
 	 */
 	constructor({ text, ast }) {
-		super({ ast, text });
+		super({ ast, text, lineEndingPattern: /\n|\r|\r\n/u });
 		this.ast = ast;
 
 		// need to traverse the source code to get the inline config nodes


lumirlumir:
For a complete fix, #376 is a prerequisite, as some internal rules are closely tied to self-implemented line break logic.

package.json
lumirlumir, line 1: It looks like the newly released Bun 1.3 now enforces strict dependencies in `package.json`. If we use transitive dependencies like `@types/unist`, `@types/mdast`, or `semver`, it throws errors:

- Bun v1.3: https://github.com/oven-sh/bun/releases/tag/bun-v1.3.0

- https://github.com/eslint/markdown/actions/runs/18428303631/job/52512384911

image

- https://github.com/eslint/markdown/actions/runs/18428377164/job/52512542729

image



So, I've added `@types/unist`, `@types/mdast`, and `semver` to the dev dependencies.

src/rules/no-html.js
lumirlumir, line 1: This rule already supports CR line endings, so I only refactored the code.

src/rules/no-reference-like-urls.js
lumirlumir, line 1: Motivated by https://github.com/eslint/markdown/pull/525, I've simplified the regex a bit.

Since `imageBang` and `title` can be retrieved directly from the node's information, I've updated the code to access them directly from the nodes.

(This rule already supports CR line endings, so I only refactored the code.)

package.json
lumirlumir: I've merged the main branch, so this is now stale.

PR #500, opened 2025-08-06
feat: add allowedIgnoreCase option to no-html

### LINKED ISSUE #479: Rule Change: Make no-html rule's allowed option case-insensitive

no-html

Generate fewer warnings

A new default behavior

### Example code

```markdown
<div></div>
<DIV></DIV>
```

Currently, if `allowed: ["DIV"]` is configured, the rule will correctly allow `<DIV></DIV>`. However, `<div></div>` will be reported as an error because the matching is case-sensitive. This leads to false positives since HTML tags are, by specification, case-insensitive.

After this change, the no-html rule will perform case-insensitive matching when checking HTML tag names against the `allowed` option list. This means that if `allowed: ["DIV"]` is configured, both `<div></div>` and `<DIV></DIV>` (and any other casing like `<DiV>`) will be correctly recognized as allowed and will not be reported as errors.

### Participation

### Additional comments

_No response_

TKDev7:
#### Prerequisites checklist

This PR adds a new `allowedIgnoreCase` option to the `no-html` rule to enable case-insensitive matching for HTML tag names against the `allowed` array. This addresses the issue where HTML tag names are case-insensitive by specification, but the rule was performing case-sensitive matching, leading to false positives.
For example, with `allowed: ["DIV"]` configured:

- Before: Only `<DIV></DIV>` was allowed, but `<div></div>` was reported as an error
- After: With `allowedIgnoreCase: true`, both `<div></div>`, `<DIV></DIV>`, and `<DiV></DiV>` are allowed

- Implemented case-insensitive matching logic that converts both the allowed tags and found tag names to lowercase when `allowedIgnoreCase` is true
- Added test cases for case-insensitive scenarios and mixed case handling
- Added examples and explanation for the new option

#### Related Issues

Fixes #479

### DIFF
--- a/docs/rules/no-html.md
+++ b/docs/rules/no-html.md
@@ -26,49 +26,49 @@ The following options are available on this rule:
 
 * `allowed: Array<string>` - when specified, HTML tags are allowed only if they match one of the tags in this array. (default: `[]`)
 
-Examples of **incorrect** code when configured as `"no-html": ["error", { allowed: ["b"] }]`:
+    Examples of **incorrect** code when configured as `"no-html": ["error", { allowed: ["b"] }]`:
 
-```markdown
-<!-- eslint markdown/no-html: ["error", { allowed: ["b"] }] -->
+    ```markdown
+    <!-- eslint markdown/no-html: ["error", { allowed: ["b"] }] -->
 
-# Heading 1
+    # Heading 1
 
-Hello <em>world!</em>
-```
+    Hello <em>world!</em>
+    ```
 
-Examples of **correct** code when configured as `"no-html": ["error", { allowed: ["b"] }]`:
+    Examples of **correct** code when configured as `"no-html": ["error", { allowed: ["b"] }]`:
 
-```markdown
-<!-- eslint markdown/no-html: ["error", { allowed: ["b"] }] -->
+    ```markdown
+    <!-- eslint markdown/no-html: ["error", { allowed: ["b"] }] -->
 
-# Heading 1
+    # Heading 1
 
-Hello <b>world!</b>
-```
+    Hello <b>world!</b>
+    ```
 
-* `allowedIgnoreCase: boolean` - when `true`, enables case-insensitive matching for HTML tag names against the `allowed` array. (default: `false`)
+* `allowedIgnoreCase: boolean` - when `true`, enables case-insensitive matching of HTML tag names against the `allowed` array described above. (default: `false`)
 
-Examples of **incorrect** code when configured as `"no-html": ["error", { allowed: ["DIV"], allowedIgnoreCase: false }]`:
+    Examples of **incorrect** code when configured as `"no-html": ["error", { allowed: ["DIV"], allowedIgnoreCase: false }]`:
 
-```markdown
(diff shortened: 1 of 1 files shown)


snitin315 (commented):
@TKDev7 Before starting work, please wait for the issue to be accepted. The team is still finalizing the solution approach, and you can find more details in the [contributing guidelines.](https://eslint.org/docs/latest/contribute/work-on-issue#starting-work)

Screenshot 2025-08-09 at 12 27 04 PM

docs/rules/no-html.md
```diff
@@ -46,6 +46,30 @@ Examples of **correct** code when configured as `"no-html": ["error", { allowed:
...
+
+```markdown
+<!-- eslint markdown/no-html: ["error", { allowed: ["DIV"], allowedIgnoreCase: true }] -->
+
+# Heading 1
+
+<div>Hello world!</div>
+<DIV>Hello world!</DIV>
+<DiV>Hello world!</DiV>
+```
```
lumirlumir:
````suggestion
    Examples of **incorrect** code when configured as `"no-html": ["error", { allowed: ["DIV"], allowedIgnoreCase: false }]`:

    ```markdown
    <!-- eslint markdown/no-html: ["error", { allowed: ["DIV"], allowedIgnoreCase: false }] -->

    # Heading 1

    <div>Hello world!</div>
    ```

    Examples of **correct** code when configured as `"no-html": ["error", { allowed: ["DIV"], allowedIgnoreCase: true }]`:

    ```markdown
    <!-- eslint markdown/no-html: ["error", { allowed: ["DIV"], allowedIgnoreCase: true }] -->

    # Heading 1

    <div>Hello world!</div>
    <DIV>Hello world!</DIV>
    <DiV>Hello world!</DiV>
    ```
````

Could we indent the options like in [this example](https://github.com/eslint/markdown/blob/f5a64b917cb0224e7e56133887b6448716f857e1/docs/rules/no-unused-definitions.md#options), including the first one?

I think it's fine not to indent if there's only one option, but when there are two, the lack of indentation makes it harder to read.

File: docs/rules/no-html.md [hunk 2]
```diff
@@ -46,6 +46,30 @@ Examples of **correct** code when configured as `"no-html": ["error", { allowed:
 Hello <b>world!</b>
 ```
 
+* `allowedIgnoreCase: boolean` - when `true`, enables case-insensitive matching for HTML tag names against the `allowed` array. (default: `false`)
```
lumirlumir:
```suggestion
* `allowedIgnoreCase: boolean` - when `true`, enables case-insensitive matching for HTML tag names against the `allowed` option's array mentioned above. (default: `false`)
```

I think it would be helpful to add a description that the `allowed` is the one mentioned above.  

Feel free to adjust it further if you think it could be made clearer.

TKDev7: I feel like this is a bit wordy. I went with "when `true`, enables case-insensitive matching of HTML tag names against the `allowed` array described above. (default: `false`)". WDYT?

lumirlumir (approved):
LGTM, Would like another review before merging.

nzakas (approved):
LGTM. Thanks!

PR #508, opened 2025-08-11
refactor: correct no-reversed-media-syntax selectors

TKDev7:
#### Prerequisites checklist

#### Example Code

```markdown
# Heading with `inline` code

Paragraph with `(label)[url]` reversed syntax.
```

#### Current Behavior

The selector syntax: `"heading html,inlineCode"` is interpreted as:
- html inside heading
- OR any `inlineCode` node anywhere in the document

While in practice this behaves as intended since `inlineCode` can only appear inside headings, paragraphs, or table cells, the syntax can be misread and is less explicit about its scoping.

#### Expected Behavior

Selectors should clearly indicate that both `html` and `inlineCode` are matched only when nested within the relevant container node.

This PR fixes selector patterns to make their scope explicit and easier to read, replacing comma-separated selectors with `:matches()` expressions.

- Replaced comma-separated selectors such as `"heading html,inlineCode"` with :matches() expressions like `"heading :matches(html, inlineCode)"`

### DIFF
--- a/src/rules/no-reversed-media-syntax.js
+++ b/src/rules/no-reversed-media-syntax.js
@@ -143,7 +143,9 @@ export default {
 		}
 
 		return {
-			"heading html,inlineCode"(node) {
+			":matches(heading, paragraph, tableCell) :matches(html, inlineCode)"(
+				node,
+			) {
 				skipRanges.push(sourceCode.getRange(node));
 			},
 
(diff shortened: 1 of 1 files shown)


src/rules/no-reversed-media-syntax.js
```diff
@@ -143,7 +143,9 @@ export default {
 		}
 
 		return {
-			"heading html,inlineCode"(node) {
+			":matches(heading, paragraph, tableCell) :matches(html, inlineCode)"(
```
lumirlumir:
```suggestion
			"heading html,heading inlineCode"(
```

I think simply adding `heading` here without modifying the existing logic would make our intention clearer.

This applies to the `paragraph` and `tableCell` ESQuery selectors.

lumirlumir (changes requested):

src/rules/no-reversed-media-syntax.js
lumirlumir:
```suggestion
			"heading :matches(html, inlineCode)"(
```

Or, the code above would be fine too.

PR #510, opened 2025-08-21
build: add `prettier` to `update-rules-docs` script

lumirlumir:
#### Prerequisites checklist

Hi,

In this PR, I added Prettier to the `update-rules-docs` script.

This change was motivated by the implementations in `eslint/json` and `eslint/css`: 

- https://github.com/eslint/json/blob/main/tools/update-rules-docs.js#L103-L113 
- https://github.com/eslint/css/blob/main/tools/update-rules-docs.js#L103-L113

image

Other plugins like JSON and CSS use Prettier when running the `update-rules-docs script`, but the Markdown plugin didn't. 

It felt like a minor inconsistency across the plugin ecosystem, and since prettifying `README.md` made it clearer to me, I decided to make this change. 

I also removed `README.md` from `.prettierignore` so its formatting will be checked in CI.

The changes in `README.md` were auto-generated by Prettier, and I reviewed them, they only affect formatting.

In this PR, I added Prettier to the `update-rules-docs` script.

#### Related Issues

N/A

N/A


.prettierignore
```diff
@@ -2,3 +2,4 @@ dist
 CHANGELOG.md
 jsr.json
 **/*.md
+!README.md
```
lumirlumir, line 5: Actually, looking at the JSON and CSS plugins, all of their Markdown files are formatted with Prettier.

- https://github.com/eslint/json/blob/main/.prettierignore
- https://github.com/eslint/css/blob/main/.prettierignore

I'd like to ask how the team would feel about formatting all Markdown files in this repo with Prettier. If that's out of scope for this PR, I'm happy to open a separate issue.

The Markdown repository includes code-block linting examples, so if the main concern is those examples, not adding Prettier here seems reasonable to me.

.prettierignore
```diff
@@ -2,3 +2,4 @@ dist
 CHANGELOG.md
 jsr.json
 **/*.md
+!/README.md
```
lumirlumir, line 5: Actually, looking at the JSON and CSS plugins, all of their Markdown files are formatted with Prettier.

- https://github.com/eslint/json/blob/main/.prettierignore
- https://github.com/eslint/css/blob/main/.prettierignore

I'd like to ask how the team would feel about formatting all Markdown files in this repo with Prettier. (If that's out of scope for this PR, I'm happy to open a separate issue.)

The Markdown repository includes code-block linting examples, so if the main concern is those examples, not adding Prettier here seems reasonable to me.

nzakas: In theory it shouldn't matter as we should be able to properly lint the Markdown files even after they're formatted. I'd say try formatting them all and then linting. If there aren't any errors (or any errors that aren't easily fixable), then we can move forward with formatting all the .md files.

lumirlumir: Thank you for taking a look.

I tried it, but Prettier is more opinionated with Markdown formatting than I expected. (It changes some of the rule examples that we intentionally included.) So, I think it's fine to just format the root-level `README.md` file as it is now.

mdjermanovic (approved):
LGTM, thanks!

PR #515, opened 2025-08-31
fix: no-*-definitions rules show original label and identifier

### LINKED ISSUE #514: Bug: no-*-definitions rules show normalized identifier in messages
### Environment

ESLint version: v9.34.0
@eslint/markdown version: 7.2.0
Node version:
npm version:
Operating System:

commonmark

### What did you do?

<details>

```
```
</details>

```m
<!-- eslint markdown/no-empty-definitions: "error" -->
<!-- eslint markdown/no-duplicate-definitions: "error" -->
<!-- eslint markdown/no-unused-definitions: "error" -->

[Mercury]: https://example.com/mercury/
[Mercury]: https://example.com/mercury/
[Mercury]: #
```

Messages should display the original label as typed in the Markdown, e.g.:
- Unexpected unused definition `Mercury` found.
- Unexpected duplicate definition `Mercury` found.
- Unexpected empty definition `Mercury` found.

### What actually happened?

Messages display the normalized identifier, not the original label:
```
Unexpected unused definition `mercury` found.
```

n/a

### Participation

### Additional comments

_No response_

crimsonjay0:
#### Prerequisites checklist

The purpose is to improve the clarity and usability of linting feedback by ensuring messages reflect the exact labels written in Markdown, helping users quickly identify and fix issues without confusion.

Updated the `no-*-definitions` rules so that reported messages display the original label as written in the Markdown source, instead of the normalized identifier. This makes error messages clearer and consistent with user expectations.

#### Related Issues

Closes #514

### DIFF
--- a/src/rules/no-duplicate-definitions.js
+++ b/src/rules/no-duplicate-definitions.js
@@ -14,6 +14,7 @@ import { normalizeIdentifier } from "micromark-util-normalize-identifier";
 //-----------------------------------------------------------------------------
 
 /**
+ * @import { Definition, FootnoteDefinition } from "mdast";
  * @import { MarkdownRuleDefinition } from "../types.js";
  * @typedef {"duplicateDefinition" | "duplicateFootnoteDefinition"} NoDuplicateDefinitionsMessageIds
  * @typedef {[{ allowDefinitions?: string[], allowFootnoteDefinitions?: string[] }]} NoDuplicateDefinitionsOptions
@@ -85,10 +86,10 @@ export default {
 			),
 		);
 
-		/** @type {Map<string, { line: number, label: string }>} */
+		/** @type {Map<string, Definition>} */
 		const definitions = new Map();
 
-		/** @type {Map<string, { line: number, label: string }>} */
+		/** @type {Map<string, FootnoteDefinition>} */
 		const footnoteDefinitions = new Map();
 
 		return {
@@ -98,22 +99,22 @@ export default {
 				}
 
 				if (definitions.has(node.identifier)) {
-					const firstDefinition = definitions.get(node.identifier);
+					const firstDefinitionNode = definitions.get(
+						node.identifier,
+					);
 					context.report({
 						node,
 						messageId: "duplicateDefinition",
 						data: {
 							identifier: node.identifier,
 							label: node.label.trim(),
-							firstLine: firstDefinition.line.toString(),
-							firstLabel: firstDefinition.label.trim(),
+							firstLine:
+								firstDefinitionNode.position.start.line.toString(),
+							firstLabel: firstDefinitionNode.label.trim(),
 						},
 					});
 				} else {
-					definitions.set(node.identifier, {
-						line: node.position.start.line,
-						label: node.label,
-					});
+					definitions.set(node.identifier, node);
 				}
 			},
 
@@ -123,7 +124,7 @@ export default {
 				}
 
 				if (footnoteDefinitions.has(node.identifier)) {
-					const firstDefinition = footnoteDefinitions.get(
+					const firstFootnoteDefinitionNode = footnoteDefinitions.get(
 						node.identifier,
 					);
 					context.report({
@@ -132,15 +133,13 @@ export default {
 						data: {
 							identifier: node.identifier,
 							label: node.label,
-							firstLine: firstDefinition.line.toString(),
-							firstLabel: firstDefinition.label,
+							firstLine:
+								firstFootnoteDefinitionNode.position.start.line.toString(),
+							firstLabel: firstFootnoteDefinitionNode.label,
 						},
 					});
 				} else {
-					footnoteDefinitions.set(node.identifier, {
-						line: node.position.start.line,
-						label: node.label,
-					});
+					footnoteDefinitions.set(node.identifier, node);
 				}
 			},
 		};


tests/rules/no-duplicate-definitions.test.js
```diff
@@ -237,7 +237,7 @@ ruleTester.run("no-duplicate-definitions", rule, {
 			errors: [
 				{
 					messageId: "duplicateDefinition",
-					data: { identifier: "mercury" },
+					data: { identifier: "Mercury" },
```
mdjermanovic: While looking at this example, I think it makes more sense to show the normalized `identifer` in the message, because `Mercury` doesn't appear elsewhere in the code, so this is a duplicate of `mercury` because of the normalization.

lumirlumir (commented):
I'm +1 with exposing the label in the message, but I'm not sure removing the identifier entirely from message is the best approach.

Could we expose both the identifier and the label in the message? Here are some reasons:

Identifiers may be normalized in non-English languages (and, for English, all letters are converted to lowercase). For example:

https://github.com/eslint/markdown/blob/4d4f12403a7df942d6e1cfb07401d603c2b07bc0/tests/rules/no-duplicate-definitions.test.js#L162-L173

If we write `Grüsse` and `GRÜẞE`, the rule reports them as the same definition identifier.

That could be confusing if only labels are shown. For example, a user might intentionally create one definition as `Grüsse` and another as `GRÜẞE` (Actually, there are many cases like this), yet the rule flags a conflict because the identifier is normalized. This may confuse users who aren't aware that definition identifiers are normalized.

---

So, If others agree, I’d like to suggest using a message format like the following:

```diff
- Unexpected duplicate definition `{{ identifier }}` found.
+ Unexpected duplicate definition `{{ identifier }}` (label: `{{ label }}`) found.
```

(The message format may vary if other suggestions are available.)

tests/rules/no-duplicate-definitions.test.js
crimsonjay0: I understand your point about the normalization in this case, but if we show the normalized identifier, we will run into issues like this:
```markdown
[Mercury]: https://example.com/mercury/
[Mercury]: https://example.com/venus/
```
In this example, showing the normalized `identifier` could be misleading, since `mercury` doesn’t appear elsewhere in the code. I think showing both the identifier and the label would be a better approach.

mdjermanovic:
> > Ideally, the message could also include location and label of the first definition (the one of which this is a duplicate of).
> 
> How about this phrasing?
> 
> ```
> "Unexpected duplicate definition `{{ identifier }}` (label: `{{ label }}`) found. First defined at line {{ firstLine }} (label: `{{ firstLabel }}`)."
> ```

Looks good to me 👍

tests/rules/no-duplicate-definitions.test.js
mdjermanovic: I agree it makes sense to show both, as in https://github.com/eslint/markdown/pull/515#issuecomment-3248588483.

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for @lumirlumir to verify.

src/rules/no-unused-definitions.js
```diff
@@ -130,7 +130,10 @@ export default {
 						context.report({
 							node: definition,
 							messageId: "unusedDefinition",
-							data: { identifier: definition.identifier },
+							data: {
+								identifier: definition.identifier,
+								label: definition.label.trim(),
```
lumirlumir, line 135:
```suggestion
								label: definition.label,
```

Could we avoid using `trim` here?

I think the purpose of exposing `label` in the message is to show users the raw form so they can identify the problematic definition more easily.

`definition.label` always provide unnormalized raw values, and since we already expose the normalized value via `identifier`, including the raw values in the message seems more appropriate.

(The same goes for the other rules.)

src/rules/no-empty-definitions.js
```diff
@@ -115,7 +115,10 @@ export default {
 					context.report({
 						loc: node.position,
 						messageId: "emptyDefinition",
-						data: { identifier: node.identifier },
+						data: {
+							identifier: node.identifier,
+							label: node.label.trim(),
```
lumirlumir, line 120:
```suggestion
							label: node.label,
```

src/rules/no-duplicate-definitions.js
```diff
@@ -98,13 +98,22 @@ export default {
...
+					const firstDefinition = definitions.get(node.identifier);
 					context.report({
 						node,
 						messageId: "duplicateDefinition",
-						data: { identifier: node.identifier },
+						data: {
+							identifier: node.identifier,
+							label: node.label.trim(),
+							firstLine: firstDefinition.line.toString(),
+							firstLabel: firstDefinition.label.trim(),
```
lumirlumir:
```suggestion
							label: node.label,
							firstLine: firstDefinition.line.toString(),
							firstLabel: firstDefinition.label,
```

src/rules/no-duplicate-definitions.js
```diff
@@ -114,13 +123,24 @@ export default {
 				}
 
 				if (footnoteDefinitions.has(node.identifier)) {
+					const firstDefinition = footnoteDefinitions.get(
```
lumirlumir:
```suggestion
					const firstFootnoteDefinition = footnoteDefinitions.get(
```

Maybe adding a `Footnote` would be clearer here.

src/rules/no-duplicate-definitions.js
```diff
@@ -85,11 +85,11 @@ export default {
 			),
 		);
 
-		/** @type {Set<string>} */
-		const definitions = new Set();
+		/** @type {Map<string, { line: number, label: string }>} */
+		const definitions = new Map();
```
lumirlumir:
```suggestion
		/** @type {Map<string, Definition>} */
		const definitions = new Map();
```

Could we use the existing `Definition` type data structure here?

Using the `Definition` type would allow us to reuse existing object references and make the code structure simpler and clearer.

FYI, the `Definition` and `FootnoteDefinition` types can be imported as follows:

```diff
/**
+* @import { Definition, FootnoteDefinition } from "mdast";
 * @import { MarkdownRuleDefinition } from "../types.js";
 * ...
 ```
 
Maybe the final code would look like this:

image

File: src/rules/no-duplicate-definitions.js [hunk 7]
```diff
@@ -85,11 +85,11 @@ export default {
...
 
-		/** @type {Set<string>} */
-		const footnoteDefinitions = new Set();
+		/** @type {Map<string, { line: number, label: string }>} */
```
lumirlumir:
```suggestion
		/** @type {Map<string, FootnoteDefinition>} */
```

Maybe the final code would look like this:

image

src/rules/no-unused-definitions.js
crimsonjay0: Hmm, not too sure about this approach. For something like `[ Mercury ]`, we end up with extra spaces in the message, which just looks a bit messy.

Might be worth checking with @mdjermanovic before moving forward?

src/rules/no-duplicate-definitions.js
crimsonjay0: But this means we’d need to store the full node when we only need the label and start line. If that’s okay, I can go ahead and implement the change.

mdjermanovic: It's okay, we'll only store references to objects that would exist anyway, so there's no perf cost (and there's even a small gain as we are not creating new objects).

src/rules/no-unused-definitions.js
lumirlumir: I think the `label` message will be wrapped in backticks, so it wouldn't be a big problem.

Also, for this change, we could make the following code more distinguishable:

```md
[    hi    ]: url1

[hi]: url2

[ hi]: url3
```

Currently, the above code shows the same error messages.

---

I'd also love to hear @mdjermanovic's thoughts on this, if available.

mdjermanovic: I think it's okay to trim labels.

lumirlumir (approved):
LGTM, Thanks.

Would like @mdjermanovic to verify the changes before merging.

mdjermanovic (approved):
LGTM, thanks!

PR #520, opened 2025-09-07
feat: export types from main entry point

lumirlumir:
#### Prerequisites checklist

In this PR, I've exported all types from `src/types.ts` through the main entry point.

Previously, to access the global types defined in `src/types.ts`, they had to be imported from `@eslint/markdown/types`. After this change, they can be imported directly from `@eslint/markdown`.

To enable importing types from the main entry point, I used a small trick.

The `types.ts` file is built into `types.js` and `types.d.ts`.

In the built `types.js`, it's possible to re-export the type declarations from the main entry point using `export * from "types.js"`.

However, that wasn't possible in the source. To bridge the gap, I added a dummy `src/types.js` file so types can be imported from `src/types.ts`.

#### Related Issues

Ref: https://github.com/eslint/json/issues/130, https://github.com/eslint/json/issues/130#issuecomment-3234058376, https://github.com/eslint/markdown/pull/453, https://github.com/eslint/markdown/pull/367#issuecomment-3010174602

N/A

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -135,3 +135,4 @@ recommendedPlugins.markdown = processorPlugins.markdown = plugin;
 
 export default plugin;
 export { MarkdownSourceCode };
+export * from "./types.js";
--- a/src/types.js
+++ b/src/types.js
@@ -1,0 +1,10 @@
+/**
+ * @fileoverview Dummy file to enable exporting types from the main module.
+ * @author 루밀LuMir(lumirlumir)
+ */
+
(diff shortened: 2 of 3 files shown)


JoshuaKGoldberg (approved):
Great! IMO a single almost-blank file is not a very high price to pay for easier type imports.

Thanks for pushing on this @lumirlumir!

lumirlumir:
@JoshuaKGoldberg Thanks for understanding that I took on the issue you originally suggested.

PR #521, opened 2025-09-13
fix: false positives in `no-reversed-media-syntax`

lumirlumir:
## Prerequisites checklist

Working in progress.


tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -255,6 +265,20 @@ ruleTester.run("no-reversed-media-syntax", rule, {
 				},
 			],
 		},
+		{
```
lumirlumir, line 268: This test case isn't related to this change, but the test for the `Image` node within a `FootnoteReference` node was missing, so I've added one.

PR #524, opened 2025-09-14
fix: `exactOptionalPropertyTypes` causes type errors when using plugins

lumirlumir:
## Prerequisites checklist

In this PR, I’ve resolved the issue mentioned in #402.

I’ve confirmed that everything is working correctly by adding `exactOptionalPropertyTypes` to the tests and testing locally with a modified `index.d.ts` file.

### Why does this bug occur?

The bug was caused by the type signature of `configs.processor`. The `configs.processor` type was showing an unintended type signature, which led to the issue.

image

For example, after removing `config.processor` from `index.d.ts` entirely, the type error no longer occurs.

---

Looking into it further, the issue was caused by the commented "key-value" pair below. If we comment out these "key-value" pairs, the error no longer occurs.

스크린샷 2025-09-15 011939

The commented-out "key-value" pairs above correspond to the JavaScript code below:

image

There were no `files?: undefined;`, `processor?: undefined;`, or similar "key-value" pairs in the original JavaScript code, but they suddenly appeared when running `tsc`. These properties ended up causing the bug.

---

To address the issue, I cast the type of `configs.processor` as `Linter.Config[]`.

With this type casting, the `configs.processor` type now appears as shown in the following screenshot.

image

In this PR, I’ve resolved the issue mentioned in #402.

## Related Issues

Fixes: #402

N/A

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,14 +19,13 @@ import rules from "./build/rules.js";
 
 /**
  * @import { Linter } from "eslint";
- * @typedef {Linter.RulesRecord} RulesRecord
  */
 
 //-----------------------------------------------------------------------------
 // Exports
 //-----------------------------------------------------------------------------
 
-/** @type {RulesRecord} */
+/** @type {Linter.RulesRecord} */
 const processorRulesConfig = {
 	// The Markdown parser automatically trims trailing
 	// newlines from code blocks.
@@ -98,7 +97,7 @@ const plugin = {
 				rules: recommendedRules,
 			},
 		],
-		processor: [
+		processor: /** @type {Linter.Config[]} */ ([
 			{
 				name: "markdown/recommended/plugin",
 				plugins: (processorPlugins = {}),
@@ -126,7 +125,7 @@ const plugin = {
 					...processorRulesConfig,
 				},
 			},
-		],
+		]),
 	},
 };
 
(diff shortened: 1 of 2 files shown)


tests/types/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "emitDeclarationOnly": false,
     "noEmit": true,
     "rootDir": "../..",
-    "strict": true
+    "strict": true,
+    "exactOptionalPropertyTypes": true
```
lumirlumir, line 8: Since the original test is well-implemented, simply adding `exactOptionalPropertyTypes` allows us to test both before and after.

If we run `npm run build && npm run test:types`, the tests will fail if we revert the changes in `src/index.js`.

---

Just for your information, `exactOptionalPropertyTypes` isn't included in `strict` mode, so it needs to be enabled manually.

https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes

src/index.js
```diff
@@ -19,14 +19,13 @@ import rules from "./build/rules.js";
...
  * @import { Linter } from "eslint";
- * @typedef {Linter.RulesRecord} RulesRecord
  */
 
 //-----------------------------------------------------------------------------
 // Exports
 //-----------------------------------------------------------------------------
 
-/** @type {RulesRecord} */
+/** @type {Linter.RulesRecord} */
```
lumirlumir, line 28: Just a minor refactor.

PR #545, opened 2025-09-30
fix: wrong location reporting in `no-invalid-label-refs`

lumirlumir:
## Prerequisites checklist

CommonMark and GFM.

### What did you do?

The error location reported by `no-invalid-label-refs` is correct for single-line text, but in multi-line text, it's one index(offset) behind where I expect.

image

I expected the error location to begin at `[` and end at `]`, but in multi-line text the end column is one index behind.

Adding the following test case to the `invalid` section would help identify the problem.

```js
		{
			code: "[\nfoo\n][\n]\n\n[foo]: http://bar.com/image.jpg",
			errors: [
				{
					messageId: "invalidLabelRef",
					data: { label: "foo" },
					line: 3,
					column: 2,
					endLine: 4,
					endColumn: 2,
				},
			],
		},
```

Originally, the `endColumn` was `1`, but it needs to be `2` so it includes `]` as it does for single-line text.

In this PR, I've fixed wrong location reporting in `no-invalid-label-refs`.

## Related Issues

N/A

N/A

### DIFF
--- a/docs/rules/no-invalid-label-refs.md
+++ b/docs/rules/no-invalid-label-refs.md
@@ -30,7 +30,6 @@ Examples of **incorrect** code for this rule:
 [eslint][ ]
 
 [eslint][
-
 ]
 ```
 
--- a/src/rules/no-invalid-label-refs.js
+++ b/src/rules/no-invalid-label-refs.js
@@ -102,7 +102,7 @@ function findInvalidLabelReferences(node, sourceCode) {
 		const startColumn = nodeStartColumn + startColumnOffset;
 		const endLine = nodeStartLine + endLineOffset;
 		const endColumn =
-			(endLine === startLine ? nodeStartColumn : 0) + endColumnOffset;
+			(endLine === startLine ? nodeStartColumn : 1) + endColumnOffset;
 
 		invalid.push({
 			label: label.trim(),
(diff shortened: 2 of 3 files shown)


docs/rules/no-invalid-label-refs.md
```diff
@@ -30,7 +30,6 @@ Examples of **incorrect** code for this rule:
 [eslint][ ]
 
 [eslint][
-
 ]
```
lumirlumir, line 33: The following example was correct code, so I've updated it to show incorrect code.

(I've also added a test for it.)

```md
[eslint][

]
```

image

src/rules/no-invalid-label-refs.js
```diff
@@ -102,7 +102,7 @@ function findInvalidLabelReferences(node, sourceCode) {
 		const startColumn = nodeStartColumn + startColumnOffset;
 		const endLine = nodeStartLine + endLineOffset;
 		const endColumn =
-			(endLine === startLine ? nodeStartColumn : 0) + endColumnOffset;
+			(endLine === startLine ? nodeStartColumn : 1) + endColumnOffset;
```
lumirlumir, line 105: The column is a 1-indexed value, so it needs to be `1`, not `0`.

https://github.com/eslint/markdown/blob/bae7da727c77e9f5535586c36ace2a1267574fcf/src/language/markdown-language.js#L118-L122

mdjermanovic (approved):
LGTM, thanks!

PR #571, opened 2025-10-29
fix: correct parent typing for rule visitors

crimsonjay0:
#### Prerequisites checklist

This PR updates `MarkdownRuleVisitor` so that the parent parameter has a precise type for non-root nodes.

- Added `ParentOf<T>` utility to infer correct parent types.
- For non-root nodes, the `parent` parameter is now precisely typed; for root, only the `node` is passed.
- Introduced `MarkdownSyntaxElement`, consistent with the CSS and JSON plugins.
- Updated `MarkdownSourceCode` to reference `MarkdownSyntaxElement`.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -8,35 +8,11 @@ import type {
 	Data,
 	Literal,
 	Parent,
-	// Nodes
-	Blockquote,
-	Break,
+	Parents,
+	// Node unions
+	Nodes,
 	Code,
-	Definition,
-	Emphasis,
-	Heading,
-	Html,
-	Image,
-	ImageReference,
-	InlineCode,
-	Link,
-	LinkReference,
-	List,
-	ListItem,
-	Paragraph,
 	Root,
-	Strong,
-	Text,
-	ThematicBreak,
-	// Extensions (GFM)
-	Delete,
-	FootnoteDefinition,
-	FootnoteReference,
-	Table,
-	TableCell,
-	TableRow,
-	// Extensions (front matter)
-	Yaml,
 } from "mdast";
 import type {
 	CustomRuleDefinitionType,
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -58,6 +34,25 @@ type WithExit<RuleVisitorType extends RuleVisitor> = {
...
 };
 
+/**
+ * Compute the precise parent type for a given node `T`.
+ * - Root has no parent (returns `never`).
+ * - Custom frontmatter nodes (`toml` / `json`) only appear directly under `Root`.
+ * - For nodes, include every parent `P` where `T` is assignable to one of
+ *   `P['children'][number]` (i.e. `T` can appear in `P.children`).
+ */
+type ParentOf<T extends MarkdownNode> = T extends Root
```
DMartens, line 44: I think this should be static for performance and simplicity reasons as the parent of the various markdown nodes should be easily groupable (inside paragraph (e.g. `Strong` and `Emphasis`) and root (e.g. `Heading` and `Code`).

src/types.ts
```diff
@@ -8,35 +8,11 @@ import type {
...
-	ThematicBreak,
-	// Extensions (GFM)
-	Delete,
-	FootnoteDefinition,
-	FootnoteReference,
-	Table,
-	TableCell,
-	TableRow,
-	// Extensions (front matter)
-	Yaml,
```
lumirlumir, line 39: Originally, there was consensus not to use the `RootContent` type.

Currently, the `Nodes` type is equivalent to `Root | RootContent`, so I'm not sure that using `Nodes` here is the desired approach.

I personally prefer not to use the `Nodes` type and to keep the original enumeration style.

Ref: https://github.com/eslint/markdown/pull/334#discussion_r2040617231

src/types.ts
lumirlumir: Agreed. Also, deeply nested conditional types make the code somewhat unclear to me.

PR #603, opened 2026-01-04
fix: recognize `Definition` node in `no-missing-link-fragments`

lumirlumir:
## Prerequisites checklist

CommonMark and GFM.

### What did you do?

I expected the `[reference]: #non-existent` definition to be reported by the `no-missing-link-fragments` rule, but it isn't.

The `Invalid` link reference is also considered valid link when using reference-style links.

image

I expected the `[reference]: #non-existent` definition to be reported by the `no-missing-link-fragments` rule.

The following example would help identify the problem:

```md
<!-- eslint markdown/no-missing-link-fragments: "error" -->

<!-- Valid -->

[Valid](#non-existent)

<!-- Invalid (False Negative) -->

[Invalid][reference]

[reference]: #non-existent
```

In this PR, I've fixed a false negative caused by an unrecognized `Definition` node in `no-missing-link-fragments`.

This issue was discovered while working on #583, and to avoid making that PR too large, I split this into a separate PR.

---

Previously, the rule did not handle a `Definition` node like the following:

https://github.com/eslint/markdown/blob/8f92161ff9ffc718e5949c08c0b777bf56af48ce/src/rules/no-missing-link-fragments.js#L129-L141

I've updated the logic to recognize `Definition` nodes accordingly.

---

I've also refactored the codebase. There was some unnecessary or duplicate logic in the rule, so I refactored it to make the rule logic clearer, and faster. I've left comments on the notable parts of this change.

## Related Issues

Ref: #583

N/A


tests/rules/no-missing-link-fragments.test.js
```diff
@@ -520,6 +547,36 @@ ruleTester.run("no-missing-link-fragments", rule, {
...
+				{
+					messageId: "invalidFragment",
+					data: { fragment: "l20-l30" },
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 27,
+				},
+			],
+		},
```
lumirlumir, line 579: These test cases were added primarily to ensure that `githubLineReferencePattern` is case-insensitive.

Before adding this test, the suite passed even when I added the `i` flag to the following regex pattern:

https://github.com/eslint/markdown/blob/8f92161ff9ffc718e5949c08c0b777bf56af48ce/src/rules/no-missing-link-fragments.js#L29

---

FYI: You can verify that the `githubLineReferencePattern` is case-insensitive using the following examples:

- OK

https://github.com/eslint/markdown/blob/main/package.json#L4
https://github.com/eslint/markdown/blob/main/package.json#L4-L6

- Not OK

https://github.com/eslint/markdown/blob/main/package.json#l4 // Unworking.
https://github.com/eslint/markdown/blob/main/package.json#l4-l6 // Unworking.

src/rules/no-missing-link-fragments.js
```diff
@@ -101,58 +100,52 @@ export default {
...
 			"heading:exit"() {
 				const customIdMatch = headingText.match(customHeadingIdPattern);
-				const baseId = customIdMatch
+				const id = customIdMatch
 					? customIdMatch.groups.id
 					: headingText;
-				const finalId = slugger.slug(baseId);
-				fragmentIds.add(ignoreCase ? finalId.toLowerCase() : finalId);
+
+				fragmentIds.add(slugger.slug(id));
```
lumirlumir, line 107: Previously, we used the pattern `ignoreCase ? finalId.toLowerCase() : finalId` to normalize the `fragmentId`. However, this wasn't necessary because `GithubSlugger#slug` already uses `toLowerCase` internally by default, so the result is the same whether we normalize it or not after being slugified.

Ref: https://github.com/Flet/github-slugger/blob/master/index.js#L75

Image

src/rules/no-missing-link-fragments.js
```diff
@@ -101,58 +100,52 @@ export default {
...
-				const fragment = url.slice(1);
-				if (!fragment) {
+				// If `url` is empty, `"#"`, or does not start with `"#"`, skip it.
+				if (url === "" || url === "#" || !url.startsWith("#")) {
 					return;
 				}
 
-				linkNodes.push({ node, fragment });
+				relevantNodes.push(node);
 			},
```
lumirlumir, line 133: This is a small improvement: I've simplified the logic to reduce a bit of memory usage and computation.

Previously, we created a new data structure type as follows:

https://github.com/eslint/markdown/blob/8f92161ff9ffc718e5949c08c0b777bf56af48ce/src/rules/no-missing-link-fragments.js#L88-L89

https://github.com/eslint/markdown/blob/8f92161ff9ffc718e5949c08c0b777bf56af48ce/src/rules/no-missing-link-fragments.js#L140

However, now we no longer create a new object and instead use an existing object reference; it's small, but it would reduce memory usage and computation.

src/rules/no-missing-link-fragments.js
DMartens: Shouldn't we then pass `!ignoreCase` as the `maintainCase` argument as now the fragment id is always lowercased?
Further, as this is the only function we use from "github-slugger", we could inline this function as it is only two lines.

lumirlumir: > Shouldn't we then pass !ignoreCase as the maintainCase argument as now the fragment id is always lowercased?

Although the `github-slugger` library allows setting case sensitivity for slugs, GitHub's actual slugging algorithm matches the default `maintainCase=false` behavior.

For example, the heading `# FOO BAR BAZ` always yields `#foo-bar-baz` regardless of capitalization.

There was also a [consensus](https://github.com/eslint/markdown/issues/369#issuecomment-2872822560) when this rule was introduced to mimic the GitHub slugging algorithm used by `markdownlint`, so I believe this is the intended behavior.

> Further, as this is the only function we use from "github-slugger", we could inline this function as it is only two lines.

I think I may have missed the exact point, but as I understand it, there are two places that use `slugger.slug(id)`, and the `slugger` instance is still needed because it needs to track previously generated slugs. (Ref: https://github.com/Flet/github-slugger/blob/master/index.js#L23-L24) 

Simply importing `slug` with `import { slug } from "github-slugger";` won't work here.

DMartens: Regarding `ignoreCase`: As the rule should reflect the original markdownlint behavior, it is weird that `ignoreCase` was used here before, as [markdownlint rule](https://github.com/DavidAnson/markdownlint/blob/63fefcbd4a7cbfa4fb49b40b5c4020c25df24c7d/lib/md051.mjs#L37) always lowercases, so you are right.
Regarding removing "github-slugger": I thought the function in the screenshot was all but there is also an additional counter

lumirlumir: Is there anything else you'd like me to adjust here, or is everything fine?

snitin315 (approved):
LGTM, thanks! Leaving it open for @DMartens

DMartens:
Changes LGTM, thanks.

PR #617, opened 2026-01-31
feat: support Math

lumirlumir:
## Prerequisites checklist

This PR supports the `math` option as described in #331.

I've reviewed all of the current rule implementations. The only affected rule was `no-reversed-media-syntax`. I've added test cases to ensure it works. I also added some tests for `no-missing-link-fragments` because I suspected this change might affect it. However, there was no impact, so I only added tests. The newly added test cases also include the examples that were mentioned in https://github.com/eslint/markdown/issues/331 as false positives.

I've left separate comments on several notable or questionable parts of this PR.

This PR supports the `math` option as described in #331.

## Related Issues

Closes: #331
Ref: https://github.com/eslint/markdown/pull/328

N/A


src/language/markdown-language.js
```diff
@@ -175,6 +188,15 @@ export class MarkdownLanguage {
 				`Invalid language option value \`${frontmatterOption}\` for frontmatter.`,
 			);
 		}
+
+		// `math` option validation
+		const mathOption = languageOptions?.math;
+
+		if (mathOption !== undefined && typeof mathOption !== "boolean") {
```
lumirlumir, line 195: Question:

Are the `undefined` checks (e.g. `mathOption !== undefined`, `frontmatterOption !== undefined`) and the optional chaining (`languageOptions?.math`, `languageOptions?.frontmatter`) still necessary? (I initially implemented it to check for `undefined`.)

https://github.com/eslint/markdown/blob/3a8169cb3f96088d7a08db40669d84219e2ab024/src/language/markdown-language.js#L171

https://github.com/eslint/markdown/blob/3a8169cb3f96088d7a08db40669d84219e2ab024/src/language/markdown-language.js#L162

---

Initially, the motivation for introducing `frontmatterOption !== undefined` was in https://github.com/eslint/markdown/pull/328. At that time, the `README.md` stated that the minimum compatible ESLint version was any v9 release, so it made sense.

Later, we discovered that rules used `defaultOptions`, and we raised the minimum ESLint version to v9.15.0 (ref: https://github.com/eslint/markdown/pull/537).

The issue is that `defaultLanguageOptions` was introduced in [ESLint v9.13.0](https://github.com/eslint/eslint/releases/tag/v9.13.0). Since we now expect users to be running ESLint v9.15.0 or later, the `frontmatter` and `math` options can't be `undefined`.

https://github.com/eslint/markdown/blob/3a8169cb3f96088d7a08db40669d84219e2ab024/src/language/markdown-language.js#L134-L136

---

However, currently there are no `peerDependencies` or other restrictions on the language plugins, so checking for `undefined` still seems sensible: some users might not depend on the rule and could only be using `MarkdownSourceCode` or `MarkdownLanguage` features with ESLint versions older than v9.13.0.

To summarize, for safety, would it be better to keep the `undefined` checks and optional chaining, or is it safe to remove them? If opening a separate issue would help, I'm happy to do that and work on it.

src/rules/no-reversed-media-syntax.js
```diff
@@ -65,12 +66,12 @@ export default {
 				nodeStartOffset = node.position.start.offset;
 			},
 
-			":matches(heading, paragraph, tableCell) :matches(html, image, imageReference, inlineCode, linkReference)"(
-				/** @type {Html | Image | ImageReference | InlineCode | LinkReference} */ node,
+			":matches(heading, paragraph, tableCell) :matches(html, image, imageReference, inlineCode, linkReference, inlineMath)"(
```
lumirlumir, line 69: This PR introduces two node types: `math` and `inlineMath`.

However, the `math` node is block-level, so it cannot be a child of a `heading`, `paragraph` or `tableCell` node. Only the `inlineMath` node can be a child of a `heading`, `paragraph` or `tableCell`, so I only added `inlineMath` to it.

tests/language/markdown-language.test.js
```diff
@@ -38,13 +27,39 @@ describe("MarkdownLanguage", () => {
...
-		it("should not throw an error when `frontmatter` is not provided and other keys are present", () => {
+		it("should not throw an error when `frontmatter` or `math` is not provided and other keys are present", () => {
 			const language = new MarkdownLanguage();
 			assert.doesNotThrow(() => {
 				language.validateLanguageOptions({ foo: "bar" });
 			});
 		});
 
+		// Validation tests for the `frontmatter` option
+		it("should throw an error if `frontmatter` is not `false`, `'yaml'`, `'toml'`, or `'json'`", () => {
```
lumirlumir, line 38: The previous test logic for `validateLanguageOptions()` mixed frontmatter-specific tests with general tests, so I moved the general test suites to the top level. I also replaced the previous frontmatter error test that used a regex with a check against the full error message for better accuracy.

In summary, this diff here mainly restructures the test suites for improved clarity.

src/language/markdown-language.js
DMartens: I agree that we should keep the optional chaining on `languageOptions` and `undefined` checks until the next major version.

src/language/markdown-language.js
```diff
@@ -175,6 +188,15 @@ export class MarkdownLanguage {
...
+			throw new Error(
+				`Invalid language option value \`${mathOption}\` for math.`,
```
DMartens: This error message should be more helpful (I know it has the same message as the error message for `frontmatter`).
I think we should update both error messages to include the expected values, e.g. "Invalid language option \`${mathOption}\` for math. Expected a boolean".

lumirlumir: Sounds good to me. I've updated it in https://github.com/eslint/markdown/pull/617/changes/c7c38ed5c0353f2bc78b22aaa73bf5b27c7cfea2.

FYI: I also looked into the error message formats used by the JSON and CSS plugins, but it seems there isn’t a common template.

- https://github.com/eslint/json/blob/main/src/languages/json-language.js#L87
- https://github.com/eslint/css/blob/main/src/languages/css-language.js#L147

PR #618, opened 2026-02-05
fix: report messages for configuration comments correctly

### LINKED ISSUE #115: This does not work with `eslint --report-unused-disable-directives`
`eslint --report-unused-disable-directives` is not reported when a `<-- eslint-disable -->` is defined in Markdown but is unused (i.e. could be removed).

`.eslintrc.yml`:

```yml
plugins: [markdown]
rules:
  no-inline-comments: 2
```

`README.md`:

````md
<!-- eslint-disable no-inline-comments -->

```js
var aa = true
```
````

`index.js`:

```js
/* eslint-disable no-inline-comments */
var aa = true
```

`package.json`:

```json
{
  "name": "eslint-comments-bug",
  "version": "1.0.0",
  "license": "ISC",
  "dependencies": {
    "eslint": "^5.16.0",
    "eslint-plugin-eslint-comments": "^3.1.1",
    "eslint-plugin-markdown": "^1.0.0"
  }
}
```

In the terminal:

```bash
$ eslint --report-unused-disable-directives README.md
$ eslint --report-unused-disable-directives index.js

/home/ehmicky/eslint-comments-bug/index.js
  1:1  error  Unused eslint-disable directive (no problems were reported from 'no-inline-comments')

```

`--report-unused-disable-directives` works for the JavaScript file, but not for the Markdown (where nothing is reported).

andreahlert:
## Summary

Previously, ESLint messages that pointed to configuration comments (such as unused disable directive warnings) were being silently dropped because the `adjustMessage` function returned `null` for any message pointing to lines before the actual code content.

This change:
- Tracks the original HTML comment positions when collecting configuration comments from the Markdown
- Builds a mapping from linted code line numbers to original comment info
- Maps messages pointing to configuration comment lines back to their original HTML comment locations in the Markdown file

This fixes the issue where `--report-unused-disable-directives` would not work with the markdown processor, as those warnings were being filtered out.

### Before
```
$ eslint --report-unused-disable-directives README.md
# (no output even when there are unused disable directives)
```

### After
```
$ eslint --report-unused-disable-directives README.md
README.md
  3:1  warning  Unused eslint-disable directive (no problems were reported from 'no-undef')
```

## Test plan
- [x] Added new test suite `reportUnusedDisableDirectives` with tests for single-line and multi-line HTML comments
- [x] All existing 1590 tests pass
- [x] Lint passes

Fixes #115

### DIFF
--- a/docs/processors/markdown.md
+++ b/docs/processors/markdown.md
@@ -180,6 +180,8 @@ Comment bodies are passed through unmodified, so the plugin supports any [config
 
 This example enables the `alert` global variable, disables the `no-alert` rule, and configures the `quotes` rule to prefer single quotes:
 
+<!-- eslint-skip -->
+
 ````markdown
 <!-- global alert -->
 <!-- eslint-disable no-alert -->
@@ -191,6 +193,8 @@ alert('Hello, world!');
 ````
 
 Each code block in a file is linted separately, so configuration comments apply only to the code block that immediately follows.
+
+<!-- eslint-skip -->
 
 ````markdown
 Assuming `no-alert` is enabled in `eslint.config.js`, the first code block will have no error from `no-alert`:
--- a/eslint.config-content.js
+++ b/eslint.config-content.js
@@ -3,7 +3,7 @@ import markdown from "./src/index.js";
 
 export default defineConfig([
 	globalIgnores(
-		["**/*.js", "**/.cjs", "**/.mjs"],
+		["**/*.js", "**/.cjs", "**/.mjs", "**/tests/fixtures/**"],
 		"markdown/content/ignores",
 	),
 
--- a/src/processor.js
+++ b/src/processor.js
@@ -20,6 +20,7 @@ import { fromMarkdown } from "mdast-util-from-markdown";
  * @typedef {Linter.LintMessage} Message
  * @typedef {Rule.Fix} Fix
  * @typedef {AST.Range} Range
+ * @typedef {{ text: string, position: { start: { line: number, column: number }, end?: { line: number, column: number } } }} CommentInfo
  */
 
 //-----------------------------------------------------------------------------
@@ -31,6 +32,19 @@ const UNSATISFIABLE_RULES = new Set([
 	"unicode-bom", // Code blocks will begin in the middle of Markdown files
 ]);
 const SUPPORTS_AUTOFIX = true;
+
+/**
+ * Wraps comment text in a JS block comment with normalized spacing so that
+ * rules like `spaced-comment` are not triggered by the generated comments.
+ * @param {string} text The comment body extracted from the HTML comment.
+ * @returns {string} A JS block comment string.
+ */
+function wrapComment(text) {
+	const prefix = /^[\s*]/u.test(text) ? "" : " ";
+	const suffix = /[\s*]$/u.test(text) ? "" : " ";
+
+	return `/*${prefix}${text}${suffix}*/`;
+}
 
 const BOM = "\uFEFF";
 
@@ -281,7 +295,7 @@ function preprocess(sourceText, filename) {
 	 * block immediately follows such a sequence, insert the comments at the
 	 * top of the code block. Any non-ESLint comment or other node type breaks
 	 * and empties the sequence.
-	 * @type {string[]}
+	 * @type {CommentInfo[]}
 	 */
 	let htmlComments = [];
 
@@ -300,13 +314,17 @@ function preprocess(sourceText, filename) {
 				/** @type {string[]} */
 				const comments = [];
 
-				for (const comment of htmlComments) {
-					if (comment.trim() === "eslint-skip") {
+				/** @type {CommentInfo[]} */
+				const commentInfos = [];
+
+				for (const commentInfo of htmlComments) {
+					if (commentInfo.text.trim() === "eslint-skip") {
 						htmlComments = [];
 						return;
 					}
 
-					comments.push(`/*${comment}*/`);
+					comments.push(wrapComment(commentInfo.text));
+					commentInfos.push(commentInfo);
 				}
 
 				htmlComments = [];
@@ -315,6 +333,10 @@ function preprocess(sourceText, filename) {
 					...node,
 					baseIndentText: getIndentText(text, node),
 					comments,
+
+					// Original HTML comment positions, used to map
+					// configuration comment messages back to Markdown
+					commentInfos,
 					rangeMap: getBlockRangeMap(text, node, comments),
 				});
 			}
@@ -329,7 +351,10 @@ function preprocess(sourceText, filename) {
 			const comment = getComment(node.value);
 
 			if (comment) {
-				htmlComments.push(comment);
+				htmlComments.push({
+					text: comment,
+					position: node.position,
+				});
 			} else {
 				htmlComments = [];
 			}
@@ -382,7 +407,7 @@ function adjustFix(block, fix) {
 /**
  * Creates a map function that adjusts messages in a code block.
  * @param {Block} block A code block.
- * @returns {(message: Message) => Message} A function that adjusts messages in a code block.
+ * @returns {(message: Message) => Message | null} A function that adjusts messages in a code block.
  */
 function adjustBlock(block) {
 	const leadingCommentLines = block.comments.reduce(
@@ -392,10 +417,33 @@ function adjustBlock(block) {
 
 	const blockStart = block.position.start.line;
 
+	/*
+	 * Build a mapping from line numbers in the linted code (1-indexed) to the
+	 * original HTML comment info. This allows us to map messages that point to
+	 * configuration comments back to their original location in the Markdown.
+	 */
+	const commentLineMap = new Map();
+
+	if (block.commentInfos) {
+		let lineNumber = 1;
+
+		for (const commentInfo of block.commentInfos) {
+			// Each comment becomes `/*...*/\n` in the linted code
+			const commentText = wrapComment(commentInfo.text);
+			const commentLineCount = commentText.split("\n").length;
+
+			for (let i = 0; i < commentLineCount; i++) {
+				commentLineMap.set(lineNumber + i, commentInfo);
+			}
+
+			lineNumber += commentLineCount;
+		}
+	}
+
 	/**
 	 * Adjusts ESLint messages to point to the correct location in the Markdown.
 	 * @param {Message} message A message from ESLint.
-	 * @returns {Message} The same message, but adjusted to the correct location.
+	 * @returns {Message | null} The same message, but adjusted to the correct location.
 	 */
 	return function adjustMessage(message) {
 		if (!Number.isInteger(message.line)) {
@@ -408,7 +456,39 @@ function adjustBlock(block) {
 
 		const lineInCode = message.line - leadingCommentLines;
 
-		if (lineInCode < 1 || lineInCode >= block.rangeMap.length) {
+		/*
+		 * If the message points to a line before the actual code (i.e., in the
+		 * configuration comments), map it back to the original HTML comment
+		 * location in the Markdown.
+		 */
+		if (lineInCode < 1) {
+			const commentInfo = commentLineMap.get(message.line);
+
+			if (commentInfo) {
+				return {
+					...message,
+					line: commentInfo.position.start.line,
+					column: commentInfo.position.start.column,
+					endLine: undefined,
+					endColumn: undefined,
+					fix: undefined,
+					suggestions: undefined,
+				};
+			}
+
+			// If we can't find the comment info, fall back to the block start
+			return {
+				...message,
+				line: blockStart,
+				column: block.position.start.column,
+				endLine: undefined,
+				endColumn: undefined,
+				fix: undefined,
+				suggestions: undefined,
+			};
+		}
+
+		if (lineInCode >= block.rangeMap.length) {
 			return null;
 		}
 
--- a/src/processor.js
+++ b/src/processor.js
@@ -16,11 +16,10 @@ import { fromMarkdown } from "mdast-util-from-markdown";
 /**
  * @import { Node, Parent, Code, Html } from "mdast";
  * @import { Linter, Rule, AST } from "eslint";
- * @import { Block, RangeMap } from "./types.js";
+ * @import { Block, MappedCommentLocation, RangeMap } from "./types.js";
  * @typedef {Linter.LintMessage} Message
  * @typedef {Rule.Fix} Fix
  * @typedef {AST.Range} Range
- * @typedef {{ text: string, position: { start: { line: number, column: number }, end?: { line: number, column: number } } }} CommentInfo
  */
 
 //-----------------------------------------------------------------------------
@@ -40,8 +39,8 @@ const SUPPORTS_AUTOFIX = true;
  * @returns {string} A JS block comment string.
  */
 function wrapComment(text) {
-	const prefix = /^[\s*]/u.test(text) ? "" : " ";
-	const suffix = /[\s*]$/u.test(text) ? "" : " ";
+	const prefix = /^\s/u.test(text) ? "" : " ";
+	const suffix = /\s$/u.test(text) ? "" : " ";
 
 	return `/*${prefix}${text}${suffix}*/`;
 }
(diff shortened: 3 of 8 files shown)


Ari4ka (approved):

andreahlert:
Thanks for the review, @Ari4ka!

Ari4ka (approved):

andreahlert:
> 
Hey @nzakas . Everybody says that to me 😅 I write lots of docs at work so it's a habit to be super detailed. And I usually generate an AI template of PR (for headers for exemple), to keep the same looking, but its all my writing.

I'll keep PR descriptions shorter going forward. Lmk if you want me to clarify anything about the implementation!

docs/processors/markdown.md
```diff
@@ -180,6 +180,8 @@ Comment bodies are passed through unmodified, so the plugin supports any [config
 
 This example enables the `alert` global variable, disables the `no-alert` rule, and configures the `quotes` rule to prefer single quotes:
 
+<!-- eslint-skip -->
```
DMartens, line 183: I think we should rather disable reporting unused disable directives by setting `linterSettings: { reportUnusedDisableDirectives: false } }` (and disable @eslint-community/eslint-comments-disable-enable-pair) in the override with the name "markdown/code-blocks/js".

src/processor.js
```diff
@@ -300,13 +314,17 @@ function preprocess(sourceText, filename) {
 				/** @type {string[]} */
 				const comments = [];
 
-				for (const comment of htmlComments) {
-					if (comment.trim() === "eslint-skip") {
+				/** @type {CommentInfo[]} */
```
DMartens: There is no longer a need for the variable `comments` as `commentInfo` has the same information in the `text` property.

src/processor.js
```diff
@@ -392,10 +417,33 @@ function adjustBlock(block) {
...
+
+	if (block.commentInfos) {
+		let lineNumber = 1;
+
+		for (const commentInfo of block.commentInfos) {
+			// Each comment becomes `/*...*/\n` in the linted code
+			const commentText = wrapComment(commentInfo.text);
+			const commentLineCount = commentText.split("\n").length;
+
+			for (let i = 0; i < commentLineCount; i++) {
```
DMartens, line 434: Question to the team: Do we want to support multiline ignore directives?

src/processor.js
```diff
@@ -408,7 +456,39 @@ function adjustBlock(block) {
...
+		 * configuration comments), map it back to the original HTML comment
+		 * location in the Markdown.
+		 */
+		if (lineInCode < 1) {
+			const commentInfo = commentLineMap.get(message.line);
+
+			if (commentInfo) {
+				return {
+					...message,
+					line: commentInfo.position.start.line,
```
DMartens, line 469: Question to the team: Is it okay that the precise location is lost?
For example when a disable directive has multiple rules but only one is unused, the location spans only the name of the unused rule instead of the whole directive.

DMartens:
Thank you for submitting a PR for this. It is a really good starting for fixing this.

Ari4ka (approved):

src/processor.js
nzakas: I'm having a hard time parsing your question. Are you saying that previously we spanned only the unused rule name and this change loses that? If so, I'd say we don't want to lose that.

Or are you saying we didn't span only the unused rule name and this PR does? Then I'd say this is the preferred approach.

docs/processors/markdown.md
andreahlert: I’d keep the doc as-is for this PR. The goal here is to fix #115 so that --report-unused-disable-directives actually works with the plugin instead of dropping those warnings. Disabling the feature in the default config would avoid the bug but wouldn’t fix it. If the team prefers to turn it off by default in the markdown override, that could be a separate follow-up. WDYT?

src/processor.js
andreahlert: Good catch. I’ll remove the comments array and use commentInfo (and its text property) only. I’ll push that change.

src/processor.js
andreahlert: I added multiline support so that when a directive is split across lines in the HTML comment, the mapping still points to the right place in the Markdown. If the team prefers to only support single-line directives for now, we can simplify and drop the multiline handling.

src/processor.js
andreahlert: To clarify for @nzakas: before this PR those messages were dropped (no report at all). With this PR we report them but the position is the whole HTML comment (start line of the comment in the Markdown), not the exact column of the unused rule name inside the comment. So we’re not losing the previous “span only the unused rule name” behavior; we’re adding reporting with a coarser location. 

If the team wants column-level precision (e.g. only the unused rule name) we’d need extra mapping from the linted code columns back to the original comment; 

I can explore that in a follow-up if needed.

Ari4ka (approved):

Ari4ka (approved):

docs/processors/markdown.md
DMartens: I would not expect the `eslint-skip` to be tests.
Rather this behaviour should be tested in a test case.

src/processor.js
DMartens: I checked the behavior for JavaScript directives and they can be multiline, I just never saw this in the wild.

src/processor.js
DMartens: I agree we should postpone this to a follow-up.

src/types.ts
```diff
@@ -68,9 +68,18 @@ export interface RangeMap {
...
+	text: string;
+	position: {
+		start: { line: number; column: number };
+		end?: { line: number; column: number };
+	};
+}
+
 export interface BlockBase {
 	baseIndentText: string;
 	comments: string[];
```
DMartens, line 73: The `comments` property should be removed as it is no longer set.

DMartens:
Sorry for the delay. Other than the small things I mentioned, the changes LGTM.

src/types.ts
nzakas: Can you double-check this? It seems like `comments` is still used.

nzakas (changes requested):
Overall this looks good to me. I'd like to rename `CommentInfo` to something that better describes what it is. `CommentWithLoc`? `MappedCommentLocation`? Just something that makes it clear what it's doing.

lumirlumir (changes requested):
Could you rebase or merge from the main branch once? While reviewing the codebase, I noticed this PR is a bit behind HEAD, so I'm concerned some things may be out of sync when it's merged into main.

Also, it seems there are CI failures that need to be addressed before merging.

src/processor.js
```diff
@@ -20,6 +20,7 @@ import { fromMarkdown } from "mdast-util-from-markdown";
  * @typedef {Linter.LintMessage} Message
  * @typedef {Rule.Fix} Fix
  * @typedef {AST.Range} Range
+ * @typedef {{ text: string, position: { start: { line: number, column: number }, end?: { line: number, column: number } } }} CommentInfo
```
lumirlumir:
```suggestion
```

I think we can import the `CommentInfo` type from the `.types.js` path. For example:

```diff
+  * @import { Block, CommentInfo, RangeMap } from "./types.js";
-  * @import { Block, RangeMap } from "./types.js";
```

tests/plugin.test.js
lumirlumir, line 1: Can we add the same test suites under the `LegacyESLint` describe block?

The processor test suites are usually run against both the Legacy Config and the Flat Config, and `reportUnusedDisableDirectives` was introduced in [ESLint v8.56.0](https://eslint.org/blog/2023/12/eslint-v8.56.0-released/), so I think it needs to be tested on both sides.

eslint.config-content.js
```diff
@@ -3,7 +3,7 @@ import markdown from "./src/index.js";
 
 export default defineConfig([
 	globalIgnores(
-		["**/*.js", "**/.cjs", "**/.mjs"],
+		["**/*.js", "**/.cjs", "**/.mjs", "**/tests/fixtures/**"],
```
lumirlumir, line 6:
```suggestion
		["**/*.js", "**/.cjs", "**/.mjs", "tests/fixtures/"],
```

Just to be more specific and to follow the existing convention.

https://github.com/eslint/markdown/blob/3a8169cb3f96088d7a08db40669d84219e2ab024/eslint.config.js#L27

src/types.ts
lumirlumir: It seems the current implementation still uses the `comments` property. Otherwise it throws a type error.

tests/plugin.test.js
```diff
@@ -2402,4 +2402,204 @@ describe("FlatESLint", () => {
...
+						processor: "markdown/markdown",
+					},
+					{
+						files: ["**/*.js"],
+						rules: {
+							"no-console": "error",
+						},
+					},
+				],
+			});
```
lumirlumir: Could we combine the initialization of `ESLint` and the declaration of this configuration across the four newly added test suites to match the existing convention? I think moving this logic up one level and using `beforeEach` makes the test's intent clearer.

https://github.com/eslint/markdown/blob/3a8169cb3f96088d7a08db40669d84219e2ab024/tests/plugin.test.js#L2322-L2341

Also, I've found that the following existing configuration works for all newly added test cases:

```js
                [
					{
						plugins: {
							markdown: plugin,
						},
						linterOptions: {
							reportUnusedDisableDirectives: "error",
						},
					},
					{
						files: ["*.md"],
						processor: "markdown/markdown",
					},
					{
						files: ["**/*.js"],
						rules: {
							"no-console": "error",
						},
					},
				],
```

src/processor.js
```diff
@@ -32,6 +33,19 @@ const UNSATISFIABLE_RULES = new Set([
...
 
+/**
+ * Wraps comment text in a JS block comment with normalized spacing so that
+ * rules like `spaced-comment` are not triggered by the generated comments.
+ * @param {string} text The comment body extracted from the HTML comment.
+ * @returns {string} A JS block comment string.
+ */
+function wrapComment(text) {
+	const prefix = /^[\s*]/u.test(text) ? "" : " ";
+	const suffix = /[\s*]$/u.test(text) ? "" : " ";
```
lumirlumir:
```suggestion
	const prefix = /^\s/u.test(text) ? "" : " ";
	const suffix = /\s$/u.test(text) ? "" : " ";
```

Is there an intention to use `*` inside the character set (`[]`) syntax? I found that the newly added test cases still pass even when I use the regex above, so I'm wondering about this. 

Also, `wrapComment` helper is only used by `commentInfo.text`, and `commentInfo.text` is retrieved from HTML node. However, the `*` character has no special meaning in HTML comments, so I'm having trouble understanding its intended use.

lumirlumir (changes requested):

src/processor.js
andreahlert: Done. Importing `MappedCommentLocation` (renamed from `CommentInfo`) from `./types.js` now.

src/processor.js
andreahlert: Fixed. Removed the `*` from the character sets, no special meaning in HTML comments.

tests/plugin.test.js
andreahlert: Added. All four test cases are now mirrored in the `LegacyESLint` describe block.

tests/plugin.test.js
andreahlert: Refactored. Using `beforeEach` with the shared config you suggested across both FlatESLint and LegacyESLint blocks.

src/types.ts
andreahlert: Confirmed, `comments` is still used in `getBlockRangeMap`, `block.comments.join` for assembling linted text, and `leadingCommentLines` calculation. Kept it as-is.

docs/processors/markdown.md
andreahlert: That `eslint-skip` was already there before this PR, it prevents the doc example from being linted by the repo's own ESLint config. Not something introduced by this change.

Ari4ka (approved):

nzakas (dismissed):
LGTM. Thanks!

Leaving open for @lumirlumir to review.

lumirlumir (changes requested):
The test in https://github.com/eslint/markdown/blob/main/tests/types/types.test.ts needs to be updated to fix the CI failure. Running `npm run fmt` will also resolve the formatting error.

Ari4ka (approved):

src/types.ts
DMartens: My point was that `comments` is unnecessary as the relevant data is available in `commentInfos`. Further on line 430 the code also recreates the data from `comments` with `wrapComment(commentInfo.text)` which would be available on the same index on `comments`.

fasttime:
This PR needs a new review.

Ari4ka (approved):

src/types.ts
lumirlumir: Could you also take a look at the comments here when you have a moment?

lumirlumir (commented):
Apologies for the delay.

This PR needs some dedicated time for me to review. I’ll take a thorough look within 2-3 days. I appreciate your patience.

eslint.config-content.js
lumirlumir: I think this comment still hasn't been addressed.

src/types.ts
```diff
@@ -60,9 +60,18 @@ export interface RangeMap {
 	md: number;
 }
 
+export interface MappedCommentLocation {
+	text: string;
+	position: {
+		start: { line: number; column: number };
+		end?: { line: number; column: number };
+	};
```
lumirlumir, line 68:
```suggestion
	position: Position;
```

The current `position` information comes from `node.position`, which uses the `Position` type from `unist`.

```js
				htmlComments.push({
					text: comment,
					position: node.position,
				});
```

https://github.com/andreahlert/markdown/blob/dca9351e298b2dd18c897774f69776e49442b148/src/processor.js#L355

However, the current position information is hard-coded and not derived from the existing type.

Could we add at the top level `import type { Position } from "unist";` and reuse that type?

The current type testing in `typest.test.ts` relies on this type information, so the type tests will need to be adjusted accordingly.

src/types.ts
```diff
@@ -60,9 +60,18 @@ export interface RangeMap {
...
+}
+
 export interface BlockBase {
 	baseIndentText: string;
 	comments: string[];
+	commentInfos: MappedCommentLocation[];
```
lumirlumir, line 74: As noted in https://github.com/eslint/markdown/pull/618#discussion_r2989247738, I also think that the `commentInfos` type can be removed and consolidated into the `comments` type to minimize changes.

For example, if we remove `MappedCommentLocation` type and simply create a `Comment` type as follows:

```ts
export interface Comment {
    text: string;
    position: Position;
}
```

And update `BlockBase` as follows:

```diff
export interface BlockBase {
    // ...
-   comments: string[];
+   comments: Comment[];
    // ...
}

File: src/processor.js [hunk 12]
```diff
@@ -392,10 +416,33 @@ function adjustBlock(block) {
 
 	const blockStart = block.position.start.line;
 
+	/*
+	 * Build a mapping from line numbers in the linted code (1-indexed) to the
+	 * original HTML comment info. This allows us to map messages that point to
+	 * configuration comments back to their original location in the Markdown.
+	 */
+	const commentLineMap = new Map();
```
lumirlumir, line 424:
```suggestion
    /** @type {...} */
	const commentLineMap = new Map();
```

Could we specify its type using JSDoc syntax? Currently it's declared as `const commentLineMap: Map<any, any>`, so the type information is lost when other code consumes it.

src/processor.js
lumirlumir: From what I can tell, the current `commentInfo` type is derived from the Markdown HTML node's `position` field, which contains full position information including `start.line`, `start.column`, `end.line`, and `end.column`.

Also, the comment is sanitized by the `getComment` helper before being pushed into the `htmlComments` array:

https://github.com/andreahlert/markdown/blob/6855867919761bc1eb57df1c1bbfb5fb4a307d47/src/processor.js#L70-L89

Therefore, the comment syntax should be valid and retain full position information (verified via logging).

Losing the position information causes the following issue: only the first character is highlighted instead of the entire line.

- Current

Image

- Expected

Image

I believe retaining the full position information here is necessary.

lumirlumir (changes requested):

lumirlumir:
Could you take a look at the merge conflict as well?

PR #624, opened 2026-02-23
feat: fix incorrect regex pattern in `no-multiple-h1`

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

CommonMark and GFM.

### What did you do?

I expected the `# Another H1` heading in the following example to be reported.

```md
<!-- eslint markdown/no-multiple-h1: "error" -->

<h1>Heading</h1       >

# Another H1
```

According to the HTML specification, I expected the following Markdown code to be reported, but it isn't. It's a false negative because the `h1TagPattern` regex does not recognize tags that contain whitespace (any amount) immediately after the tag name, as shown below:

```md
<!-- eslint markdown/no-multiple-h1: "error" -->

<h1>Heading</h1       >

# Another H1
```

https://html.spec.whatwg.org/multipage/syntax.html#end-tags

image

The following example would help identify the problem:

It should report an error for `# Another H1`, but it doesn't.

```md
<!-- eslint markdown/no-multiple-h1: "error" -->

<h1>Heading</h1       >

# Another H1
```

This PR fixes an incorrect regex pattern in `no-multiple-h1`.

Although this is a fix PR, per our [Semantic Versioning Policy](https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy)'s clause "A bug fix in a rule that results in ESLint reporting more linting errors", I've marked the PR prefix as `feat`.

## Related Issues

N/A

N/A

### DIFF
--- a/tests/rules/no-multiple-h1.test.js
+++ b/tests/rules/no-multiple-h1.test.js
@@ -345,12 +345,38 @@ ruleTester.run("no-multiple-h1", rule, {
 			},
 		},
 		'<h1 class="title">Heading</h1>',
+		// This heading tag is invalid because there is a whitespace between `<` and `h1`.
+		// HTML Spec: https://html.spec.whatwg.org/multipage/syntax.html#start-tags
+		dedent`
+			< h1>Heading</h1>
+
+            -^---------------
+
+			# Another H1
+		`,
+		// This heading tag is invalid because there are whitespaces between `<` and `h1`.
+		// HTML Spec: https://html.spec.whatwg.org/multipage/syntax.html#start-tags
+		dedent`
+			<  h1>Heading</h1>
+
+            -^^---------------
+
+			# Another H1
+		`,
+		// This heading tag is invalid because there is a newline whitespace between `<` and `h1`.
+		// HTML Spec: https://html.spec.whatwg.org/multipage/syntax.html#start-tags
+		dedent`
+			<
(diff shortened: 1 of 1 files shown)


tests/rules/no-multiple-h1.test.js
lumirlumir, line 1: Whitespace removal in the test cases was applied automatically by `.editorconfig`, and I've verified that this change has no side effects.

tests/rules/no-multiple-h1.test.js
```diff
@@ -345,6 +345,50 @@ ruleTester.run("no-multiple-h1", rule, {
 			},
 		},
 		'<h1 class="title">Heading</h1>',
+		// This heading tag is invalid because there is a whitespace between `</` and `h1`.
+		// HTML Spec: https://html.spec.whatwg.org/multipage/syntax.html#end-tags
+		dedent`
+			<h1>Heading</ h1>
```
snitin315, line 377: Can you add test case with whitespaces in start tag like `<h1   >`

lumirlumir: Sure, I've updated it in https://github.com/eslint/markdown/pull/624/changes/dd3359a4546f235341af4b736a69200d6ad5c8ae.

I've also updated the `----^-` marker pattern to be slightly more precise.

PR #626, opened 2026-02-27
docs(no-reference-like-urls): fix typo

ShayanTheNerd:
#### Prerequisites checklist

#### AI acknowledgment

Documentation improvement

I clarified the example to specify that the URL should match the reference identifier instead of being a direct link.


lumirlumir (commented):
Hi @ShayanTheNerd, thanks for the PR.

However, I believe the previous example is correct, since the `no-reference-like-urls` rule detects error-prone syntax and reports the incorrect use of `()` instead of `[]`.

Because `mercury` is the definition identifier, someone who wants to use a link reference referring to `mercury` should use the `[]` syntax rather than `()`. So this rule treats `()` as a mistake and suggests using `[]` instead.

That said, if you'd like to make what I mentioned above clearer in the documentation, that would be welcome.

PR #652, opened 2026-05-05
fix: allow `InlineConfigComment` node to be reported

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

CommonMark and GFM.

### What did you do?

When writing rules, I expected nodes from `sourceCode#getInlineConfigNodes` to be reportable by `context.report`, but it shows a type error like in the screenshot below:

image

This is a valid violation report, and ESLint shows the correct positional information when using this method, but only a TypeScript type error is reported.

No type error occurs.

1. Create the following file at `src/rules/test.js` in `@eslint/markdown` repository.

```js
/**
 * @fileoverview Test
 */

// Type Definitions

/**
 * @typedef {'message'} MessageIds
 * @typedef {[]} Options
 * @typedef {MarkdownRuleDefinition<{ RuleOptions: Options, MessageIds: MessageIds }>} RuleDefinition
 */

// Rule Definition

/** @type {RuleDefinition} */
export default {
	meta: {
		type: "problem",

		docs: {
			recommended: true,
			description: "Enforce test",
			url: "https://github.com/eslint/markdown/blob/main/docs/rules/test.md",
		},

		messages: {
			message: "Test Test Test.",
		},
	},

	create(context) {
		const inlineConfigNodes = context.sourceCode.getInlineConfigNodes();

		return {
			"root:exit"() {
				for (const node of inlineConfigNodes) {
					context.report({
						node,
						messageId: "message",
					});
				}
			},
		};
	},
};
```

2. Now, TypeScript reports a type incompatibility for `node`, as shown below:

```txt
Property 'type' is missing in type 'InlineConfigComment' but required in type 'Node'.
```

This PR fixes a similar issue that also affected the JSON language in https://github.com/eslint/json/pull/210.

The root cause is that the `Node` type from `mdast` has a `type` property, but the `InlineConfigComment` class doesn't declare a `type` property, and this discrepancy causes the type error.

For example, in rough form:

```ts
// Node
interface Node {
    type: string;
    position?: Position | undefined;
    data?: Data | undefined;
}

// InlineConfigComment
interface InlineConfigComment {
    value: string;
    position: Position;
}
```

I've updated the `MarkdownRuleDefinition` type to accept `InlineConfigComment` instances as valid violation reports.

## Related Issues

Ref: https://github.com/eslint/json/pull/210

Just a note that the return types of `CSSSourceCode#getInlineConfigNodes` and `JSONSourceCode#getInlineConfigNodes` are already compatible with their respective `CSSRuleDefinition` and `JSONRuleDefinition`.

### DIFF
--- a/src/language/markdown-source-code.js
+++ b/src/language/markdown-source-code.js
@@ -38,7 +38,7 @@ const htmlComment = /<!--(.*?)-->/gsu;
 /**
  * Represents an inline config comment in the source code.
  */
-class InlineConfigComment {
+export class InlineConfigComment {
 	/**
 	 * The comment text.
 	 * @type {string}
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,7 +49,10 @@ import type {
 	CustomRuleTypeDefinitions,
 	CustomRuleVisitorWithExit,
 } from "@eslint/plugin-kit";
-import type { MarkdownSourceCode } from "./index.js";
+import type {
+	InlineConfigComment,
+	MarkdownSourceCode,
+} from "./language/markdown-source-code.js";
 
 //------------------------------------------------------------------------------
 // Exports: Processor
@@ -135,6 +138,11 @@ export interface MarkdownLanguageOptions extends LanguageOptions {
  */
 export type MarkdownLanguageContext = LanguageContext<MarkdownLanguageOptions>;
 
+/**
+ * A Markdown syntax element, including nodes and comments.
+ */
+type MarkdownSyntaxElement = Node | InlineConfigComment;
+
 export interface MarkdownRuleVisitor
 	extends
 		RuleVisitor,
@@ -187,7 +195,7 @@ export type MarkdownRuleDefinition<
 		LangOptions: MarkdownLanguageOptions;
 		Code: MarkdownSourceCode;
 		Visitor: MarkdownRuleVisitor;
-		Node: Node;
+		Node: MarkdownSyntaxElement;
 	},
 	Options
 >;
--- a/tests/language/markdown-source-code.test.js
+++ b/tests/language/markdown-source-code.test.js
@@ -296,9 +296,9 @@ describe("MarkdownSourceCode", () => {
 
 	describe("getInlineConfigNodes()", () => {
 		it("should return `InlineConfigComment` instances", () => {
-			const nodes = sourceCode.getInlineConfigNodes();
+			const inlineConfigNodes = sourceCode.getInlineConfigNodes();
 
-			for (const node of nodes) {
+			for (const node of inlineConfigNodes) {
 				assert.ok(node instanceof InlineConfigComment);
 			}
 		});
--- a/tests/language/markdown-source-code.test.js
+++ b/tests/language/markdown-source-code.test.js
@@ -8,7 +8,10 @@
 //------------------------------------------------------------------------------
 
 import assert from "node:assert";
-import { MarkdownSourceCode } from "../../src/language/markdown-source-code.js";
+import {
+	InlineConfigComment,
+	MarkdownSourceCode,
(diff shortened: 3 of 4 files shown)


src/language/markdown-source-code.js
```diff
@@ -38,7 +38,7 @@ const htmlComment = /<!--(.*?)-->/gsu;
 /**
  * Represents an inline config comment in the source code.
  */
-class InlineConfigComment {
+export class InlineConfigComment {
```
lumirlumir, line 41: Just a note that even if we export this `InlineConfigComment`, `index.js` doesn't re-export the class, it won't affect end users.

src/types.ts
```diff
@@ -135,6 +138,11 @@ export interface MarkdownLanguageOptions extends LanguageOptions {
  */
 export type MarkdownLanguageContext = LanguageContext<MarkdownLanguageOptions>;
 
+/**
+ * A Markdown syntax element, including nodes and comments.
+ */
+type MarkdownSyntaxElement = Node | InlineConfigComment;
```
lumirlumir, line 144:
```suggestion
export type MarkdownSyntaxElement = Node | InlineConfigComment;
```

I'd like `MarkdownSyntaxElement` to be exported for consistency with the JSON and CSS plugins, but,   this would be a feature request, I’d like to raise it to gather consensus on exporting the type.

For example, the JSON plugin exports `JSONSyntaxElement` on the following line:

https://github.com/eslint/json/blob/main/src/types.ts#L43

CSS plugin exports `CSSSyntaxElement` on the following line:

https://github.com/eslint/css/blob/main/src/types.ts#L26

DMartens:
I agree that `InlineConfigureComment`s should be reportable but I think we should accomplish this another way.
Currently the linter only uses the node to get the [`loc`](https://github.com/DMartens/eslint/blob/eac71fb77113de7bf199ff20c6ee44cefcb59848/lib/linter/file-report.js#L556) (before it also set `nodeType`).
As such we should only require the respective property to be present in the shared type [`ViolationReportLocation`](https://github.com/eslint/rewrite/blob/767aa4094acc2d99f3c03f716e626f54e6eb3a83/packages/core/src/types.ts#L490).
I am not feeling strongly about this suggestion as plugin authors still can just pass the location directly via `loc`.

lumirlumir:
I'm not entirely sure of the original intention behind the `node` field, but looking at the [CSS](https://github.com/eslint/css/blob/main/src/types.ts#L26) and [JSON](https://github.com/eslint/json/blob/main/src/types.ts#L43) implementations, they currently seem designed to report only the most common node types for a specific AST.

> Currently the linter only uses the node to get the [loc](https://github.com/DMartens/eslint/blob/eac71fb77113de7bf199ff20c6ee44cefcb59848/lib/linter/file-report.js#L556) (before it also set nodeType).
>
> plugin authors still can just pass the location directly via loc.

I agree that the `loc` property appears to be intended for that purpose, while the `node` property is meant to be more semantically precise during reporting.

https://github.com/eslint/markdown/blob/f9d480236b06a30c3a9833abaf8ea50e1e765938/src/rules/heading-increment.js#L91

> I agree that InlineConfigureComments should be reportable but I think we should accomplish this another way.

I think this could be handled in a separate issue for further discussion if needed, as CSS and JSON already follow the same strategy used in this PR.

DMartens:
The `node` property passed to `context.report` was available in the `LintMessage` but was replaced with `nodeType` because of memory constraints (the whole AST was kept in memory because of the `node` reference).
Later `nodeType` ([end of 2025](https://github.com/eslint/eslint/pull/20096)) was also removed because no integration used it, so now it only is a a way to provide the `loc`.
Yes I agree as this change would affect multiple repositories we should discuss this separately.

DMartens (approved):

PR #664, opened 2026-05-28
feat!: add `meta.languages` to rules and update `recommended` config

### LINKED ISSUE #339: Rule Change: Support `language` and `dialects` to `meta`

Every rule

N/A

Other

### Example code

```markdown
meta: {
		type: "problem",

		docs: {
			description: "Disallow HTML tags",
		},

		messages: {
			disallowedElement: 'HTML element "{{name}}" is not allowed.',
		},

		schema: [
			{
				type: "object",
				properties: {
					allowed: {
						type: "array",
						items: {
							type: "string",
						},
						uniqueItems: true,
					},
				},
				additionalProperties: false,
			},
		],
	},
```

Currently, none of the rules have the `language` or `dialects` fields in their `meta`.

I noticed that all new rules added to ESLint for TypeScript support now include the `language` and `dialects` fields:

https://github.com/eslint/eslint/pull/19562/files#diff-d46d32957ba664c4b406c7587c5e047ab7f719ea98a597e0316837d61be4fc48R150-R151

Like we did for JavaScript, I’d like to suggest adding the `language` and `dialects` fields to `meta` as well.

I think we could add them like this:

```js
meta: {
    // ...
    language: 'markdown',
    dialects: ['commonmark', 'gfm'],
    // ...
}
```

The `language` field should be `'markdown'`, and the `dialects` can be `'commonmark'` and `'gfm'`.

### Participation

### Additional comments

_No response_

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

This PR resolves the issue mentioned in #339 and follows the same strategy used in https://github.com/eslint/json/pull/238 and https://github.com/eslint/css/pull/447.

Unlike the JSON and CSS implementations, Markdown has two rules that only work in GFM mode:

1. `no-bare-urls`
2. `table-column-count`

These rules should be marked as GFM-only, since they are no-ops in CommonMark mode.

Added `meta.languages` and `meta.docs.dialects`, along with test cases.

## Related Issues

Fixes: #339
Refs: https://github.com/eslint/json/pull/238, https://github.com/eslint/css/pull/447

N/A


src/index.js
```diff
@@ -92,7 +93,7 @@ const plugin = {
 			{
 				name: "markdown/recommended",
 				files: ["**/*.md"],
-				language: "markdown/commonmark",
+				language: "markdown/gfm",
```
lumirlumir: Since the `table-column-count` rule only works in GFM mode, using the `recommended` config with `commonmark` causes the following error:

```js
Oops! Something went wrong! :(

ESLint: 10.6.0

The following rules do not support the language "markdown/commonmark":
	- "markdown/table-column-count"

To fix this error, either:
- Remove the rule from your configuration, or set its severity to "off".
- Use the "files" option to apply the rule only to files of the supported language, for example:
  {
    files: ["**/*.js"],
    rules: { "markdown/table-column-count": "error" }
  }

See https://eslint.org/docs/latest/use/configure/rules for more information.
```

---

I think there are two possible approaches to resolve this:

The first is to set the `recommended` config's `language` to `gfm`.

The second is to remove the `table-column-count` rule from the `recommended` config and keep `commonmark` mode.

I'm personally inclined to use `gfm` since GFM syntax is widely used in Markdown documentation, but I'd like to bring this up in case others have a different perspective.

src/index.js
```diff
@@ -92,7 +93,7 @@ const plugin = {
...
 				plugins: (recommendedPlugins = {}),
```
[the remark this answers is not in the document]
lumirlumir: Updated it in https://github.com/eslint/markdown/pull/664/commits/9195f23645d1561a8617207c9ff60979255a45af

src/index.js
fasttime: Since rules currently can't tell which language they're running under (the rule context exposes `languageOptions` but not `language`) this doesn’t seem solvable in the rule itself.

Given that, I wonder if the better option might be to keep `markdown/commonmark` as the default language, since it's the more general subset, and remove rules that aren't compatible with CommonMark from the recommended config. Interested to hear what others think. @eslint/eslint-team

DMartens: I agree that the default `recommended` config should stay with `commonmark` mode.
As such I think we should introduce `recommended-gfm` which sets the language to `markdown/gfm` and includes `table-column-count`.
For `recommended` we can remove `table-column-count` as it would never report anyway as there are no `table` elements in `commonmark`.

fasttime: Does the [`table`](https://github.com/eslint/markdown/blob/3986839466936d73f529ebfd6db6985a1e8c1e77/src/rules/table-column-count.js#L58) selector used in `table-column-count` only exist in GFM? If so, perhaps we could declare that the rule supports both `"markdown/commonmark"` and `"markdown/gfm"`, but it would only report problems in GFM.

DMartens: Yes in Commonmark mode a table is [interpreted as a `paragraph`](https://explorer.eslint.org/#eslint-explorer=v2.eyJzdGF0ZSI6eyJ0b29sIjoiYXN0IiwiY29kZSI6eyJqYXZhc2NyaXB0IjoiLS0tXG5cbi0tLSIsImpzb24iOiIvKipcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cbiAqXG4gKiBUaGUgdGFicyBhcmU6XG4gKlxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbiAqICAgY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxuICpcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXG4gKiBieSBjbGlja2luZyBcIkpTT05cIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XG4gKiBvcHRpb25zLlxuICpcbiAqIFRoaXMgZXhhbXBsZSBpcyBpbiBKU09OQyBtb2RlLCB3aGljaCBhbGxvd3MgY29tbWVudHMuXG4gKi9cblxue1xuICAgIFwia2V5MVwiOiBbdHJ1ZSwgZmFsc2UsIG51bGxdLFxuICAgIFwia2V5MlwiOiB7XG4gICAgICAgIFwia2V5M1wiOiBbMSwgMiwgXCIzXCIsIDFlMTAsIDFlLTNdXG4gICAgfVxufSIsIm1hcmtkb3duIjoifGF8YnxcbnwtfC18XG58Y3xkfCIsImNzcyI6IkBrZXlmcmFtZXMgdGVzdCB7XG4gIDAvKiBDb21tZW50ICovJSB7fVxufSIsImh0bWwiOiI8IURPQ1RZUEUgaHRtbD5cbjwhLS1cblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cblxuVGhlIHRhYnMgYXJlOlxuXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cbi0tPlxuXG48aHRtbCBsYW5nPVwiZW5cIj5cbiAgICA8aGVhZD5cbiAgICAgICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cbiAgICA8L2hlYWQ-XG4gICAgPGJvZHk-XG4gICAgICAgIDxwPlRleHQ8L3A-XG4gICAgPC9ib2R5PlxuPC9odG1sPiJ9LCJsYW5ndWFnZSI6Im1hcmtkb3duIiwianNPcHRpb25zIjp7InBhcnNlciI6ImVzcHJlZSIsInNvdXJjZVR5cGUiOiJtb2R1bGUiLCJlc1ZlcnNpb24iOiJsYXRlc3QiLCJpc0pTWCI6dHJ1ZX0sImpzb25PcHRpb25zIjp7Impzb25Nb2RlIjoianNvbmMiLCJhbGxvd1RyYWlsaW5nQ29tbWFzIjpmYWxzZX0sImNzc09wdGlvbnMiOnsiY3NzTW9kZSI6ImNzcyIsInRvbGVyYW50IjpmYWxzZX0sIm1hcmtkb3duT3B0aW9ucyI6eyJtYXJrZG93bk1vZGUiOiJjb21tb25tYXJrIiwibWFya2Rvd25Gcm9udG1hdHRlciI6InlhbWwiLCJtYXJrZG93bk1hdGgiOmZhbHNlfSwiaHRtbE9wdGlvbnMiOnsidGVtcGxhdGVFbmdpbmVTeW50YXgiOiJub25lIiwiZnJvbnRtYXR0ZXIiOmZhbHNlfSwid3JhcCI6dHJ1ZSwidmlld01vZGVzIjp7ImFzdFZpZXciOiJqc29uIiwic2NvcGVWaWV3IjoiZmxhdCIsInBhdGhWaWV3IjoiZ3JhcGgifSwicGF0aEluZGV4Ijp7ImluZGV4IjowLCJpbmRleGVzIjoxfSwiZXNxdWVyeVNlbGVjdG9yIjp7InNlbGVjdG9yIjoiIn19LCJ2ZXJzaW9uIjowfQ), so the visitor would never be called.

lumirlumir: While the `table-column-count` rule can run in `commonmark` mode, it’s always a no-op, so I’m not sure adding `"markdown/commonmark"` for this rule is the right direction.

If using `gfm` mode by default isn’t preferred, I’d be in favor of removing the `table-column-count` rule from the `recommended` config, since users who used the default `recommended` config before this change wouldn’t get any benefit from `table-column-count`—the rule runs, but it can never report an error.

fasttime: I'm fine with removing `table-column-count` from the recommended rules, let's do it.

lumirlumir: Thanks everyone for the suggestions. I’ve updated it in https://github.com/eslint/markdown/pull/664/changes/b3fb755cfb7295b86739f5f2b8c562e294e3620c.

Just as a note: I removed `recommended: true` entirely instead of changing it to `recommended: false`, to follow the existing rule convention in Markdown.

fasttime (approved):
LGTM, thanks! Since this is a breaking change, it will need another approval by a TSC member before merging.

DMartens (approved):

mdjermanovic (approved):
LGTM, thanks!

PR #669, opened 2026-06-02
fix!: remap locations for unused disable directives

### LINKED ISSUE #115: This does not work with `eslint --report-unused-disable-directives`

xbinaryx:
#### Prerequisites checklist

#### AI acknowledgment

This PR enables the reporting of unused directive comments (such as `<!-- eslint-disable no-console -->`) in Markdown files. When `reportUnusedDisableDirectives` is enabled, ESLint flags these configuration comments if they do not actively suppress any rules. Because the Markdown processor prepends these comments to the generated code, they fall outside the block boundaries. Previously, the processor returned `null` for these messages, causing ESLint to skip them. This PR intercepts unused directive reports and correctly maps them back to their absolute line, column, and autofix offsets in the Markdown source.

- Changed `comments` from an array of strings to objects that track original source location `position`.
- Updated the processor to identify unused directive messages and intercept them for remapping.
- Added test coverage validating correct remapping and autofix offsets across single and multiple comment scenarios.

#### Related Issues

Fixes #115

## Summary by CodeRabbit

* **New Features**
  * Added support for reporting unused ESLint directives in Markdown HTML comments.
  * Added autofixes to remove unused directives or individual rules from single-line and multiline comments.
  * Diagnostics and fixes now map back to the original Markdown comment locations.

* **Documentation**
  * Updated configuration examples to exclude selected code blocks from ESLint checks.

* **Tests**
  * Added coverage for legacy and flat ESLint configurations, multiline comments, multiple directives, and paired enable comments.


src/processor.js
```diff
@@ -240,6 +242,74 @@ function getBlockRangeMap(text, node, comments) {
...
+}
+
+/**
+ * Finds the inserted comment that contains the given generated line.
+ * @param {Block} block The code block to search.
+ * @param {number} line The generated JS line number.
+ * @returns {Comment | undefined} The matching inserted comment location.
+ */
+function findCommentLocation(block, line) {
+	let currentJsLine = 1;
```
DMartens: This name is misleading as the code block does not need to contain JavaScript.

src/processor.js
```diff
@@ -240,6 +242,74 @@ function getBlockRangeMap(text, node, comments) {
...
+	let foundComment;
+
+	for (const comment of block.comments) {
+		const commentLines = comment.text.split("\n").length;
```
DMartens: Can you please add a test case for multiline comments, as `commentLines` is currently only `1`.

src/processor.js
```diff
@@ -240,6 +242,74 @@ function getBlockRangeMap(text, node, comments) {
...
+
+	const adjustedMessage = /** @type {Message} */ ({
+		...messageWithoutFix,
+		line: start.line,
+		column: start.column,
+	});
+
+	if (fix) {
+		adjustedMessage.fix = {
+			range: [start.offset, end.offset],
```
DMartens: This would be wrong in the case that one rule is used and one is unused, e.g.
`eslint-disable used, unused`, as it would delete the whole comment instead of only the respective rule.

tests/plugin.test.js
```diff
@@ -1569,6 +1569,73 @@ describe("FlatESLint", () => {
...
+					assert.strictEqual(results.length, 1);
+					assert.strictEqual(results[0].messages.length, 1);
+					assert.strictEqual(
+						results[0].messages[0].message,
+						"Unused eslint-disable directive (no problems were reported from 'no-console').",
+					);
+					assert.strictEqual(results[0].messages[0].line, 4);
+					assert.strictEqual(results[0].messages[0].column, 1);
+				});
+			});
```
DMartens, line 2142: Can you please also add a test case for `eslint-disable-line` or `eslint-disable-next-line` and `eslint-enable`?

tests/plugin.test.js
lumirlumir, line 1: Like in the previous implementation in https://github.com/eslint/markdown/pull/618/changes#diff-a49973ae7d0aaf6a05e9afce27257be7fdcd99f22e347485870b7d159cb239fd, I think this test needs to be tested against the legacy config, as the [`--report-unused-disable-directives`](https://eslint.org/docs/latest/use/command-line-interface#--report-unused-disable-directives) flag was added in [ESLint v8.52.0](https://eslint.org/blog/2023/10/eslint-v8.52.0-released/#main), and this Markdown plugin still supports legacy configs.

lumirlumir (changes requested):

src/processor.js
xbinaryx: Good catch 👍

tests/plugin.test.js
```diff
@@ -1569,6 +1840,285 @@ describe("FlatESLint", () => {
 				);
 				assert.strictEqual(results[0].messages[0].line, 5);
 			});
+
+			describe("unused disable directives", () => {
```
DMartens, line 1855: All the test cases seem to be repeated with `const` substituted with `var`?

src/processor.js
```diff
@@ -240,6 +242,80 @@ function getBlockRangeMap(text, node, comments) {
...
+			fix.range[0] <= jsOffset &&
+			fix.range[1] >= jsOffset + foundComment.text.length;
+
+		if (isFullRemoval) {
+			adjustedMessage.fix = {
+				range: [start.offset, end.offset],
+				text: fix.text,
+			};
+		} else {
+			const offsetDelta = start.offset + 4 - (jsOffset + 2);
```
DMartens, line 293: Can you please document the magic numbers 4 and 2?
I think 4 for the length of "<!--" and 2 for the length of // or /*?

xbinaryx: Yes, exactly. I've added a comment in the latest commit.

tests/plugin.test.js
xbinaryx: They are nearly identical, but the first block uses `initLegacyESLint` for the old system, and the second uses `initFlatESLint` for the new flat config.

DMartens (dismissed):
Changes LGTM, thanks. Leaving open for lumir.

lumirlumir:
Could you take a look at the merge conflict?

lumirlumir:
I’m sorry for the delay on my end. I’ll be sure to review it by the end of this weekend.

src/types.ts
```diff
@@ -64,9 +65,14 @@ export interface RangeMap {
...
 
+export interface Comment {
+	text: string;
+	position: Position;
+}
+
 export interface BlockBase {
 	baseIndentText: string;
-	comments: string[];
+	comments: Comment[];
```
lumirlumir:
```suggestion
export interface BlockBase {
    baseIndentText: string;
	comments: {
		text: string;
		position: Position;
	}[];
```

I’d say let’s avoid exporting a new type interface and just inline the `Comment` type here.

All the types in `types.ts` are exported through the main entry point, `index.js`, so exporting a new type is effectively a feature. I don’t think the `Comment` type is general enough to be useful for other rule authors as a public type.

For `src/processor.js`, which imports `Comment`, I think we can use `Block['comments'][number]` instead. Or, to keep it simple, we could declare `@typedef { Block['comments'][number] } Comment` in the “Type Definitions” section, which seems cleaner.

tests/types/types.test.ts
```diff
@@ -72,7 +81,9 @@ validBlock.data satisfies CodeData | undefined;
 
 // Verify `Block` has `BlockBase` properties
 validBlock.baseIndentText satisfies string;
-validBlock.comments satisfies string[];
+validBlock.comments satisfies Comment[];
```
lumirlumir:
```suggestion

```

The type test also needs to be adjusted based on the previous comment.

docs/processors/markdown.md
lumirlumir, line 1: Could you also rebase onto or merge the main branch once? A Prettier commit landed in the meantime, so there could be formatting errors when this PR is merged.

https://github.com/eslint/markdown/pull/675

src/processor.js
```diff
@@ -237,6 +239,81 @@ function getBlockRangeMap(text, node, comments) {
...
+	);
+}
+
+/**
+ * Adjusts an unused directive message in an inserted JS comment.
+ * @param {Block} block The code block containing the inserted comment.
+ * @param {LintMessage} message The message to adjust.
+ * @returns {LintMessage} The adjusted message, if it can be mapped.
+ */
+function adjustCommentMessage(block, message) {
```
lumirlumir: `adjustCommentMessage()` currently scans `block.comments` and recomputes
comment line counts/JS offsets for each unused directive message.

If $n$ is the number of leading directive comments and $m$ is the number of
unused directive messages, this is $O(n * m)$ in the worst case. Since $m$ can
grow with $n$ when many consecutive directive comments are unused, this can
effectively become $O(n²)$.

Can we precompute the comment line ranges and JS offsets once per block in
`adjustBlock()` and reuse that mapping when remapping comment messages? That
would reduce the remapping work from quadratic to linear, $O(n + m)$, which
should be much faster for files with many consecutive directive comments.

lumirlumir (changes requested):

DMartens (dismissed):

lumirlumir (commented):
Could you take a look at the CI failure? I think it’ll be resolved once we merge `main` or rebase the branch.

lumirlumir:
I’m sorry for the delay on this PR. I’m going through a busy period, but I’ll be sure to review it by next weekend.

lumirlumir (commented):
Since there’s a pending next major release, and the changed `BlockBase` type can break existing types, I’m marking this PR as a breaking change.

src/processor.js
```diff
@@ -237,6 +241,67 @@ function getBlockRangeMap(text, node, comments) {
...
+/**
+ * Adjusts an unused directive message in an inserted JS comment.
+ * @param {LintMessage} message The message to adjust.
+ * @param {CommentMapping[]} commentMappings Precomputed comment line ranges and offsets.
+ * @returns {LintMessage} The adjusted message, if it can be mapped.
+ */
+function adjustCommentMessage(message, commentMappings) {
+	const mapping = commentMappings.find(
+		m => message.line >= m.startLine && message.line < m.endLine,
+	);
```
lumirlumir: It seems that precomputing `commentMappings` does not achieve the requested $O(n + m)$ complexity because each message still calls `commentMappings.find()`, resulting in $O(n × m)$ work in the worst case.

Could we build a `Map<number, CommentMapping>` in `adjustBlock()`, keyed by each generated line covered by a comment, and use `map.get(message.line)` here? This would provide $O(1)$ lookup per message and $O(n + m)$ overall complexity.

src/processor.js
```diff
@@ -237,6 +241,67 @@ function getBlockRangeMap(text, node, comments) {
...
+
+	const { comment: foundComment, jsOffset } = mapping;
+	const { start, end } = foundComment.position;
+	const { fix, ...messageWithoutFix } = message;
+
+	const adjustedMessage = /** @type {LintMessage} */ ({
+		...messageWithoutFix,
+		line: start.line,
+		column: start.column,
+	});
```
lumirlumir, line 279: `foundComment.position` contains the full range of the original HTML comment, but the adjusted message only maps its start position. This leaves range-aware consumers with a point-only diagnostic:

```text
Current:

1 | <!--
    ^
2 | eslint-disable no-console
3 | -->

Expected:

1 | <!--
    ^^^^
2 | eslint-disable no-console
    ^^^^^^^^^^^^^^^^^^^^^^^^^
3 | -->
    ^^^
```

Could we also set `endLine: end.line` and `endColumn: end.column`, and add a multiline test verifying that the entire HTML comment is highlighted?

lumirlumir (changes requested):

src/processor.js
xbinaryx: I’ve applied both of your suggestions.

DMartens (approved):

nzakas:
@lumirlumir once you approve this PR we can do a new release. :tada:

src/processor.js
```diff
@@ -237,6 +241,67 @@ function getBlockRangeMap(text, node, comments) {
...
+			fix.range[1] >= jsOffset + foundComment.text.length;
+
+		if (isFullRemoval) {
+			adjustedMessage.fix = {
+				range: [start.offset, end.offset],
+				text: fix.text,
+			};
+		} else {
+			// '4' is the length of '<!--' and '2' is the length of '/*'.
+			const offsetDelta = start.offset + 4 - (jsOffset + 2);
```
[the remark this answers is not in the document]
lumirlumir: One last suggestion:

Regarding the AI review, the current autofix for nested blockquotes (`>`) and list items produces an incorrect fix in this case. Even after applying the autofix, the error location becomes incorrect.

It would be helpful to fix this bug and add regression tests to prevent it from recurring.

- Before: 

image

- After: 

image

PR #681, opened 2026-07-10
chore: update TypeScript to v7

sethamus:
#### Prerequisites checklist

#### AI acknowledgment

This PR updates TypeScript to v7.

Updated `typescript` to `^7.0.2` and replaced the `7.x` tests with `6.x` tests.

#### Related Issues

https://github.com/eslint/eslint/issues/21070

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,8 +99,8 @@ jobs:
               run: npm run test:types:5.3
             - name: Check Types (TypeScript 5.x)
               run: npm run test:types:5.x
-            - name: Check Types (TypeScript 7 preview)
-              run: npm run test:types:7.x
+            - name: Check Types (TypeScript 6.x)
+              run: npm run test:types:6.x
 
     are_the_types_wrong:
         name: Are the types wrong?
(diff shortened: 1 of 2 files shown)


DMartens:
Can you please checkout the CI problem.
The problem seems to be that the example in `examples/typescript` uses `typescript-eslint` which cannot use TypeScript 7 yet as it has no API.
This can be circumvented by also providing TypeScript 6 as [described in the release note](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0).

lumirlumir:
If there’s no clean solution for this, I think another possible approach would be to mark this PR as blocked, like in the `eslint`, `rewrite`, and `js` repositories, as mentioned in https://github.com/eslint/eslint/issues/21070.

Since the generated `.d.ts` files would be the same with TS 6 and TS 7 (and TS 6 was mainly intended to bridge the transition to TS 7), I don’t think there’s a need to rush this PR.

Also, the type tests still run against the latest `@typescript/native-preview`, which should be very close to `typescript@7.0.2`.

sethamus:
I agree with lumirlumir. There's no need to rush this, so I'll convert this PR to a draft for now.

PR #685, opened 2026-07-17
docs: add eslint to install command

KumJungMin:
#### Prerequisites checklist

#### AI acknowledgment

The `@eslint/markdown` package does not declare `eslint` as a peer dependency. 
As a result, users may assume that installing `@eslint/markdown` will also install ESLint, but ESLint must actually be installed separately in order to use the plugin. 

The existing installation instructions only include `@eslint/markdown`, which may cause new users to miss a required dependency. To prevent this, the documentation now instructs users to install `eslint` before installing `@eslint/markdown`.

Updated the installation instructions for @eslint/markdown to clarify that eslint must be installed separately before installing the plugin, consistent with the change made in eslint/json#258.


nzakas (approved):
LGTM. Thanks.

PR #686, opened 2026-07-17
ci: declare least-privilege workflow-level contents: read

hu6r1s:
#### Prerequisites checklist

#### AI acknowledgment

Explicitly declares contents: read permissions for the affected workflows instead of relying on the default GITHUB_TOKEN permissions.

Explicitly declares permissions: contents: read for workflows that only require read access instead of relying on the default GITHUB_TOKEN permissions.

Since the affected workflows only check out the repository and do not perform any write operations, this change follows the principle of least privilege and improves workflow security by reducing the permissions granted to the GITHUB_TOKEN.

* Added explicit `permissions: contents: read` to the affected GitHub Actions workflows.
* Confirmed that the workflows only require read access to check out the repository and do not perform any write operations.
* Applied the same workflow permission hardening approach as [eslint/.github#20](https://github.com/eslint/.github/pull/20).

### DIFF
--- a/.github/workflows/update-readme.yml
+++ b/.github/workflows/update-readme.yml
@@ -6,6 +6,9 @@ on:
 
     workflow_dispatch:
 
+permissions:
+    contents: read
+
 jobs:
     update-readme:
         uses: eslint/workflows/.github/workflows/update-readme.yml@main


lumirlumir (approved):
LGTM, thanks!

PR #692, opened 2026-07-21
docs: correct allowPattern value in no-missing-link-fragments example

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

- Documentation update

image

Hi! This is a small fix for the `allowPattern` example in the `no-missing-link-fragments` docs.

The heading introduces the example as configured with `{ allowPattern: "" }`, but the code block right below it actually uses `{ allowPattern: "^figure-" }`, so the two don't line up.

`"^figure-"` is the value that makes the example correct: it matches the `#figure-19` fragment so the rule ignores it. With an empty `allowPattern`, that fragment isn't ignored and the rule reports `Link fragment '#figure-19' does not reference a heading or anchor in this document`, so the snippet wouldn't be a "correct" example. Updating the heading to `"^figure-"` matches the code block.

Changed the `allowPattern` value in the heading from `""` to `"^figure-"` so it matches the example below it. One line in `docs/rules/no-missing-link-fragments.md`.

#### Related Issues

None

Nothing in particular. It's a docs-only one-line change.

### DIFF
--- a/docs/rules/no-missing-link-fragments.md
+++ b/docs/rules/no-missing-link-fragments.md
@@ -94,7 +94,7 @@ This rule supports the following options:
 * `allowPattern: string` -
     A regular expression string. If a link fragment matches this pattern, it will be ignored by the rule. This is useful for fragments that are dynamically generated or handled by other tools. (default: `""`)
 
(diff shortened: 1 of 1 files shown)


lumirlumir (approved):
Nice catch, thanks!

PR #693, opened 2026-07-21
fix: prevent ReDoS in no-reversed-media-syntax

### LINKED ISSUE #690: Bug: `no-reversed-media-syntax` hangs on parenthesis-dense prose (catastrophic backtracking)
### Environment

Node version: v26.5.0
npm version: 11.17.0
Local ESLint version: v10.7.0
Global ESLint version: -
Operating System: macOS (darwin 25.5.0)

### What parser are you using?

Default (@eslint/markdown)

### What did you do?

`markdown/no-reversed-media-syntax` exhibits catastrophic backtracking. On prose containing an unclosed `(` followed by a moderate number of parenthesised spans (and no `)[` anywhere), lint time grows exponentially and the run never completes.

```js
// eslint.config.mjs

export default [
  {
    files: ["**/*.md"],
    plugins: { markdown },
    language: "markdown/gfm",
    rules: { "markdown/no-reversed-media-syntax": "error" },
  },
];
```

Generate the fixture:

```sh
node -e "require('fs').writeFileSync('repro.md', 'Text (unclosed ' + 'and (x) '.repeat(30) + 'end\n')"
npx eslint repro.md
```

ESLint completes promptly and reports no `no-reversed-media-syntax` violations (there is no reversed `(label)[url]` syntax in the input).

### What actually happened?

ESLint hangs, pinning a CPU core indefinitely. Measured wall-clock against the repeat count `n`:

| `n` (repeats of `and (x) `) | eslint runtime |
| --------------------------- | -------------- |
| 26                          | 16 s           |
| 28                          | 64 s           |
| 30                          | > 90 s (killed) |
| 32                          | > 90 s (killed) |

Roughly 4× per two additional spans — exponential, not merely slow.

### Analysis

The rule's `reversedPattern` (`src/rules/no-reversed-media-syntax.js`):

```js
const reversedPattern =
  /(?<=(?<!\\)(?:\\{2})*)\((?<label>(?:\\.|[^()\\]|\([\s\S]*\))*)\)\[(?<url>(?:\\.|[^\]\\\r\n])*)\](?!\()/gu;
```

The `label` group is `(?: \\. | [^()\\] | \([\s\S]*\) )*` — an unbounded `[\s\S]*` nested inside an outer `*`, where the third alternative can also span the same region the first two can. When the overall match ultimately fails (no `)[` follows), the engine must explore every way of partitioning the remaining text between the outer quantifier and the inner `[\s\S]*`, which is exponential in the number of `(...)` spans.

This reproduces identically on **8.0.2 and 8.0.3** — the rule source is byte-identical between the two releases.

### Real-world impact

This is not a synthetic-input-only issue. It was hit on an ordinary hand-written `CLAUDE.md`: a single Markdown list item of documentation prose containing ~260 parentheses (nested parentheticals inside a long bullet). The file is otherwise unremarkable, and every other file in the repository lints in milliseconds. Because the rule is in `recommended`, any project with sufficiently dense parenthetical prose will hang with no indication of which rule is responsible — `--debug` stops at `With flat config: <file>` and emits nothing further.

### Possible fix

Make the alternation unambiguous and bound the nested span, e.g. replace `\([\s\S]*\)` with a non-backtracking form such as `\([^()]*\)` (or make the group atomic / rewrite so each character has exactly one parse). Happy to open a PR if a preferred direction is indicated.

### Participation

virajp:
#### Prerequisites checklist

#### AI acknowledgment

Fix the catastrophic backtracking in `no-reversed-media-syntax` reported in #690.

Bounded the nested group in `reversedPattern` from `\([\s\S]*\)` to `\([^()]*\)`, as suggested in review.

The `label` group is `(?: \\. | [^()\\] | \([\s\S]*\) )*`. The third alternative could match text the first two could also match, so when the overall match failed the engine explored every way of splitting the input between the outer `*` and the inner `[\s\S]*`. Bounding the nested group removes that overlap: each character now has exactly one possible parse.

Timings for the reproduction in #690:

| repeats of `and (x) ` | before | after |
| --------------------- | ------ | ----- |
| 26 | 16s | <1ms |
| 28 | 64s | <1ms |
| 30 | >90s (killed) | <1ms |

Growth is now linear — 50,000 repeats parse in 2.4ms, and `"(".repeat(50000)` in 0.7ms.

Added two `valid` cases from the issue as regression tests.

#### Related Issues

Fixes #690

**This narrows what a label can contain, which is a behavior change.** `\([\s\S]*\)` accepted arbitrarily deep nesting; `\([^()]*\)` accepts one level:

```js
"((a))[u]"             // still reported (1 level)
"(x (a) y)[u]"         // still reported (1 level)
"(x (a (b)) y)[u]"     // no longer reported (2 levels)
```

All 93 existing tests pass unchanged, so nothing currently covered regresses. I added `"(x (a (b)) y)[url]"` as a `valid` case to document the new limit — happy to remove it if you would rather not pin that, or to extend the pattern by one more level if two-deep labels are worth supporting.

I verified equivalence against the old pattern by differential testing over 500,000 random inputs built from parentheses, brackets, backslashes, spaces, line breaks and astral characters; the only divergences are the nesting-depth cases above.

### DIFF
--- a/tests/rules/no-reversed-media-syntax.test.js
+++ b/tests/rules/no-reversed-media-syntax.test.js
@@ -110,7 +110,7 @@ ruleTester.run("no-reversed-media-syntax", rule, {
 		},
 		// https://github.com/eslint/markdown/issues/690
 		`Text (unclosed ${"and (x) ".repeat(30)}end`,
-		"(".repeat(10000),
+		"(".repeat(100_000),
 		// A label may contain at most one level of nested parentheses.
 		"(x (a (b)) y)[url]",
 	],
--- a/tests/rules/no-reversed-media-syntax.test.js
+++ b/tests/rules/no-reversed-media-syntax.test.js
@@ -108,6 +108,11 @@ ruleTester.run("no-reversed-media-syntax", rule, {
 				math: true,
 			},
 		},
+		// https://github.com/eslint/markdown/issues/690
+		`Text (unclosed ${"and (x) ".repeat(30)}end`,
+		"(".repeat(10000),
+		// A label may contain at most one level of nested parentheses.
+		"(x (a (b)) y)[url]",
 	],
 	invalid: [
 		{
--- a/docs/rules/no-reversed-media-syntax.md
+++ b/docs/rules/no-reversed-media-syntax.md
@@ -46,6 +46,20 @@ Examples of **correct** code for this rule:
 | [ESLint](https://eslint.org/) | ![A beautiful sunset](sunset.png) |
 ```
 
+## Known Limitations
+
+This rule uses a regular expression to find reversed syntax, and JavaScript regular expressions cannot match arbitrarily nested parentheses. A label may contain at most one level of nested parentheses, so `(ESLint (the linter))[https://eslint.org/]` is reported, while labels nesting two or more levels deep are not.
+
+Examples of reversed syntax that this rule **does not** report:
+
+```markdown
+<!-- eslint markdown/no-reversed-media-syntax: "error" -->
+
+(ESLint (the (JavaScript) linter))[https://eslint.org/]
+
+!(A sunset (over (the) sea))[sunset.png]
+```
+
 ## When Not To Use It
 
 If you don't need to enforce correct link and image syntax, you can safely disable this rule.
--- a/src/rules/no-reversed-media-syntax.js
+++ b/src/rules/no-reversed-media-syntax.js
@@ -20,9 +20,13 @@
 // Helpers
 //-----------------------------------------------------------------------------
 
-/** Matches reversed link/image syntax like `(text)[url]`, ignoring escaped characters like `\(text\)[url]`. */
+/**
+ * Matches reversed link/image syntax like `(text)[url]`, ignoring escaped characters like `\(text\)[url]`.
+ * The nested group is `\([^()]*\)`, not `\([\s\S]*\)`, so that it cannot overlap the
+ * other alternatives. Overlapping alternatives cause catastrophic backtracking.
+ */
 const reversedPattern =
-	/(?<=(?<!\\)(?:\\{2})*)\((?<label>(?:\\.|[^()\\]|\([\s\S]*\))*)\)\[(?<url>(?:\\.|[^\]\\\r\n])*)\](?!\()/gu;
+	/(?<=(?<!\\)(?:\\{2})*)\((?<label>(?:\\.|[^()\\]|\([^()]*\))*)\)\[(?<url>(?:\\.|[^\]\\\r\n])*)\](?!\()/gu;
 
 //-----------------------------------------------------------------------------
 // Rule Definition


lumirlumir (changes requested):
Thanks for the PR, but I think this solution is too verbose and does more than what we really want to fix in this PR. I’d say let’s pinpoint the problematic regex pattern and fix it surgically, as you mentioned in the issue, such as replacing `\([\s\S]*\)` with a non-backtracking form like `\([^()]*\)`, or something similar.

Also, please follow our [AI Usage Policy](https://eslint.org/docs/latest/contribute/ai-policy) and update the PR description to match our template.

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -108,6 +108,11 @@ ruleTester.run("no-reversed-media-syntax", rule, {
 				math: true,
 			},
 		},
+		// https://github.com/eslint/markdown/issues/690
+		`Text (unclosed ${"and (x) ".repeat(30)}end`,
+		"(".repeat(10000),
```
lumirlumir:
```suggestion
		"(".repeat(100_000),
```

This test still passes without exceeding the timeout, even when I revert the change to use the `\s\S` pattern instead. Since `100_000` causes a timeout error in the test, can we use that instead?

tests/rules/no-reversed-media-syntax.test.js
```diff
@@ -108,6 +108,11 @@ ruleTester.run("no-reversed-media-syntax", rule, {
...
+		// A label may contain at most one level of nested parentheses.
+		"(x (a (b)) y)[url]",
```
lumirlumir, line 115: Nice catch on the regression.

As mentioned in the PR description, if there’s a way to keep reporting this pattern, I’d be in favor of continuing to report it.

Since the rule expects that case to be transformed into `[x (a (b)) y](url)` (with the `[` and `(` patterns swapped), and that’s still valid Markdown `Link` syntax.

[Example](https://explorer.eslint.org/#eslint-explorer=v2.eyJzdGF0ZSI6eyJ0b29sIjoiYXN0IiwiY29kZSI6eyJqYXZhc2NyaXB0IjoiY29uc3QgaGkgPSA0MjtcblxuZGVidWdnZXI7XG5cbmRlYnVnZ2VyO1xuXG5FU0xpbnQiLCJqc29uIjoie1xuICBcIm5hbWVcIjogXCJAZXNsaW50L2pzb25cIixcbiAgXCJ2ZXJzaW9uXCI6IFwiMi4wLjFcIixcbiAgXCJkZXNjcmlwdGlvblwiOiBcIkpTT04gbGludGluZyBwbHVnaW4gZm9yIEVTTGludFwiLFxuICBcImF1dGhvclwiOiBcIk5pY2hvbGFzIEMuIFpha2FzXCIsXG4gIFwidHlwZVwiOiBcIm1vZHVsZVwiLFxuICBcIm1haW5cIjogXCJkaXN0L2luZGV4LmpzXCIsXG4gIFwidHlwZXNcIjogXCJkaXN0L2luZGV4LmQudHNcIixcbiAgXCJleHBvcnRzXCI6IHtcbiAgICBcIi5cIjoge1xuICAgICAgXCJ0eXBlc1wiOiBcIi4vZGlzdC9pbmRleC5kLnRzXCIsXG4gICAgICBcImRlZmF1bHRcIjogXCIuL2Rpc3QvaW5kZXguanNcIlxuICAgIH1cbiAgfSxcbiAgXCJmaWxlc1wiOiBbXG4gICAgXCJkaXN0XCJcbiAgXSxcbiAgXCJwdWJsaXNoQ29uZmlnXCI6IHtcbiAgICBcImFjY2Vzc1wiOiBcInB1YmxpY1wiXG4gIH0sXG4gIFwiZ2l0SG9va3NcIjoge1xuICAgIFwicHJlLWNvbW1pdFwiOiBcImxpbnQtc3RhZ2VkXCJcbiAgfSxcbiAgXCJsaW50LXN0YWdlZFwiOiB7XG4gICAgXCIqLmpzXCI6IFtcbiAgICAgIFwiZXNsaW50IC0tZml4XCIsXG4gICAgICBcInByZXR0aWVyIC0td3JpdGVcIlxuICAgIF0sXG4gICAgXCIhKCouanMpXCI6IFwicHJldHRpZXIgLS13cml0ZSAtLWlnbm9yZS11bmtub3duXCIsXG4gICAgXCJ7c3JjL3J1bGVzLyouanMsdG9vbHMvdXBkYXRlLXJ1bGVzLWRvY3MuanMsUkVBRE1FLm1kfVwiOiBbXG4gICAgICBcIm5wbSBydW4gYnVpbGQ6dXBkYXRlLXJ1bGVzLWRvY3NcIixcbiAgICAgIFwiZ2l0IGFkZCBSRUFETUUubWRcIlxuICAgIF1cbiAgfSxcbiAgXCJyZXBvc2l0b3J5XCI6IHtcbiAgICBcInR5cGVcIjogXCJnaXRcIixcbiAgICBcInVybFwiOiBcImdpdCtodHRwczovL2dpdGh1Yi5jb20vZXNsaW50L2pzb24uZ2l0XCJcbiAgfSxcbiAgXCJidWdzXCI6IHtcbiAgICBcInVybFwiOiBcImh0dHBzOi8vZ2l0aHViLmNvbS9lc2xpbnQvanNvbi9pc3N1ZXNcIlxuICB9LFxuICBcImhvbWVwYWdlXCI6IFwiaHR0cHM6Ly9naXRodWIuY29tL2VzbGludC9qc29uI3JlYWRtZVwiLFxuICBcInNjcmlwdHNcIjoge1xuICAgIFwiYnVpbGRcIjogXCJucG0gcnVuIGJ1aWxkOnJ1bGVzICYmIG5wbSBydW4gYnVpbGQ6dHlwZXMgJiYgbnBtIHJ1biBidWlsZDp1cGRhdGUtcnVsZXMtZG9jc1wiLFxuICAgIFwiYnVpbGQ6cnVsZXNcIjogXCJub2RlIHRvb2xzL2J1aWxkLXJ1bGVzLmpzXCIsXG4gICAgXCJidWlsZDp0eXBlc1wiOiBcInRzY1wiLFxuICAgIFwiYnVpbGQ6dXBkYXRlLXJ1bGVzLWRvY3NcIjogXCJub2RlIHRvb2xzL3VwZGF0ZS1ydWxlcy1kb2NzLmpzXCIsXG4gICAgXCJwcmVwYXJlXCI6IFwibnBtIHJ1biBidWlsZFwiLFxuICAgIFwicHJldGVzdFwiOiBcIm5wbSBydW4gYnVpbGRcIixcbiAgICBcImxpbnRcIjogXCJlc2xpbnRcIixcbiAgICBcImxpbnQ6Zml4XCI6IFwiZXNsaW50IC0tZml4XCIsXG4gICAgXCJsaW50OnR5cGVzXCI6IFwiYXR0dyAtLXBhY2sgLS1wcm9maWxlIGVzbS1vbmx5XCIsXG4gICAgXCJsaW50OnVudXNlZFwiOiBcImtuaXBcIixcbiAgICBcImZtdFwiOiBcInByZXR0aWVyIC0td3JpdGUgLlwiLFxuICAgIFwiZm10OmNoZWNrXCI6IFwicHJldHRpZXIgLS1jaGVjayAuXCIsXG4gICAgXCJ0ZXN0XCI6IFwibW9jaGEgXFxcInRlc3RzLyoqLyoudGVzdC5qc1xcXCJcIixcbiAgICBcInRlc3Q6Y292ZXJhZ2VcIjogXCJjOCBucG0gdGVzdFwiLFxuICAgIFwidGVzdDpqc3JcIjogXCJucHggLXkganNyQGxhdGVzdCBwdWJsaXNoIC0tZHJ5LXJ1blwiLFxuICAgIFwidGVzdDp0eXBlc1wiOiBcInRzYyAtcCB0ZXN0cy90eXBlcy90c2NvbmZpZy5qc29uXCIsXG4gICAgXCJ0ZXN0OnR5cGVzOjUuM1wiOiBcIm5weCAtcCB0eXBlc2NyaXB0QDUuMyAteSAtLSB0c2MgLXAgdGVzdHMvdHlwZXMvdHNjb25maWcubGVnYWN5Lmpzb25cIixcbiAgICBcInRlc3Q6dHlwZXM6NS54XCI6IFwibnB4IC1wIHR5cGVzY3JpcHRANS54IC15IC0tIHRzYyAtcCB0ZXN0cy90eXBlcy90c2NvbmZpZy5qc29uXCIsXG4gICAgXCJ0ZXN0OnR5cGVzOjYueFwiOiBcIm5weCAtcCB0eXBlc2NyaXB0QDYueCAteSAtLSB0c2MgLXAgdGVzdHMvdHlwZXMvdHNjb25maWcuanNvblwiLFxuICAgIFwidGVzdDp0eXBlczphbGxcIjogXCJucG0gcnVuIHRlc3Q6dHlwZXMgJiYgbnBtIHJ1biB0ZXN0OnR5cGVzOjUuMyAmJiBucG0gcnVuIHRlc3Q6dHlwZXM6NS54ICYmIG5wbSBydW4gdGVzdDp0eXBlczo2LnhcIlxuICB9LFxuICBcImtleXdvcmRzXCI6IFtcbiAgICBcImVzbGludFwiLFxuICAgIFwiZXNsaW50LXBsdWdpblwiLFxuICAgIFwiZXNsaW50cGx1Z2luXCIsXG4gICAgXCJqc29uXCIsXG4gICAgXCJsaW50aW5nXCJcbiAgXSxcbiAgXCJsaWNlbnNlXCI6IFwiQXBhY2hlLTIuMFwiLFxuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJAZXNsaW50L2NvcmVcIjogXCJeMS4yLjFcIixcbiAgICBcIkBlc2xpbnQvcGx1Z2luLWtpdFwiOiBcIl4wLjcuMlwiLFxuICAgIFwiQGh1bWFud2hvY29kZXMvbW9tb2FcIjogXCJeMy4zLjEwXCIsXG4gICAgXCJuYXR1cmFsLWNvbXBhcmVcIjogXCJeMS40LjBcIlxuICB9LFxuICBcImRldkRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJAYXJldGhldHlwZXN3cm9uZy9jbGlcIjogXCJeMC4xOC4zXCIsXG4gICAgXCJAdHlwZXMvbmF0dXJhbC1jb21wYXJlXCI6IFwiXjEuNC4zXCIsXG4gICAgXCJAdHlwZXMvbm9kZVwiOiBcIl4yMC4xOS4wXCIsXG4gICAgXCJjOFwiOiBcIl4xMS4wLjBcIixcbiAgICBcImRlZGVudFwiOiBcIl4xLjUuM1wiLFxuICAgIFwiZXNsaW50XCI6IFwiXjEwLjQuMFwiLFxuICAgIFwiZXNsaW50LWNvbmZpZy1lc2xpbnRcIjogXCJeMTQuMC4wXCIsXG4gICAgXCJlc2xpbnQtcGx1Z2luLWVzbGludC1wbHVnaW5cIjogXCJeNy4zLjJcIixcbiAgICBcImdsb2JhbHNcIjogXCJeMTcuMC4wXCIsXG4gICAgXCJrbmlwXCI6IFwiXjYuMC4wXCIsXG4gICAgXCJsaW50LXN0YWdlZFwiOiBcIl4xNi4wLjBcIixcbiAgICBcIm1kYXN0LXV0aWwtZnJvbS1tYXJrZG93blwiOiBcIl4yLjAuMlwiLFxuICAgIFwibW9jaGFcIjogXCJeMTEuMy4wXCIsXG4gICAgXCJwcmV0dGllclwiOiBcIjMuOS41XCIsXG4gICAgXCJ0eXBlc2NyaXB0XCI6IFwiXjcuMC4yXCIsXG4gICAgXCJ5b3JraWVcIjogXCJeMi4wLjBcIlxuICB9LFxuICBcImVuZ2luZXNcIjoge1xuICAgIFwibm9kZVwiOiBcIl4yMC4xOS4wIHx8IF4yMi4xMy4wIHx8ID49MjRcIlxuICB9XG59IiwibWFya2Rvd24iOiJbeCAoYSAoYikpIHldKHVybCkiLCJjc3MiOiJAa2V5ZnJhbWVzIHRlc3Qge1xuXHQwJSB7XG5cdFx0b3BhY2l0eTogMDtcblx0fVxuXG4gICAgZnJvbSB7XG5cdFx0b3BhY2l0eTogMDtcbiAgICB9XG5cblx0MTAwJSB7XG5cdFx0b3BhY2l0eTogMTtcblx0fVxufSIsImh0bWwiOiI8IURPQ1RZUEUgaHRtbD5cbjwhLS1cblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cblxuVGhlIHRhYnMgYXJlOlxuXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cbi0tPlxuXG48aHRtbCBsYW5nPVwiZW5cIj5cbiAgICA8aGVhZD5cbiAgICAgICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cbiAgICA8L2hlYWQ-XG4gICAgPGJvZHk-XG4gICAgICAgIDxwPlRleHQ8L3A-XG4gICAgPC9ib2R5PlxuPC9odG1sPiJ9LCJsYW5ndWFnZSI6Im1hcmtkb3duIiwianNPcHRpb25zIjp7InBhcnNlciI6ImVzcHJlZSIsInNvdXJjZVR5cGUiOiJtb2R1bGUiLCJlc1ZlcnNpb24iOiJsYXRlc3QiLCJpc0pTWCI6dHJ1ZX0sImpzb25PcHRpb25zIjp7Impzb25Nb2RlIjoianNvbjUiLCJhbGxvd1RyYWlsaW5nQ29tbWFzIjpmYWxzZX0sImNzc09wdGlvbnMiOnsiY3NzTW9kZSI6ImNzcyIsInRvbGVyYW50IjpmYWxzZX0sIm1hcmtkb3duT3B0aW9ucyI6eyJtYXJrZG93bk1vZGUiOiJjb21tb25tYXJrIiwibWFya2Rvd25Gcm9udG1hdHRlciI6Im9mZiIsIm1hcmtkb3duTWF0aCI6ZmFsc2V9LCJodG1sT3B0aW9ucyI6eyJ0ZW1wbGF0ZUVuZ2luZVN5bnRheCI6Im5vbmUiLCJmcm9udG1hdHRlciI6ZmFsc2V9LCJ3cmFwIjp0cnVlLCJ2aWV3TW9kZXMiOnsiYXN0VmlldyI6InRyZWUiLCJzY29wZVZpZXciOiJmbGF0IiwicGF0aFZpZXciOiJncmFwaCJ9LCJwYXRoSW5kZXgiOnsiaW5kZXgiOjAsImluZGV4ZXMiOjF9LCJlc3F1ZXJ5U2VsZWN0b3IiOnsic2VsZWN0b3IiOiIifX0sInZlcnNpb24iOjB9)

lumirlumir (changes requested):

tests/rules/no-reversed-media-syntax.test.js
virajp: Happy to keep reporting it — but I want to check which trade-off you'd prefer, because JS regex has no recursion, so any pure-pattern fix has *some* fixed nesting bound. The question is only where we put it.

**Option A — move the bound one level deeper.** Nest the group once more:

```js
// label's third alternative: \((?:\\.|[^()\\]|\([^()]*\))*\)
/(?<=(?<!\\)(?:\\{2})*)\((?<label>(?:\\.|[^()\\]|\((?:\\.|[^()\\]|\([^()]*\))*\))*)\)\[(?<url>(?:\\.|[^\]\\\r\n])*)\](?!\()/gu
```

The alternatives stay disjoint, so this is still backtracking-free — `"(".repeat(100_000)` runs in ~1.3ms, the #690 input in <0.01ms, and 20,000 nested `(x)` groups in ~0.6ms.

| input | current PR | Option A |
| --- | --- | --- |
| `(x (a) y)[url]` | reported | reported |
| `(x (a (b)) y)[url]` | not reported | reported |
| `(x (a (b (c))) y)[url]` | not reported | not reported |

So it fixes the case you flagged, but the cliff just moves from depth 2 to depth 3, and each extra level makes the pattern noticeably harder to read.

**Option B — match the label with a balanced-paren scan** instead of encoding nesting in the regex. That handles arbitrary depth in linear time, but it's the more verbose direction you steered me away from in the first review, so I don't want to go there without your say-so.

My read is that depth 3+ in a reversed link label is rare in real Markdown, so Option A buys most of the benefit for a two-token change — but you know the rule's users better than I do. Which would you like? If Option A, I'll push it and turn `"(x (a (b)) y)[url]"` into an `invalid` case, with a `valid` case pinning the new depth-3 limit (or drop that pin entirely if you'd rather
not commit to it).

lumirlumir: Thanks for the detailed summary! I now understand the regex-based limitations clearly.

> My read is that depth 3+ in a reversed link label is rare in real Markdown

As you said, I also agree that depth 3+ is rare in Markdown documents, so I think it’s fine to keep the current solution.

Just to clarify the nested-depth limitation, could we add a small note to https://github.com/eslint/markdown/blob/main/docs/rules/no-reversed-media-syntax.md explaining that 3+ nested depth patterns are not supported, with a small example?

It would be helpful if this section could be placed under `## Options` as `## Known Limitations`, similar to https://eslint.org/docs/latest/rules/no-throw-literal#known-limitations.

virajp: I have updated the doc with what is supported by the commit

lumirlumir (approved):
LGTM, thanks!

Since nesting at a depth of two or greater is quite an edge case, but this change alters the rule’s behavior, I’d appreciate the opinion of @eslint/eslint-tsc on whether it should be treated as a breaking change.

Since there’s another breaking-change PR, https://github.com/eslint/markdown/pull/664, perhaps this change could be aligned together.

mdjermanovic (approved):
LGTM, thanks! I think the nesting-depth limitation is fine, and since this change can only produce fewer linting errors I don't think it is breaking.

Issue #560, opened 2025-10-13
Bug: Add combined mode to run Markdown rules and code-block linting concurrently for .md files


CyberT33N:
### Environment

Summary
When enabling @eslint/markdown’s processor for `**/*.md`, ESLint lints code blocks (`**/*.md/**`) but stops reporting `markdown/*` rules on the Markdown document itself. When disabling the processor, Markdown rules report but code blocks are not linted. Users cannot get both sets of diagnostics in a single run on the same `.md` file.

Context & Background
- ESLint v9 flat config introduces `language` (Markdown parsing) and still supports processors (code-block extraction).
- @eslint/markdown provides:
  - recommended: `language: "markdown/commonmark"` for `**/*.md` (Markdown rules).
  - processor: `processor: "markdown/markdown"` for `**/*.md` plus `**/*.md/**` code-block handling.
- In practice, processor handling takes precedence per file; both flows don’t apply together.

Environment
- Project version: Not determined (consumer project)
- @eslint/markdown: 7.4.0
- ESLint: 9.36.0
- Node: 22.14.0
- OS: Linux (6.8.0-83-generic)

Affected Versions / Branch
- Affects @eslint/markdown 7.4.0 with ESLint v9 flat config
- Introduced in: Not determined (behavior appears by design in current architecture)

Steps to Reproduce
1. Configure ESLint flat config to include `...markdown.configs.processor` for `**/*.md`.
2. Also add a config item with `language: "markdown/gfm"` (or commonmark) for `**/*.md`.
3. Open a `.md` that has both Markdown rule violations (e.g., duplicate headings) and JS/TS code blocks with lint errors.
4. Observe that with the processor enabled, only code-block diagnostics appear; `markdown/*` rules on the `.md` are absent. Removing the processor restores Markdown diagnostics but drops code-block diagnostics.

Expected Behavior
A single ESLint run should be able to report both:
- Markdown AST rules (`markdown/*`) on the `.md` file.
- JS/TS rules on fenced code blocks extracted from that `.md`.

Actual Behavior
- Processor enabled: only code-block diagnostics show; Markdown rules do not.
- Processor disabled: only Markdown rules show; code-block diagnostics do not.

Impact & Severity
- Major usability gap: authors must choose between Markdown structure and code sample quality in editors/CI, or maintain dual-run setups.
- Documentation and enterprise repos often need both concurrently.

Logs & Evidence
```text
With processor enabled:
[...]

Issue #297, opened 2024-10-30
Bug: Cannot lint both inside and outside code blocks


ehmicky:
### Environment

ESLint version: 9.13.0
@eslint/markdown version: 6.2.1
Node version: 23.1.0
npm version: 10.9.0
Operating System: Ubuntu 24.04

### Which language are you using?

commonmark

## Configuration

```js
// eslint.config.js
import markdown from '@eslint/markdown'

export default [
  {
    plugins: {
      markdown,
    },
  },
  {
    files: ['**/*.md'],
    processor: 'markdown/markdown',
    language: 'markdown/commonmark',
    rules: {
      'markdown/no-duplicate-headings': 'error',
    }
  },
  {
    files: ['**/*.md/*.js'],
    rules: {
      'no-undef': 'error',
    },
  },
]
```

## Markdown file

````md
This is test.md

# Title

# Title

```js
test
```
````

## Actual behavior

```sh
$ ls
eslint.config.js  node_modules  package.json  package-lock.json  test.md
$ eslint

/home/ether/Desktop/try_eslint/test.md
  6:1  error  'test' is not defined  no-undef

✖ 1 problem (1 error, 0 warnings)
```

If I comment out the `processor: 'markdown/markdown'` line in my `eslint.config.js`, I get this instead:

```sh
$ eslint

/home/ether/Desktop/try_eslint/test.md
  3:1  error  Duplicate heading "Title" found  markdown/no-duplicate-headings

✖ 1 problem (1 error, 0 warnings)
```

### What did you expect to happen?

I expect to get both linting errors at the same time.

### What actually happened?

I either get linting applied to the contents outside of the code blocks, or inside of the code blocks, depending on whether `processor` is specified in my `eslint.config.js`.

### Link to Minimal Reproducible Example

https://github.com/ehmicky/eslint-markdown-bug

### Participation

- [ ] I am willing to submit a pull request for this issue.

nzakas:
This is a limitation of ESLint itself, not with this plugin. We're working on it:
https://github.com/eslint/eslint/issues/14745

ehmicky:
Hi @nzakas,

Thanks and sorry as this is a duplicate of https://github.com/eslint/markdown/issues/276.

It seems like the above issue might take a little time to resolve. Are there any known workarounds in the meantime?

nzakas:
The only workaround is to run ESLint twice, one with each configuration. That's what this repo does.

ehmicky:
[...]

Issue #276, opened 2024-08-30
Bug: If `processor` detection block is used it will cause `markdown` rule to not work.


m1rn:
### Environment

ESLint version: 9.9.1
@eslint/markdown version: 6.0.0
Node version: 22.1.0
npm version: 10.7.0
Operating System: Windows

### Which language are you using?

commonmark

### What did you do?

<details>
<summary>Configuration</summary>

```js
import markdown from '@eslint/markdown';

export default [
  {
    files: ['**/*.js'],
    rules: {
      'no-console': 2,
    },
  },
  {
    files: ['**/*.md'],
    plugins: {
      markdown,
    },
    language: 'markdown/commonmark',
    processor: 'markdown/markdown',
    rules: {
      'markdown/fenced-code-language': 2,
      'markdown/heading-increment': 2,
      'markdown/no-empty-links': 2,
      'markdown/no-invalid-label-refs': 2,
      'markdown/no-missing-label-refs': 2,
    },
  },
  {
    files: ['**/*.md/**/*.js'],
    rules: {
      'no-console': 2,
    },
  },
];
```
</details>

```md
# H1

#### Expected to be wrong, but passed

```js
const message = 'Hello, world!'
console.log(message)
```

### What did you expect to happen?

`markdown/heading-increment` should throw an error, but it passes if a `processor` is provided.

### What actually happened?

-

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/stackblitz-starters-hu6gry?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

nzakas:
This is a limitation of ESLint -- you can't use `processor` together with rules right now.

PR #697, opened 2026-07-24
docs: clarify that content and code fences can't be linted together

### LINKED ISSUE #695: Docs: inability to simultaneously lint content and code fences
### Docs page(s)

https://github.com/eslint/markdown/blob/main/README.md

### What documentation issue do you want to solve?

I assumed it possible to simultaneously lint markdown content itself and code inside code fences. However that is not currently possible, due to limitations in the "processor".

That was surprising because each use case is supported separately, and other eslint plugins do allow such behaviour (e.g. eslint-plugin-html, but it avoids the processor problem by using a different "parser").

It took significant time to [discover this limitation](https://github.com/eslint/eslint/issues/14745). Apparently there are others who struggle with this: https://github.com/eslint/markdown/issues/276, https://github.com/eslint/markdown/issues/560, https://github.com/eslint/markdown/issues/297, ...

*Please explicitly document this current limitation.*

Also, please advise us of workarounds, e.g.
- Two configuration files (one for content and one for code fences); works well in git hooks and CI, but not in an IDE (as there's only one eslint language server running)
- Use eslint for code fences, and separate tool (e.g. markdownlint) for content; increases tooling complexity, but allows linting in the IDE (as markdownlint has a vscode extension)
- Possibly using the eslint mdx plugin instead (which uses remarklint to lint content), but it means shoehorning the problem into an inferior solution
- Any other workarounds?

For now, linting content using `markdownlint`, `markdownlint-cli2` and the vscode extension. There is more config to maintain, but I think most linting takes place during <kbd>ctrl+s</kbd>.

The two config files approach is simpler, but there would be no comprehensive linting until git hooks are run, much later (probably pre-push).

### Participation

### AI acknowledgment

### Additional comments

_No response_

lee-kyu-hwan:
#### Prerequisites checklist

#### AI acknowledgment

@eslint/markdown's markdown processor cannot lint a file's Markdown content and its fenced code blocks in a single ESLint run. When the processor is applied, ESLint lints only the extracted code blocks, so Markdown-specific rules configured for the original file do not run.

This limitation was only mentioned briefly in docs/processors/markdown.md and was easy to miss, leading to repeated confusion in #276, #560, #297, and #695.

- Added a brief warning about this limitation to the root `README.md`, immediately after the Processors table, with a link to the processor documentation.
- Expanded the existing `**Important:**` note in `docs/processors/markdown.md` to clarify that only the extracted code blocks are linted and that Markdown-specific rules configured for the original file are silently ignored.
- Added a new "Workarounds" section to `docs/processors/markdown.md` describing two approaches: running ESLint twice with separate configuration files, or using a dedicated Markdown linter such as `markdownlint-cli2` alongside the processor.

#### Related Issues

fixes #695

Whether the wording and placement of the new "Workarounds" section—at the end of docs/processors/markdown.md and linked from both the README and the processor documentation's Important note—match what was discussed in #695, and whether the two listed workarounds are the approaches you would recommend to users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,8 @@ export default defineConfig([
 | ------------------------------------------- | ----------------------------------------------------------------------------------- |
 | [`markdown`](./docs/processors/markdown.md) | Extract fenced code blocks from the Markdown code so they can be linted separately. |
 
+**Important:** You cannot lint a file's Markdown content and its fenced code blocks in the same run. When the `markdown` processor is applied to a file, only the extracted code blocks are linted — any Markdown rules (such as `markdown/heading-increment`) configured for that file are silently ignored. This is an ESLint limitation, not something this plugin can work around. See [Workarounds](./docs/processors/markdown.md#workarounds) for ways to lint both.
+
 ## Migration from `eslint-plugin-markdown`
 
 See [Migration](./docs/migration.md#from-eslint-plugin-markdown).
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ export default defineConfig([
 | ------------------------------------------- | ----------------------------------------------------------------------------------- |
 | [`markdown`](./docs/processors/markdown.md) | Extract fenced code blocks from the Markdown code so they can be linted separately. |
 
-**Important:** You cannot lint a file's Markdown content and its fenced code blocks in a single ESLint run. When the `markdown` processor is applied to a file, only the extracted code blocks are linted; any Markdown-specific rules configured for the original file, such as `markdown/heading-increment`, are silently ignored. This is a current limitation of ESLint processors. See [Workarounds](./docs/processors/markdown.md#workarounds) for ways to lint both.
+**Important:** You cannot lint a file's Markdown content and its fenced code blocks in a single ESLint run. When the `markdown` processor is applied, only the extracted code blocks are linted. This is a current limitation of ESLint processors. See [Linting Markdown Content and Code Blocks](./docs/processors/markdown.md#linting-markdown-content-and-code-blocks) for a way to lint both.
 
 ## Migration from `eslint-plugin-markdown`
 
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ export default defineConfig([
 | ------------------------------------------- | ----------------------------------------------------------------------------------- |
 | [`markdown`](./docs/processors/markdown.md) | Extract fenced code blocks from the Markdown code so they can be linted separately. |
 
-**Important:** You cannot lint a file's Markdown content and its fenced code blocks in the same run. When the `markdown` processor is applied to a file, only the extracted code blocks are linted — any Markdown rules (such as `markdown/heading-increment`) configured for that file are silently ignored. This is an ESLint limitation, not something this plugin can work around. See [Workarounds](./docs/processors/markdown.md#workarounds) for ways to lint both.
+**Important:** You cannot lint a file's Markdown content and its fenced code blocks in a single ESLint run. When the `markdown` processor is applied to a file, only the extracted code blocks are linted; any Markdown-specific rules configured for the original file, such as `markdown/heading-increment`, are silently ignored. This is a current limitation of ESLint processors. See [Workarounds](./docs/processors/markdown.md#workarounds) for ways to lint both.
 
 ## Migration from `eslint-plugin-markdown`
 
--- a/docs/processors/markdown.md
+++ b/docs/processors/markdown.md
@@ -38,7 +38,7 @@ This is plain text and doesn't get linted.
 
 Unless a fenced code block's syntax appears as a file extension in file patterns in your config file, it will be ignored.
 
-**Important:** You cannot combine this processor and Markdown-specific linting rules. You can either lint the code blocks or lint the Markdown, but not both. This is an ESLint limitation.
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -271,6 +271,8 @@ export default defineConfig([
 | ------------------------------------------- | ----------------------------------------------------------------------------------- |
 | [`markdown`](./docs/processors/markdown.md) | Extract fenced code blocks from the Markdown code so they can be linted separately. |
 
+**Important:** You cannot lint a file's Markdown content and its fenced code blocks in a single ESLint run. When the `markdown` processor is applied to a file, only the extracted code blocks are linted; any Markdown-specific rules configured for the original file, such as `markdown/heading-increment`, are silently ignored. This is a current limitation of ESLint processors. See [Workarounds](./docs/processors/markdown.md#workarounds) for ways to lint both.
```
DMartens: The sentence "any Markdown-specific rules configured for the original file, such as `markdown/heading-increment`, are silently ignored" is untrue, as the rules are not run in this case.
Can you please remove this sentence?

docs/processors/markdown.md
```diff
@@ -238,3 +238,18 @@ highlighting. Skip the block to prevent warnings about invalid syntax.
...
+- **Run ESLint with two configuration files.** Keep your existing configuration with the `markdown` processor for code blocks, and add a second configuration with the Markdown language and rules for content. Then run ESLint twice:
+
+  ```sh
+  eslint .                              # lints fenced code blocks
+  eslint -c eslint.config-content.js .  # lints Markdown content
+  ```
+
+  This works well in CI and git hooks. However, most editor integrations use only one ESLint configuration at a time, so Markdown content errors may not appear in the editor. This repository uses this approach in its lint scripts and git hooks.
+
+- **Use a separate Markdown linter.** Keep the `markdown` processor for code blocks, and use a dedicated tool such as [`markdownlint-cli2`](https://github.com/DavidAnson/markdownlint-cli2) to lint the Markdown content. An editor extension such as [`vscode-markdownlint`](https://github.com/DavidAnson/vscode-markdownlint) can provide real-time feedback while you type. The tradeoff is having another tool and configuration to maintain.
```
DMartens: I would not consider using a different tool a workaround,
as this option always exists for the user if they are unsatisfied.
The point is that the user wants to use ESLintfor linting Markdown and the embedded code blocks.

docs/processors/markdown.md
```diff
@@ -238,3 +238,18 @@ highlighting. Skip the block to prevent warnings about invalid syntax.
...
+
+## Workarounds
+
+You can't lint both in a single ESLint run, but you can use one of the following approaches:
+
+- **Run ESLint with two configuration files.** Keep your existing configuration with the `markdown` processor for code blocks, and add a second configuration with the Markdown language and rules for content. Then run ESLint twice:
+
+  ```sh
+  eslint .                              # lints fenced code blocks
+  eslint -c eslint.config-content.js .  # lints Markdown content
```
DMartens: Can you please add the relevant sections from the ESLint configuration or link to the existing examples from the README?

docs/processors/markdown.md
lonix1: Good point. So maybe avoid the word "workaround", but give people the knowledge to solve this issue?

I personally prefer the pure eslint two-config-file approach as it's simpler. However I quickly realised that most linting takes place during <kbd>ctrl+s</kbd> - which isn't possible using that approach - so using a separate tool with IDE support is more practical albeit slightly more complex.

README.md
lee-kyu-hwan: Removed the sentence. The note now states that when the processor is applied, ESLint lints only the extracted code blocks.

docs/processors/markdown.md
lee-kyu-hwan: Added links to the existing examples for both the processor and recommended configurations. The section also names eslint.config.js and eslint.config-content.js to match the commands shown.

docs/processors/markdown.md
lee-kyu-hwan: Removed the separate Markdown linter option and renamed the section to Linting Markdown Content and Code Blocks, so the ESLint-only approach is no longer described as a workaround.

@lonix1 The section still notes that editor integrations may not report Markdown content errors while you type, but it no longer recommends a separate tool for that case.

@DMartens Would it be appropriate to briefly mention a dedicated Markdown linter as an alternative when real-time editor feedback matters, without presenting it as an ESLint workaround? Or would you prefer this page to remain focused on ESLint-only solutions?

lumirlumir: @lonix1 Just a note for reference: there is an ongoing RFC for the `Prelint` concept, and if it is merged and implemented in this plugin one day, running both the processor and markdown plugin in a single run would be possible.

https://github.com/eslint/rfcs/pull/105

lumirlumir (approved):
LGTM, thanks!

Leaving it open for @DMartens to verify the changes before merging.

docs/processors/markdown.md
DMartens: In my opinion it should ESLint-only as my point is the user wants to use certain ESLint rules. Even if most rules are also covered by markdownlint, most of them do not cover other available ESLint plugins, e.g. [markdown-preferences](https://github.com/ota-meshi/eslint-plugin-markdown-preferences) which the user may want to use.

DMartens (approved):
Changes LGTM, thanks.

PR #708, opened 2026-08-05
fix: add `mdast` type augmentation for TOML and JSON nodes

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

CommonMark and GFM.

### What did you do?

I expected the `Toml` and `Json` node types to be present in `FrontmatterContentMap` and `RootContentMap` from [`mdast`](https://github.com/syntax-tree/mdast), but they aren’t.

This prevents users from accessing the `Toml` and `Json` node types when using the `FrontmatterContent` or `RootContent` types, which are typically accessed through `node.children` when finding child nodes or directly via `@types/mdast`.

I expected the `Toml` and `Json` node types to be present in `FrontmatterContentMap` and `RootContentMap` from [`mdast`](https://github.com/syntax-tree/mdast), but they aren’t.

### What actually happened?

It produces type errors when accessing the `Toml` and `Json` types, as shown below:

```ts
import type { FrontmatterContent } from "mdast";

({}) as Toml satisfies FrontmatterContent;

({}) as Json satisfies FrontmatterContent;
```

Pasting the following code into `tests/types/types.test.ts` produces TypeScript type errors:

```ts
import type {
	FrontmatterContent,
	FrontmatterContentMap,
	RootContent,
	RootContentMap,
} from "mdast";

// Check that `Toml` is registered as mdast front matter and root content.
({}) as Toml satisfies FrontmatterContent;
({}) as Toml satisfies FrontmatterContentMap["toml"];
({}) as Toml satisfies RootContent;
({}) as Toml satisfies RootContentMap["toml"];

// Check that `Json` is registered as mdast front matter and root content.
({}) as Json satisfies FrontmatterContent;
({}) as Json satisfies FrontmatterContentMap["json"];
({}) as Json satisfies RootContent;
({}) as Json satisfies RootContentMap["json"];
```

This change adds `mdast` type augmentation and type tests for TOML and JSON nodes, which is a common pattern in the `mdast` type system. This allows us to define custom nodes and extend the `mdast` typings.

For example, `mdast-util-math`, which is used as a dependency in our project, also declares the following type augmentation so that users can access the extended types directly from the `mdast` typings:

https://github.com/syntax-tree/mdast-util-math/blob/main/index.d.ts#L93-L107

```ts
// Add nodes to tree.
declare module 'mdast' {
  interface BlockContentMap {
    math: Math
  }


  interface PhrasingContentMap {
    inlineMath: InlineMath
  }


  interface RootContentMap {
    inlineMath: InlineMath
    math: Math
  }
}
```

The JSDoc in `mdast` also mentions that this can be augmented when custom node types are used:

image

## Related Issues

N/A

I identified this problem while working on Rust parser integration and my custom plugins.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -114,6 +114,21 @@ export interface Json extends Literal {
  */
 export interface JsonData extends Data {}
 
+/**
+ * Registers additional mdast nodes as valid front matter and root content.
+ */
+declare module "mdast" {
+	interface FrontmatterContentMap {
+		toml: Toml;
+		json: Json;
+	}
+
+	interface RootContentMap {
+		toml: Toml;
+		json: Json;
+	}
+}
+
 //------------------------------------------------------------------------------
 // Exports: Language and Source Code
 //------------------------------------------------------------------------------
--- a/tests/types/types.test.ts
+++ b/tests/types/types.test.ts
@@ -47,6 +47,15 @@ import type {
 	TableRow,
 	// Extensions (front matter)
 	Yaml,
+	// Contents
+	BlockContent,
+	BlockContentMap,
+	FrontmatterContent,
+	FrontmatterContentMap,
+	PhrasingContent,
+	PhrasingContentMap,
+	RootContent,
+	RootContentMap,
 } from "mdast";
 import type { InlineMath, Math } from "mdast-util-math";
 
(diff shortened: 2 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks. While I find it weird that consumers have to add the type but this also documented in their [README](https://github.com/syntax-tree/mdast-util-frontmatter#types).

PR #716, opened 2026-08-16
feat: add no-heading-like-paragraph rule

### LINKED ISSUE #700: New Rule: no-heading-like-paragraph
### Rule details

This rule should report when the first line of a paragraph starts with seven or more hash characters (`#`) followed by a space or tab.  CommonMark ATX headings support one to six hash characters. Therefore, content such as `####### Installation` is parsed as a regular paragraph rather than a heading.

### Example code

```markdown
// The following code should trigger a warning
####### Installation

######## Configuration
```

### Prior Art

remark-lint already provides a rule with the same purpose:

[`remark-lint-no-heading-like-paragraph`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-heading-like-paragraph)

This rule also warns about paragraphs that start with seven or more hash characters and appear to be intended as headings.

### Participation

### AI acknowledgment

### Additional comments

_No response_

 Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

Gaic4o:
#### Prerequisites checklist

#### AI acknowledgment

This PR implements the `no-heading-like-paragraph` rule proposed and accepted in #700.

CommonMark ATX headings support at most six `#` characters. As a result, content such as `####### Installation` is parsed as a paragraph rather than a heading, even though it can easily look like an intended heading in the source.

The rule reports these heading-like paragraphs so that likely heading mistakes can be identified.

* Added the `no-heading-like-paragraph` rule for paragraphs that look like ATX headings with seven or more leading `#` characters.
* The rule checks the raw source to avoid false positives for escaped or character-referenced input.
* Added suggestions to either convert the paragraph to a valid level-6 heading or escape the first `#` to keep it as a paragraph.
* No automatic fix is provided because the author's intent is ambiguous.
#### Related Issues

fixes #700

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

## Summary by CodeRabbit

* **New Features**
  * Added a rule that detects paragraph lines beginning with seven or more hash characters where headings could appear.
  * Provides suggestions to reduce the sequence to six hashes or escape the leading hash.
* **Documentation**
  * Added comprehensive rule documentation, examples, configuration guidance, and CommonMark references.
  * Listed the rule in the README rules table as not recommended.
* **Tests**
  * Added extensive coverage for valid and invalid Markdown, including indentation, block quotes, lists, line endings, footnotes, code spans, and link or image titles.

### DIFF
--- a/docs/rules/no-heading-like-paragraph.md
+++ b/docs/rules/no-heading-like-paragraph.md
@@ -4,15 +4,15 @@ Disallow paragraphs that look like ATX headings.
 
 ## Background
 
-In Markdown, an ATX heading opens with one to six hash (`#`) characters followed by a space or a tab, so `###### Installation` is a level 6 heading. A seventh hash character exceeds the maximum heading depth, and `####### Installation` doesn't create a heading at all. Markdown renders it as a paragraph whose text starts with seven literal hash characters.
+In Markdown, an ATX heading opens with one to six hash (`#`) characters followed by a space, a tab, or a line ending, so `###### Installation` is a level 6 heading. Seven or more hash characters aren't heading syntax at all, so Markdown reads `####### Installation` as paragraph text that begins with seven literal hash characters.
 
 This is almost always a typo, and it's easy to miss in review because the source still reads like a heading.
 
 ## Rule Details
 
-This rule flags a paragraph that begins with seven or more hash characters followed by a space, a tab, a line ending, or the end of the paragraph.
+This rule flags a line of a paragraph that begins with seven or more hash characters followed by a space, a tab, a line ending, or the end of the paragraph. It checks continuation lines as well as the first line, because six or fewer hash characters in the same position would open a real heading. Block quote markers and up to three spaces of indentation may precede the hash characters, the same positions where an ATX heading is allowed to start.
 
-This rule ignores anything that can't open an ATX heading. `#######Installation` has no whitespace to delimit the hash characters, and `\####### Installation` and `&#35;###### Installation` escape their leading hash character on purpose.
+This rule ignores anything that can't open an ATX heading. `#######Installation` has no whitespace to delimit the hash characters, `\####### Installation` and `&#35;###### Installation` escape their leading hash character on purpose, and four or more spaces of indentation are too many for a heading.
 
 This rule provides suggestions rather than an automatic fix, because the number of hash characters alone doesn't reveal which correction the author intended:
 
@@ -31,6 +31,13 @@ Examples of **incorrect** code for this rule:
 > ####### Usage
 
 - ####### Options
+
+Install the package first.
+####### Installation
+
+> foo
+> ####### hi
+> bar
 ```
 
 Examples of **correct** code for this rule:
@@ -49,16 +56,18 @@ Examples of **correct** code for this rule:
 \####### Not a heading
 
 Seven ####### characters in the middle of a paragraph.
-```
 
-This rule only checks the beginning of a paragraph, so it ignores hash characters on a continuation line:
-
-```markdown
 Install the package first.
-####### Installation
+###### Installation
+
+> foo
+> ###### hi
+> bar
 ```
 
-Because `####### Installation` can't start a heading, Markdown folds it into the preceding paragraph as a lazy continuation line. The same text with six or fewer hash characters would interrupt the paragraph and become a real heading.
+## Options
+
+This rule has no options.
 
 ## When Not to Use It
 
--- a/docs/rules/no-heading-like-paragraph.md
+++ b/docs/rules/no-heading-like-paragraph.md
@@ -1,0 +1,73 @@
+# no-heading-like-paragraph
+
+Disallow paragraphs that look like ATX headings.
+
+## Background
+
+In Markdown, an ATX heading opens with one to six hash (`#`) characters followed by a space or a tab, so `###### Installation` is a level 6 heading. A seventh hash character exceeds the maximum heading depth, and `####### Installation` doesn't create a heading at all. Markdown renders it as a paragraph whose text starts with seven literal hash characters.
+
+This is almost always a typo, and it's easy to miss in review because the source still reads like a heading.
+
+## Rule Details
+
+This rule flags a paragraph that begins with seven or more hash characters followed by a space, a tab, a line ending, or the end of the paragraph.
+
+This rule ignores anything that can't open an ATX heading. `#######Installation` has no whitespace to delimit the hash characters, and `\####### Installation` and `&#35;###### Installation` escape their leading hash character on purpose.
+
+This rule provides suggestions rather than an automatic fix, because the number of hash characters alone doesn't reveal which correction the author intended:
+
+* Replace the leading hash characters with `######`, which makes the paragraph a level 6 heading. `####### Installation` becomes `###### Installation`.
+* Escape the leading hash character, which leaves the rendered output unchanged. `####### Installation` becomes `\####### Installation`.
+
+Examples of **incorrect** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-heading-like-paragraph: "error" -->
+
+####### Installation
+
+######## Configuration
+
+> ####### Usage
+
+- ####### Options
+```
+
+Examples of **correct** code for this rule:
+
+```markdown
+<!-- eslint markdown/no-heading-like-paragraph: "error" -->
+
+###### Installation
+
+> ###### Usage
+
+- ###### Options
+
+#######Configuration
+
+\####### Not a heading
+
+Seven ####### characters in the middle of a paragraph.
+```
+
+This rule only checks the beginning of a paragraph, so it ignores hash characters on a continuation line:
+
+```markdown
+Install the package first.
+####### Installation
+```
+
+Because `####### Installation` can't start a heading, Markdown folds it into the preceding paragraph as a lazy continuation line. The same text with six or fewer hash characters would interrupt the paragraph and become a real heading.
+
+## When Not to Use It
+
+If you intentionally write paragraphs that begin with seven or more hash characters, you can safely disable this rule.
+
+## Prior Art
+
+* [remark-lint-no-heading-like-paragraph](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-heading-like-paragraph)
+
+## Further Reading
+
+* [CommonMark Spec: ATX Headings](https://spec.commonmark.org/0.31.2/#atx-headings)
--- a/src/rules/no-heading-like-paragraph.js
+++ b/src/rules/no-heading-like-paragraph.js
@@ -38,7 +38,6 @@ export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
 		type: "problem",
 
 		docs: {
-			recommended: false,
 			description: "Disallow paragraphs that look like ATX headings",
 			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-heading-like-paragraph.md",
 		},
@@ -48,7 +47,8 @@ export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
 		messages: {
 			headingLikeParagraph:
 				"Unexpected paragraph starting with {{count}} hash characters. ATX headings support at most 6.",
-			useMaxDepthHashes: 'Replace "{{hashes}}" with "{{maxDepthHashes}}".',
+			useMaxDepthHashes:
+				'Replace "{{hashes}}" with "{{maxDepthHashes}}".',
 			escapeLeadingHash: "Escape the leading hash character.",
 		},
 	},
@@ -84,11 +84,6 @@ export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
 					},
 					messageId: "headingLikeParagraph",
 					data: { count: hashes.length },
-
-					/*
-					 * There's no autofix, because the number of hash characters alone
-					 * doesn't reveal which of the two corrections the author intended.
-					 */
 					suggest: [
 						{
 							messageId: "useMaxDepthHashes",
--- a/src/rules/no-heading-like-paragraph.js
+++ b/src/rules/no-heading-like-paragraph.js
@@ -19,12 +19,21 @@
 //-----------------------------------------------------------------------------
 
 /**
- * Matches seven or more hash characters at the start of a paragraph, followed by a
- * space, a tab, a line ending, or the end of the paragraph. This mirrors the way
- * CommonMark delimits the opening sequence of an ATX heading, so a no-break space
- * doesn't count as a delimiter.
+ * Matches seven or more hash characters at the start of a line within a paragraph,
+ * followed by a space, a tab, a line ending, or the end of the paragraph. This mirrors
+ * the way CommonMark delimits the opening sequence of an ATX heading, so a no-break
+ * space doesn't count as a delimiter.
+ *
+ * This pattern avoids the `m` flag, which would also treat U+2028 and U+2029 as line
+ * boundaries even though Markdown doesn't. `(?:^|(?<=[\r\n]))` starts a new line only
+ * after an actual carriage return or line feed.
+ *
+ * Block quote markers and up to three spaces of indentation may precede the hash
+ * characters, because a heading with six or fewer hash characters would still open in
+ * that position.
  */
-const headingLikeParagraphPattern = /^#{7,}(?=[ \t\r\n]|$)/u;
+const headingLikeParagraphPattern =
+	/(?:^|(?<=[\r\n]))(?: {0,3}>[ \t]?)* {0,3}(?<hashes>#{7,})(?=[ \t\r\n]|$)/gu;
 
 /** The longest opening sequence an ATX heading allows. */
 const maxDepthHashes = "######";
@@ -65,47 +74,52 @@ export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
 				 * paragraph whose text starts with seven hash characters, but in each case
 				 * the author escaped the leading hash on purpose.
 				 */
-				const match = headingLikeParagraphPattern.exec(
-					sourceCode.getText(node),
-				);
+				const text = sourceCode.getText(node);
 
-				if (match === null) {
-					return;
+				/** @type {RegExpExecArray | null} */
+				let match;
+
+				while (
+					(match = headingLikeParagraphPattern.exec(text)) !== null
+				) {
+					const { hashes } = match.groups;
+					const startOffset =
+						node.position.start.offset +
+						match.index +
+						match[0].length -
+						hashes.length;
+					const endOffset = startOffset + hashes.length;
+
+					context.report({
+						loc: {
+							start: sourceCode.getLocFromIndex(startOffset),
+							end: sourceCode.getLocFromIndex(endOffset),
+						},
+						messageId: "headingLikeParagraph",
+						data: { count: hashes.length },
+						suggest: [
+							{
+								messageId: "useMaxDepthHashes",
+								data: { hashes, maxDepthHashes },
+								fix(fixer) {
+									return fixer.replaceTextRange(
+										[startOffset, endOffset],
+										maxDepthHashes,
+									);
+								},
+							},
+							{
+								messageId: "escapeLeadingHash",
+								fix(fixer) {
+									return fixer.insertTextBeforeRange(
+										[startOffset, startOffset + 1],
+										"\\",
+									);
+								},
+							},
+						],
+					});
 				}
-
-				const [hashes] = match;
-				const startOffset = node.position.start.offset;
-				const endOffset = startOffset + hashes.length;
-
-				context.report({
-					loc: {
-						start: sourceCode.getLocFromIndex(startOffset),
-						end: sourceCode.getLocFromIndex(endOffset),
-					},
-					messageId: "headingLikeParagraph",
-					data: { count: hashes.length },
-					suggest: [
-						{
-							messageId: "useMaxDepthHashes",
-							data: { hashes, maxDepthHashes },
-							fix(fixer) {
-								return fixer.replaceTextRange(
-									[startOffset, endOffset],
-									maxDepthHashes,
-								);
-							},
-						},
-						{
-							messageId: "escapeLeadingHash",
-							fix(fixer) {
-								return fixer.insertTextBeforeRange(
-									[startOffset, startOffset + 1],
-									"\\",
-								);
-							},
-						},
-					],
-				});
 			},
 		};
 	},
--- a/src/rules/no-heading-like-paragraph.js
+++ b/src/rules/no-heading-like-paragraph.js
@@ -1,0 +1,117 @@
+/**
+ * @fileoverview Rule to disallow paragraphs that look like ATX headings in Markdown.
+ * @author Gaic4o
+ */
+
+//-----------------------------------------------------------------------------
+// Type Definitions
+//-----------------------------------------------------------------------------
+
+/**
+ * @import { MarkdownRuleDefinition } from "../types.js";
+ * @typedef {"headingLikeParagraph" | "useMaxDepthHashes" | "escapeLeadingHash"} NoHeadingLikeParagraphMessageIds
+ * @typedef {[]} NoHeadingLikeParagraphOptions
+ * @typedef {MarkdownRuleDefinition<{ RuleOptions: NoHeadingLikeParagraphOptions, MessageIds: NoHeadingLikeParagraphMessageIds }>} NoHeadingLikeParagraphRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/**
+ * Matches seven or more hash characters at the start of a paragraph, followed by a
+ * space, a tab, a line ending, or the end of the paragraph. This mirrors the way
+ * CommonMark delimits the opening sequence of an ATX heading, so a no-break space
+ * doesn't count as a delimiter.
+ */
+const headingLikeParagraphPattern = /^#{7,}(?=[ \t\r\n]|$)/u;
+
+/** The longest opening sequence an ATX heading allows. */
+const maxDepthHashes = "######";
+
+//-----------------------------------------------------------------------------
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
+	meta: {
+		type: "problem",
+
+		docs: {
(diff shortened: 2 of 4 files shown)


tests/rules/no-heading-like-paragraph.test.js
```diff
@@ -0,0 +1,522 @@
...
+//------------------------------------------------------------------------------
+// GFM Tests
+//------------------------------------------------------------------------------
+
+const gfmRuleTester = new RuleTester({
+	plugins: {
+		markdown,
+	},
+	language: "markdown/gfm",
+});
```
lumirlumir: I think we can use `language: "markdown/gfm"` where necessary to test GFM mode, as shown below. Consolidating these `RuleTester` test cases with the ones above also seems helpful.

https://github.com/eslint/markdown/blob/ac31775656f65f78bc02c0a377c0a85bd131f9f8/tests/rules/no-missing-label-refs.test.js#L126-L131

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,117 @@
...
+//-----------------------------------------------------------------------------
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
+	meta: {
+		type: "problem",
+
+		docs: {
+			recommended: false,
```
lumirlumir:
```suggestion

```

Non-blocking stylistic choice: The repository usually omits the `recommended` field when it is `false`.

tests/rules/no-heading-like-paragraph.test.js
```diff
@@ -0,0 +1,522 @@
...
+			errors: [
+				{
+					messageId: "headingLikeParagraph",
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 8,
+					suggestions: [
+						{
+							messageId: "useMaxDepthHashes",
```
lumirlumir, line 154: Could you add the missing `data` properties to the `invalid` test cases? This applies to all other `invalid` test cases as well.

lumirlumir (changes requested):
> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca): confirmed.

Can you take a look at the CI failure? Running `npm run fmt` should resolve the problem.

docs/rules/no-heading-like-paragraph.md
```diff
@@ -0,0 +1,73 @@
...
+- ###### Options
+
+#######Configuration
+
+\####### Not a heading
+
+Seven ####### characters in the middle of a paragraph.
+```
+
+This rule only checks the beginning of a paragraph, so it ignores hash characters on a continuation line:
```
DMartens: This case should still be handled by the rule as the writer expects to create a heading (which would be created if it would use valid heading syntax).
This could be accomplished by setting the `m`(ultiline) flag for `headingLikeParagraphPattern`.

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,117 @@
...
+			paragraph(node) {
+				/*
+				 * Read the raw source text instead of the `value` of the first `text`
+				 * child, because `value` already resolves character escapes and character
+				 * references. Both `\####### Foo` and `&#35;###### Foo` render as a
+				 * paragraph whose text starts with seven hash characters, but in each case
+				 * the author escaped the leading hash on purpose.
+				 */
+				const match = headingLikeParagraphPattern.exec(
+					sourceCode.getText(node),
```
DMartens:
```suggestion
					sourceCode.getText(node.children[0]),
```
Why not use just the first child. This would make the text potentially smaller which is always good when using regular expressions.

tests/rules/no-heading-like-paragraph.test.js
```diff
@@ -0,0 +1,522 @@
...
+						{
+							messageId: "escapeLeadingHash",
+							output: "\\#######\tInstallation",
+						},
+					],
+				},
+			],
+		},
+		{
+			code: "#######",
```
DMartens, line 170: I do not think this should be an invalid test case as the text after the hashes are missing.
Some may use thisas decoration.

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,117 @@
...
+
+				context.report({
+					loc: {
+						start: sourceCode.getLocFromIndex(startOffset),
+						end: sourceCode.getLocFromIndex(endOffset),
+					},
+					messageId: "headingLikeParagraph",
+					data: { count: hashes.length },
+
+					/*
```
DMartens: This comment is unnecessary as two suggestions are provided, it is clear that there can be no autofix.

docs/rules/no-heading-like-paragraph.md
Gaic4o: I agree. I’ll update the rule to handle continuation lines as well. I think this will also keep it consistent with no-missing-atx-heading-space.

However, since the m flag alone can miss continuation lines inside blockquotes or list items, would it be okay to handle those container cases as well as the top-level case and add tests for them?

tests/rules/no-heading-like-paragraph.test.js
Gaic4o: That makes sense. If there’s no text after the hashes, I agree that it’s much less clear whether the author actually intended to create a heading. For reference, `remark-lint-no-heading-like-paragraph` does report a bare `#######`, but I agree that this case could reasonably be treated as decoration.

I’d just like to clarify the intended scope. Should cases like `####### `, where the hashes are followed only by trailing whitespace, and `#######\nText`, where the first line of a multi-line paragraph contains only the hashes, also be ignored? Or should only the single-line bare `#######` case be excluded?

src/rules/no-heading-like-paragraph.js
Gaic4o: Using only the first child could introduce false positives when inline markup immediately follows the hashes. For example, in the current valid test case `#######*Installation*`, the source of the first child ends with `#######`, so the `$` condition in the regular expression could match. However, in the full paragraph source, the hashes are followed by `*`, so it does not satisfy the ATX heading delimiter condition.

Also, checking only the first child would not be sufficient for handling continuation lines discussed in the other thread. For example, in `Install **first**.\n####### Config`, the heading-like text can be contained in a later child rather than the first one, so this case would be missed.

For correctness, I think matching against the whole paragraph source is safer than checking only the first child.

src/rules/no-heading-like-paragraph.js
Gaic4o: I removed the unnecessary comment as you suggested. Thank you!

tests/rules/no-heading-like-paragraph.test.js
Gaic4o: Added the required data properties to all invalid test cases.

src/rules/no-heading-like-paragraph.js
Gaic4o: Removed recommended: false to match the existing repository style. Thanks!

tests/rules/no-heading-like-paragraph.test.js
Gaic4o: Integrated the GFM cases into the existing RuleTester as suggested.

docs/rules/no-heading-like-paragraph.md
DMartens: I do not think we need to check for continuation lines inside a container (like blockquotes).
What do you think @lumirlumir?

src/rules/no-heading-like-paragraph.js
DMartens: For this edge case you can pass `afterCount` to get also the following characters, so
`sourceCode.getText(node.children[0], 0, 1)`.

DMartens: Scratch this suggestion. The whole paragraph text is required if the paragraph has continuation lines.

lumirlumir:
I’m sorry for the delay. I’m having a fairly busy week and expect to remain busy through next week, but I’ll be sure to revisit this PR in about a week.

docs/rules/no-heading-like-paragraph.md
lumirlumir: > This could be accomplished by setting the m(ultiline) flag for headingLikeParagraphPattern.

The `m` flag also treats LS (U+2028) and PS (U+2029) characters as line breaks, but they aren’t line breaks in Markdown. So using the `(?:^|(?<=[\r\n]))` pattern in a regex, as shown below, would be helpful in this case:

https://github.com/eslint/markdown/blob/5bcfe20078568453f2e28b2c818bb245b2976ba0/src/rules/no-missing-atx-heading-space.js#L22

Also, adding a regression test for LS and PS would be a nice addition here.

> However, since the m flag alone can miss continuation lines inside blockquotes or list items, would it be okay to handle those container cases as well as the top-level case and add tests for them?

If a user wanted to write a blockquote with heading such as below, the `###### hi` is correctly recognized as a valid heading.

[Playground](https://explorer.eslint.org/#eslint-explorer=v2.eyJzdGF0ZSI6eyJ0b29sIjoiYXN0IiwiY29kZSI6eyJqYXZhc2NyaXB0IjoiIiwianNvbiI6IiIsIm1hcmtkb3duIjoiPiBmb29cbj4gIyMjIyMjIGhpXG4-IGJhciIsImNzcyI6IkBrZXlmcmFtZXMgdGVzdCB7XG5cdDAlIHtcblx0XHRvcGFjaXR5OiAwO1xuXHR9XG5cbiAgICBmcm9tIHtcblx0XHRvcGFjaXR5OiAwO1xuICAgIH1cblxuXHQxMDAlIHtcblx0XHRvcGFjaXR5OiAxO1xuXHR9XG59IiwiaHRtbCI6IjwhRE9DVFlQRSBodG1sPlxuPCEtLVxuVHlwZSBvciBwYXN0ZSBzb21lIEhUTUwgaGVyZSB0byBsZWFybiBtb3JlIGFib3V0XG50aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxuXG5UaGUgdGFicyBhcmU6XG5cbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cbmJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIHRoZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUuIFlvdVxuY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxuLS0-XG5cbjxodG1sIGxhbmc9XCJlblwiPlxuICAgIDxoZWFkPlxuICAgICAgICA8bWV0YSBjaGFyc2V0PVwiVVRGLThcIj5cbiAgICAgICAgPHRpdGxlPkhUTUw8L3RpdGxlPlxuICAgIDwvaGVhZD5cbiAgICA8Ym9keT5cbiAgICAgICAgPHA-VGV4dDwvcD5cbiAgICA8L2JvZHk-XG48L2h0bWw-In0sImxhbmd1YWdlIjoibWFya2Rvd24iLCJqc09wdGlvbnMiOnsicGFyc2VyIjoiZXNwcmVlIiwic291cmNlVHlwZSI6Im1vZHVsZSIsImlzVmVyc2lvbiI6ImxhdGVzdCIsImlzSlNYIjpmYWxzZSwiZXNWZXJzaW9uIjoibGF0ZXN0In0sImpzb25PcHRpb25zIjp7Impzb25Nb2RlIjoianNvbjUiLCJhbGxvd1RyYWlsaW5nQ29tbWFzIjpmYWxzZX0sImNzc09wdGlvbnMiOnsiY3NzTW9kZSI6ImNzcyIsInRvbGVyYW50IjpmYWxzZX0sIm1hcmtkb3duT3B0aW9ucyI6eyJtYXJrZG93bk1vZGUiOiJjb21tb25tYXJrIiwibWFya2Rvd25Gcm9udG1hdHRlciI6Im9mZiIsIm1hcmtkb3duTWF0aCI6ZmFsc2V9LCJodG1sT3B0aW9ucyI6eyJ0ZW1wbGF0ZUVuZ2luZVN5bnRheCI6Im5vbmUiLCJmcm9udG1hdHRlciI6Im9mZiJ9LCJ3cmFwIjp0cnVlLCJ2aWV3TW9kZXMiOnsiYXN0VmlldyI6InRyZWUiLCJzY29wZVZpZXciOiJmbGF0IiwicGF0aFZpZXciOiJncmFwaCJ9LCJwYXRoSW5kZXgiOnsiaW5kZXgiOjAsImluZGV4ZXMiOjF9LCJlc3F1ZXJ5U2VsZWN0b3IiOnsic2VsZWN0b3IiOiIifX0sInZlcnNpb24iOjB9)

```md
> foo
> ###### hi
> bar
```

So, if I understood the thread correctly, and if our intention is to detect `#######` (seven or more repeated hashes) in cases like the one below, I think it should also be reported, since it can be a valid heading when 1–6 `#` characters are used in this case.

```md
> foo
> ####### hi
> bar
```

tests/rules/no-heading-like-paragraph.test.js
lumirlumir: Per the CommonMark specification, and given that `remark-lint-no-heading-like-paragraph` reports this case, I think we should report this case as well. This behavior is spec-conformant, and it would be best to follow the spec, especially since there have been many prior requests about specification deviations.

Reference: https://spec.commonmark.org/0.31.2/#example-79

image

docs/rules/no-heading-like-paragraph.md
```diff
@@ -0,0 +1,73 @@
...
+
+```markdown
+Install the package first.
+####### Installation
+```
+
+Because `####### Installation` can't start a heading, Markdown folds it into the preceding paragraph as a lazy continuation line. The same text with six or fewer hash characters would interrupt the paragraph and become a real heading.
+
```
lumirlumir, line 71: I think it would be good to add a dedicated `## Options` section explaining that this rule has no options, since there was a prior request in ESLint to always include an options section even when none are available: https://github.com/eslint/eslint/issues/20196

docs/rules/no-heading-like-paragraph.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+This rule only checks the beginning of a paragraph, so it ignores hash characters on a continuation line:
+
+```markdown
+Install the package first.
+####### Installation
+```
+
+Because `####### Installation` can't start a heading, Markdown folds it into the preceding paragraph as a lazy continuation line. The same text with six or fewer hash characters would interrupt the paragraph and become a real heading.
```
lumirlumir:
```suggestion
```

Similar to the previous comment, this line is misleading, since the following is recognized as a valid paragraph and heading:

[Playground](https://explorer.eslint.org/#eslint-explorer=v2.eyJzdGF0ZSI6eyJ0b29sIjoiYXN0IiwiY29kZSI6eyJqYXZhc2NyaXB0IjoiIiwianNvbiI6IiIsIm1hcmtkb3duIjoiSW5zdGFsbCB0aGUgcGFja2FnZSBmaXJzdC5cbiMjIyMjIyBJbnN0YWxsYXRpb24iLCJjc3MiOiJAa2V5ZnJhbWVzIHRlc3Qge1xuXHQwJSB7XG5cdFx0b3BhY2l0eTogMDtcblx0fVxuXG4gICAgZnJvbSB7XG5cdFx0b3BhY2l0eTogMDtcbiAgICB9XG5cblx0MTAwJSB7XG5cdFx0b3BhY2l0eTogMTtcblx0fVxufSIsImh0bWwiOiI8IURPQ1RZUEUgaHRtbD5cbjwhLS1cblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cblxuVGhlIHRhYnMgYXJlOlxuXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cbi0tPlxuXG48aHRtbCBsYW5nPVwiZW5cIj5cbiAgICA8aGVhZD5cbiAgICAgICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cbiAgICA8L2hlYWQ-XG4gICAgPGJvZHk-XG4gICAgICAgIDxwPlRleHQ8L3A-XG4gICAgPC9ib2R5PlxuPC9odG1sPiJ9LCJsYW5ndWFnZSI6Im1hcmtkb3duIiwianNPcHRpb25zIjp7InBhcnNlciI6ImVzcHJlZSIsInNvdXJjZVR5cGUiOiJtb2R1bGUiLCJpc1ZlcnNpb24iOiJsYXRlc3QiLCJpc0pTWCI6ZmFsc2UsImVzVmVyc2lvbiI6ImxhdGVzdCJ9LCJqc29uT3B0aW9ucyI6eyJqc29uTW9kZSI6Impzb241IiwiYWxsb3dUcmFpbGluZ0NvbW1hcyI6ZmFsc2V9LCJjc3NPcHRpb25zIjp7ImNzc01vZGUiOiJjc3MiLCJ0b2xlcmFudCI6ZmFsc2V9LCJtYXJrZG93bk9wdGlvbnMiOnsibWFya2Rvd25Nb2RlIjoiY29tbW9ubWFyayIsIm1hcmtkb3duRnJvbnRtYXR0ZXIiOiJvZmYiLCJtYXJrZG93bk1hdGgiOmZhbHNlfSwiaHRtbE9wdGlvbnMiOnsidGVtcGxhdGVFbmdpbmVTeW50YXgiOiJub25lIiwiZnJvbnRtYXR0ZXIiOiJvZmYifSwid3JhcCI6dHJ1ZSwidmlld01vZGVzIjp7ImFzdFZpZXciOiJ0cmVlIiwic2NvcGVWaWV3IjoiZmxhdCIsInBhdGhWaWV3IjoiZ3JhcGgifSwicGF0aEluZGV4Ijp7ImluZGV4IjowLCJpbmRleGVzIjoxfSwiZXNxdWVyeVNlbGVjdG9yIjp7InNlbGVjdG9yIjoiIn19LCJ2ZXJzaW9uIjowfQ)

```md
Install the package first.
###### Installation
```

tests/rules/no-heading-like-paragraph.test.js
```diff
@@ -0,0 +1,600 @@
...
+		"#######*Installation*",
+
+		// Escapes and character references
+		"\\####### Installation",
+		"&#35;###### Installation",
+
+		// Code
+		"```md\n####### Installation\n```",
+		"    ####### Installation",
+		"`####### Installation`",
```
lumirlumir, line 54:
```suggestion

        // InlineCode
		"`####### Installation`",
```

Non-blocking request: to avoid confusion, it might help to clarify the node type, since a single backtick is used for `InlineCode`, not `Code`.

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,112 @@
...
+// Helpers
+//-----------------------------------------------------------------------------
+
+/**
+ * Matches seven or more hash characters at the start of a paragraph, followed by a
+ * space, a tab, a line ending, or the end of the paragraph. This mirrors the way
+ * CommonMark delimits the opening sequence of an ATX heading, so a no-break space
+ * doesn't count as a delimiter.
+ */
+const headingLikeParagraphPattern = /^#{7,}(?=[ \t\r\n]|$)/u;
```
lumirlumir: I think it would be nice to add tests related to lone CR line endings, since there are currently no test cases that verify them.

lumirlumir (changes requested):

docs/rules/no-heading-like-paragraph.md
Gaic4o: I think the direction you suggested aligns well with the intent of this rule. In particular, it seems more consistent to handle the actual Markdown line break behavior and headings inside blockquotes in the same way, so I updated the implementation accordingly and also added regression tests for LS/PS.

tests/rules/no-heading-like-paragraph.test.js
Gaic4o: I agree. In this case, I think it’s more consistent to follow the CommonMark specification and `remark-lint-no-heading-like-paragraph` rather than introducing an exception, so I kept the bare `#######` case as reportable.

docs/rules/no-heading-like-paragraph.md
Gaic4o: Sounds good to me. I added a `## Options` section to keep the documentation structure consistent.

docs/rules/no-heading-like-paragraph.md
Gaic4o: I understand what you mean. I’ve updated the explanation accordingly.

tests/rules/no-heading-like-paragraph.test.js
Gaic4o: That makes sense, so I updated the comment to `InlineCode`. Thank you for reviewing the tests so carefully and for leaving such detailed feedback.

src/rules/no-heading-like-paragraph.js
Gaic4o: I added a regression test for lone CR line endings as well. Thank you!

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,126 @@
...
+ * This pattern avoids the `m` flag, which would also treat U+2028 and U+2029 as line
+ * boundaries even though Markdown doesn't. `(?:^|(?<=[\r\n]))` starts a new line only
+ * after an actual carriage return or line feed.
+ *
+ * Block quote markers and up to three spaces of indentation may precede the hash
+ * characters, because a heading with six or fewer hash characters would still open in
+ * that position.
+ */
+const headingLikeParagraphPattern =
+	/(?:^|(?<=[\r\n]))(?: {0,3}>[ \t]?)* {0,3}(?<hashes>#{7,})(?=[ \t\r\n]|$)/gu;
```
lumirlumir, line 36: It seems that the overlapping whitespace matches (`[ \t]?` and the following ` {0,3}`) cause exponential backtracking for nested blockquotes without hashes.

A small input with 25 levels took approximately 1.2 seconds to lint. It would be nice to make the prefix matching unambiguous and add this as a regression case to `valid`.

```js
`${"> ".repeat(30)}foo\n${"> ".repeat(30)}bar`,
```

lumirlumir, line 36: The ` {0,3}` limit includes container indentation, so this case is not reported:

```markdown
10. Intro
    ####### Heading
```

Those four spaces belong to the list item. Replacing the seven hashes with six produces a valid heading. It'd be helpful to account for container indentation and add regression tests for continuation lines in ordered lists, nested lists, and GFM footnotes.

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,126 @@
...
+		return {
+			paragraph(node) {
+				/*
+				 * Read the raw source text instead of the `value` of the first `text`
+				 * child, because `value` already resolves character escapes and character
+				 * references. Both `\####### Foo` and `&#35;###### Foo` render as a
+				 * paragraph whose text starts with seven hash characters, but in each case
+				 * the author escaped the leading hash on purpose.
+				 */
+				const text = sourceCode.getText(node);
```
lumirlumir, line 77: Scanning the entire paragraph also reports hashes inside multiline inline code and link-title:

```md
`example
####### text
`
```

```md
[link](https://example.com "
####### title
")
```

Also, both suggestions change its content: reducing the hashes breaks the code span, while escaping adds a literal backslash. 

Can we exclude inline code and link-title ranges from matching, and add regression tests for both cases?

---

The pattern used in the `no-reversed-media-syntax` rule, which masks the original source text based on the node range, would be a helpful solution for this case:

https://github.com/eslint/markdown/blob/186ad3ee58982d006e63d3bbe046d69d8ee36e33/src/rules/no-reversed-media-syntax.js#L64-L83

lumirlumir (changes requested):

src/rules/no-heading-like-paragraph.js
```diff
@@ -0,0 +1,126 @@
...
+//-----------------------------------------------------------------------------
+
+export default /** @satisfies {NoHeadingLikeParagraphRuleDefinition} */ ({
+	meta: {
+		type: "problem",
+
+		docs: {
+			description: "Disallow paragraphs that look like ATX headings",
+			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-heading-like-paragraph.md",
+		},
```
lumirlumir, line 52:
```suggestion
		type: "problem",
		languages: ["markdown/commonmark", "markdown/gfm"],

		docs: {
			description: "Disallow paragraphs that look like ATX headings",
			dialects: ["CommonMark", "GFM"],
			url: "https://github.com/eslint/markdown/blob/main/docs/rules/no-heading-like-paragraph.md",
		},
```

One more review comment I missed: PR #664 has been merged, so the properties above will be needed.

lumirlumir (changes requested):

src/rules/no-heading-like-paragraph.js
Gaic4o: Added the languages and dialects metadata as suggested. Thanks for catching this!

src/rules/no-heading-like-paragraph.js
Gaic4o: Thanks for the suggestion. I updated the rule to exclude inline code and title ranges from matching while preserving the original source offsets, and added regression tests for these cases.

src/rules/no-heading-like-paragraph.js
Gaic4o: Thanks for pointing this out. I updated the rule to account for container indentation and added regression tests for ordered lists, nested lists, and GFM footnotes.

Gaic4o: Thanks for catching this. I updated the prefix handling to avoid the ambiguous whitespace matching and added the nested blockquote case as a regression test.

PR #719, opened 2026-08-18
fix: support unquoted aria-hidden value in require-alt-text

### LINKED ISSUE #718: Bug: require-alt-text does not recognize an unquoted aria-hidden=true attribute
### Environment

ESLint version: 10.8.0
@eslint/markdown version: 8.0.3
Node version: v22.22.3
npm version: 10.8.2

commonmark

### What did you do?

<details> <summary>Configuration</summary>

```js
import { defineConfig } from "eslint/config";

export default defineConfig([
    {
        files: ["**/*.md"],
        plugins: {
            markdown,
        },
        rules: {
            "markdown/require-alt-text": "error",
        },
    },
]);
```
</details>

```md
<!-- OK: aria-hidden has a quoted value -->
<img src="quoted.png" aria-hidden="true">

<!-- Expected OK: an unquoted HTML attribute value is valid -->
<img src="unquoted.png" aria-hidden=true>
```

Neither `` element should produce an error.

HTML allows attribute values to be specified without quotes, so `aria-hidden=true` is valid HTML. 
Therefore, require-alt-text should treat the following forms equivalently:

```html
<img src="quoted.png" aria-hidden="true">
<img src="unquoted.png" aria-hidden=true>
```
Because both images have an aria-hidden value of true, the alternative text check should be skipped, consistent with the existing behavior of require-alt-text.

### What actually happened?

The quoted aria-hidden="true" value is recognized correctly, but the unquoted aria-hidden=true value is not.

As a result, the second  element produces the following error:

```md
5:1  error  Alternative text for image is required  markdown/require-alt-text
```
The regular expression currently used by require-alt-text to inspect HTML attributes only captures values enclosed in single or double quotes.

As a result, the two forms are handled differently:

```md
aria-hidden="true" → "true" is captured
aria-hidden=true   → the attribute value is not captured
```

https://stackblitz.com/edit/vitejs-vite-o7wsgrqu?file=eslint.config.js,test.md

### Participation

### AI acknowledgment

- [ ] I did not use AI to generate this issue report.

### Additional comments

> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

Related issue: [#481](https://github.com/eslint/markdown/issues/481)

Issue #481 improved `no-missing-link-fragments` to recognize unquoted HTML `id` and `name` attribute values.

```html
<h1 id=bookmark>Bookmark</h1>
```

Because the HTML specification defines unquoted attribute values as valid syntax, `require-alt-text` should treat `aria-hidden=true` the same as `aria-hidden="true"`.

```html
<img src="unquoted.png" aria-hidden=true>
```

Related HTML specification:
[https://html.spec.whatwg.org/multipage/syntax.html#syntax-attributes](https://html.spec.whatwg.org/multipage/syntax.html#syntax-attributes)

KumJungMin:
> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

#### Prerequisites checklist

#### AI acknowledgment

This PR updates require-alt-text to recognize an unquoted aria-hidden=true attribute.

HTML permits attribute values to be written without quotes, but the existing implementation handles the following forms differently:
```html
<img src="quoted.png" aria-hidden="true" />
<img src="unquoted.png" aria-hidden=true />
```
Since both forms set the value of aria-hidden to true, this PR updates the rule to skip the alternative text check in both cases.

- Added a dedicated regular expression that recognizes double-quoted, single-quoted, and unquoted `aria-hidden=true` attributes.
- Added support for optional whitespace around the equals sign.
- Made matching case-insensitive for both the attribute name and value.
- Added a boundary check to ensure that the alternative text check is skipped only when the value is exactly true.
- Added valid test cases for the following unquoted forms:
   - `aria-hidden=true`
   - `aria-hidden = true`
   - `ARIA-HIDDEN=TRUE`
- Added invalid test cases to verify that `aria-hidden=false` and `aria-hidden=truefoo` are still reported.

The regex test cases for this change are available on [RegExr](https://regexr.com/8o80k).

#### Related Issues

fixes #718

To keep the scope of this change minimal, the implementation adds a dedicated regular expression for detecting `aria-hidden=true` instead of extending `getHtmlAttributeRe()`.

There are two reasons for this approach:

1. It avoids affecting the existing `alt` attribute parsing behavior and avoids recreating the `aria-hidden` regular expression each time an image tag is processed.
2. `aria-hidden` requires checking whether its value is exactly `true`, while `getHtmlAttributeRe("alt")` captures the value of the `alt` attribute to determine whether it consists only of whitespace. I considered these attributes to have different validation purposes and criteria.

I would appreciate feedback on whether this dedicated regular expression is appropriate or whether it would be better to generalize `getHtmlAttributeRe()` to handle both quoted and unquoted attribute values. :)

### DIFF
--- a/src/rules/require-alt-text.js
+++ b/src/rules/require-alt-text.js
@@ -26,6 +26,8 @@ import { stripHtmlComments } from "../util.js";
 //-----------------------------------------------------------------------------
 
 const imgTagPattern = /<img(?:\s(?:[^>"']|"[^"]*"|'[^']*')*)?\/?>/giu;
+const ariaHiddenTruePattern =
+	/\saria-hidden\s*=\s*(?:"true"|'true'|true)(?=\s|\/?>)/iu;
 
 /**
  * Creates a regex to match HTML attributes
@@ -78,14 +80,8 @@ export default /** @satisfies {RequireAltTextRuleDefinition} */ ({
 
 				while ((match = imgTagPattern.exec(text)) !== null) {
 					const imgTag = match[0];
-					const ariaHiddenMatch = imgTag.match(
-						getHtmlAttributeRe("aria-hidden"),
-					);
-					if (
-						ariaHiddenMatch &&
-						ariaHiddenMatch[1] &&
-						ariaHiddenMatch[1].toLowerCase() === "true"
-					) {
+
+					if (ariaHiddenTruePattern.test(imgTag)) {
 						continue;
 					}
 
--- a/src/rules/require-alt-text.js
+++ b/src/rules/require-alt-text.js
@@ -27,7 +27,7 @@ import { stripHtmlComments } from "../util.js";
 
 const imgTagPattern = /<img(?:\s(?:[^>"']|"[^"]*"|'[^']*')*)?\/?>/giu;
 const ariaHiddenTruePattern =
-	/\saria-hidden\s*=\s*(?:"true"|'true'|true)(?=\s|\/?>)/iu;
+	/[\s"']aria-hidden\s*=\s*(?:"true"|'true'|true(?=\s|\/?>))/iu;
 
 /**
  * Creates a regex to match HTML attributes
--- a/tests/rules/require-alt-text.test.js
+++ b/tests/rules/require-alt-text.test.js
@@ -68,6 +68,8 @@ ruleTester.run("require-alt-text", rule, {
 		'<img src="image.png" aria-hidden=true/>',
 		'<img src="image.png" aria-hidden = true/>',
 		'<img src="image.png" ARIA-HIDDEN=TRUE />',
+		'<img before="true"aria-hidden="true"after="true">',
+		"<img before='true'aria-hidden='true'after='true'>",
 
 		'<p><img src="image.png" alt="Descriptive text" /></p>',
 		'<!-- <img src="image.png" /> -->',
--- a/tests/rules/require-alt-text.test.js
+++ b/tests/rules/require-alt-text.test.js
@@ -402,14 +402,14 @@ ruleTester.run("require-alt-text", rule, {
 			],
 		},
 		{
-			code: '<img src="image.png" aria-hidden=falsefoo/>',
+			code: '<img src="image.png" aria-hidden=truefoo/>',
 			errors: [
 				{
 					messageId: "altTextRequired",
 					line: 1,
 					column: 1,
 					endLine: 1,
-					endColumn: 44,
+					endColumn: 43,
 				},
 			],
 		},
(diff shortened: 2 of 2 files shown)


src/rules/require-alt-text.js
```diff
@@ -26,6 +26,8 @@ import { stripHtmlComments } from "../util.js";
 //-----------------------------------------------------------------------------
 
 const imgTagPattern = /<img(?:\s(?:[^>"']|"[^"]*"|'[^']*')*)?\/?>/giu;
+const ariaHiddenTruePattern =
+	/\saria-hidden\s*=\s*(?:"true"|'true'|true)(?=\s|\/?>)/iu;
```
DMartens: There does not have to be whitespace around an HTML attribute, e.g. `<a before="true"aria-hidden="true"after="true"></a>` is valid.
In other words the before "aria-hidden" can be a quote or whitespace and alpha characters if the attribute value is quoted.

KumJungMin: I’d like to update it as follows to account for the additional cases.

Case | Before | After | Reason
-- | -- | -- | --
Boundary before aria-hidden | \s | (?:\s\|["']) | Allows aria-hidden to be preceded not only by whitespace, but also directly by the closing quote (" or ') of the previous attribute.
Boundary after true | (?:"true"\|'true'\|true)(?=\s\|\/?>) | (?:"true"\|'true'\|true(?=\s\|\/?>)) | Allows another attribute to immediately follow "true" or 'true' without whitespace, while keeping the existing boundary check for unquoted true so that cases like aria-hidden=truefoo are not matched incorrectly.

I’m thinking of updating the final pattern as follows. Is there any case I may have misunderstood? (https://regexr.com/8o87r)
```js
/(?:\s|["'])aria-hidden\s*=\s*(?:"true"|'true'|true(?=\s|\/?&gt;))/iu
```

DMartens: Yes the new pattern works for the mentioned cases.
Can you please add test cases for this and replace `&gt;` with `>` (no need to escape it in this case) in the pattern?

KumJungMin: I updated the regular expression to address the `regexp/prefer-character-class` ESLint error.

스크린샷 2026-08-20 오후 7 31 10




The previous pattern used a disjunction of single-character alternatives:

```js
(?:\s|["'])
```

Since this matches either whitespace, `"` or `'`, it can be expressed more directly as a character class without changing its meaning:

```js
[\s"']
```

So I updated the pattern as follows:

```js
/[\s"']aria-hidden\s*=\s*(?:"true"|'true'|true(?=\s|\/?>))/iu
```

Here, `[\s"']` allows `aria-hidden` to be preceded by whitespace or by the closing quote of the previous attribute. For the unquoted `true` form, the lookahead `(?=\s|\/?>)` is kept to ensure that values such as `aria-hidden=truefoo` are not matched.

I also added test cases for attributes without whitespace between them and confirmed that the previously discussed cases still work as expected. (https://regexr.com/8o8jo)

KumJungMin: The updated commit is https://github.com/eslint/markdown/pull/719/changes/b403e09b2778ba84bc631b7b3056813101fc06c0

DMartens (approved):
Changes LGTM, thanks.

PR #722, opened 2026-08-27
fix: handle `symbol` in `validateLanguageOptions`

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

CommonMark and GFM.

### What did you do?

When I use `MarkdownLanguage#validateLanguageOptions` with a `symbol` value, it produces an incorrect error message.

I expected the error message to be:

```sh
Invalid language option value `Symbol(frontmatter)` for frontmatter. Expected one of `false`, `"yaml"`, `"toml"`, or `"json"`.
```

### What actually happened?

However, I got:

```sh
'Cannot convert a Symbol value to a string'
```

Create a `test.js` file in the root of the `markdown` project.

```js
import { MarkdownLanguage } from "./src/language/markdown-language.js";

const language = new MarkdownLanguage();

for (const languageOptions of [
	{ frontmatter: Symbol("frontmatter") },
	{ math: Symbol("math") },
]) {
	try {
		language.validateLanguageOptions(languageOptions);
	} catch (error) {
		console.error(error.message);
	}
}
```

This is a simple bug fix. I’ve coerced the message variable with `String` so that it can also handle an incorrectly passed `symbol` value.

This is a subtle edge case, but I noticed it while working on #706.

## Related Issues

N/A

I’ve verified that the JSON and CSS language plugins don’t have this type of issue.

### DIFF
--- a/src/language/markdown-language.js
+++ b/src/language/markdown-language.js
@@ -185,7 +185,7 @@ export class MarkdownLanguage {
 			!validFrontmatterOptions.has(frontmatterOption)
 		) {
 			throw new Error(
-				`Invalid language option value \`${frontmatterOption}\` for frontmatter. Expected one of \`false\`, \`"yaml"\`, \`"toml"\`, or \`"json"\`.`,
+				`Invalid language option value \`${String(frontmatterOption)}\` for frontmatter. Expected one of \`false\`, \`"yaml"\`, \`"toml"\`, or \`"json"\`.`,
 			);
 		}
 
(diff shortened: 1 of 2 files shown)


DMartens:
The JS, CSS and JSON language do not have the problem as they do not print the value.
I think the explicit `String` cast only makes a difference for `Symbol` and as such I think the change is fine.
Leaving open for a potential second review.

DMartens (approved):

fasttime (approved):
LGTM, thanks!

PR #736, opened 2026-09-13
fix: prevent ReDoS in `no-reference-like-urls`

### LINKED ISSUE #734: Bug: `no-reference-like-urls` regex has catastrophic backtracking (ReDoS)
### Environment

ESLint version: v10.8.1
@eslint/markdown version: 8.0.3
Node version: v24.14.1
npm version: v11.11.0
Operating System: darwin 25.6.0 (macOS 26.6.2)

commonmark (also reproduces under gfm)

### What did you do?

`no-reference-like-urls`'s regex has the exact same shape of vulnerability that `no-reversed-media-syntax` had, fixed in #693. This rule just never got the same fix. Both rules are `recommended`, so any project using the recommended config has this on by default.

```js
// eslint.config.mjs

export default [
  {
    files: ["**/*.md"],
    plugins: { markdown },
    rules: { "markdown/no-reference-like-urls": "error" },
  },
];
```

Generate a single-line fixture and lint it:

```sh
node -e 'const l = "[", r = "]"; require("fs").writeFileSync("repro.md", l + "()".repeat(28) + r + "(http://example.com \"a\\\"b\")\n")'
npx eslint repro.md
```

This is a complete, well-formed inline link (square-bracket label, then a parenthesized URL and title) — not a malformed one. The hang happens while the rule is still trying to parse the label, before it even gets to comparing the URL against known reference identifiers.

ESLint finishes near-instantly and reports nothing (there's no reference definition anywhere in the file for the rule to match against, so linting this file should be a no-op for this rule regardless of file size).

### What actually happened?

Wall-clock time for `npx eslint repro.md`, varying only the repeat count `n` in `'()'.repeat(n)`:

| n | eslint runtime |
|---|---|
| 18 | 0.74s |
| 20 | 0.92s |
| 22 | 2.64s |
| 24 | 9.80s |
| 26 | 37.57s |
| 28 | 149.43s |
| 30 | 219.33s |

(Measured with `time npx eslint repro.md`. Absolute times will vary by machine, but the exponential growth pattern itself shouldn't be hardware-specific.)

Going from n=18 to n=30 (just 12 more repeats) pushes the runtime from under a second to 219s — exponential, not linear — and it pins one CPU core near 100% the whole time. `--debug` gives no indication of which rule is stuck.

Calling the rule's own regex directly (no ESLint involved) shows the same curve, confirming the regex itself is the bottleneck and not something in the rule's surrounding logic:

```js
const linkOrImagePattern =
  /\[(?<label>(?:\\.|[^()\\]|\([\s\S]*\))*?)\]\((?<destination>[ \t]*\r?\n?(?<![ \t])[ \t]*(?:<[^>]*>|[^ \t()]+))(?:[ \t]*\r?\n?(?<![ \t])[ \t]*(?:"[^"]*"|'[^']*'|\([^)]*\)))?[ \t]*\r?\n?(?<![ \t])[ \t]*\)$/u;

const l = "[", r = "]";
const testInput = l + "()".repeat(24) + r + "(http://example.com \"a\\\"b\")";
linkOrImagePattern.exec(testInput);
// ~9s just for this one exec() call
```

N/A — the snippet above is a complete, self-contained reproduction.

### Participation

### AI acknowledgment

- [ ] I did not use AI to generate this issue report.

### Additional comments

**Suggested fix.** Apply the same fix used in #693 to `linkOrImagePattern` — bound `\([\s\S]*\)` to `\([^()]*\)`.

```diff
- /\[(?<label>(?:\\.|[^()\\]|\([\s\S]*\))*?)\]\(.../u;
+ /\[(?<label>(?:\\.|[^()\\]|\([^()]*\))*?)\]\(.../u;
```

I ran this rule's existing test suite against the fix and all 82 tests pass.

There's a side effect here that's different in kind from #693's trade-off, though. For `no-reversed-media-syntax`, paren structure itself is what the rule is judging, so limiting nesting depth was a meaningful semantic restriction. This rule only cares whether the URL matches a reference definition, so nested parens in the label shouldn't matter to it at all — yet with this fix, a label with 2+ levels of nesting fails to match the regex entirely, so the destination never gets extracted and the check is silently skipped even when the URL genuinely matches a definition. For example, a link whose label is `see (x (a (b)) y)` and whose destination is `https://example.com/doc` extracts `destination` fine under the current regex, but fails to match at all under the bounded one.

> Disclosure: I'm a participant in [open source contribution program OSSCA](https://github.com/eslint-ossca).

sohxxny:
#### Prerequisites checklist

#### AI acknowledgment

Fix the catastrophic backtracking (ReDoS) in `no-reference-like-urls` reported in #734. The rule's `linkOrImagePattern` regex could take exponential time to parse a well-formed inline link, so any project using the recommended config was exposed to this by default.

Bounded the nested group in `linkOrImagePattern` from `\([\s\S]*\)` to `\([^()]*\)` — the same fix applied to `no-reversed-media-syntax` in #693. The unbounded group let a label be parsed two different ways (as a bare non-paren run, or as a nested-paren span). On a failed overall match, the engine tried every way of splitting the input between these two overlapping alternatives.

Timings for `time npx eslint repro.md` (`n` = repeats of `()` in the label):

| n | before (#734) | after |
| --- | --- | --- |
| 28 | ~149s | ~1s |
| 30 | ~219s | ~1s |

("after" is dominated entirely by Node/ESLint startup overhead — the regex itself now resolves in well under 1ms regardless of `n`.)

Added a `valid` case reproducing the issue as a regression test, plus a second `valid` case documenting the new nesting-depth limit this fix introduces.

#### Related Issues

fixes #734

This narrows what a label can contain, which is a behavior change: a label with two or more levels of nested parentheses (e.g. `[a (b (c)) d](x)`) is no longer matched, so this rule silently skips its check on such links even when the destination matches a definition. All 82 existing tests pass unchanged (84 total after the 2 added here).

> Disclosure: I'm a participant in [open source contribution program OSSCA](https://github.com/eslint-ossca).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved processing of reference-style links and images with nested parentheses in labels.
  - Prevented excessive processing for complex link patterns.
  - Correctly handles labels followed by reference definitions.

- **Documentation**
  - Documented the limitation that label parsing supports at most one nested parenthesis level.

- **Tests**
  - Added coverage for deeply nested parentheses and related reference-style links.


DMartens:
I think we should not use a regular expression at all.
In the original [PR](https://github.com/eslint/markdown/pull/433) this was introduced because angle brackets are [omitted in the AST representation](https://github.com/eslint/markdown/pull/433#discussion_r2174911199).
Should we rather refactor this code rather than updating the regular expression?
The regular expression is used for two things:
- extracting the label for the fixer -> only change `[` and `]` to `(` and `)` rather than updating the whole link node
- use `sourceCode.getText()` to check for angle brackets

PR #745, opened 2026-09-20
fix: ignore attribute values when detecting alt in require-alt-text

### LINKED ISSUE #741: Bug: `require-alt-text` does not report `<img>` when "alt" appears inside another attribute's value
### Environment

ESLint version: v10.10.0
@eslint/markdown version: v8.0.3
Node version: v24.21.0
npm version: v11.19.0
Operating System: win32 10.0.26200

commonmark

### What did you do?

<details>

```js
import { defineConfig } from "eslint/config";

export default defineConfig([
   {
      files: ["**/*.md"],
      plugins: { markdown },
      extends: ["markdown/recommended"],
   },
]);
```
</details>

```md
# Title

<img src="a.png">

<img src="b.png" title="an alt">

<img src="c.png" class="no alt">

<img src="d.png" data-caption="Photo alt">

<img src="e.png" alt="Logo">
```

I ran `npx eslint test.md`.

`markdown/require-alt-text` should report four images: `a.png`, `b.png`, `c.png`, and `d.png`.

None of `b.png`, `c.png`, and `d.png` has an `alt` attribute. The word "alt" only appears inside the value of another attribute (`title`, `class`, `data-caption`).

### What actually happened?

Only `a.png` is reported:

```
test.md
  3:1  error  Alternative text for image is required  markdown/require-alt-text

```

`b.png`, `c.png`, and `d.png` are not reported, even though they have no alternative text.

https://stackblitz.com/edit/stackblitz-starters-iqmooju2

### Participation

### AI acknowledgment

### Additional comments

> Disclosure: I'm a participant of the open source contribution program [OSSCA](https://github.com/eslint-ossca).

minseonkkim:
#### Prerequisites checklist

#### AI acknowledgment

This PR fixes false negatives in `require-alt-text` when the text "alt" appears inside another attribute's value.

```markdown
<img src="b.png" title="an alt">
<img src="c.png" class="no alt">
<img src="d.png" data-caption="Photo alt">
```

None of these images has an `alt` attribute, but none of them was reported. `getHtmlAttributeRe("alt")` searched the whole tag string for `\salt`, so it also matched text inside a quoted value and the rule treated the image as having a valueless `alt` attribute.

- Replaced `getHtmlAttributeRe(name)` with `getHtmlAttribute(tag, name)`.
  - Instead of searching the raw tag string, it reads the tag's attributes one by one: a name followed by an optional double-quoted, single-quoted, or unquoted value.
  - It returns the value of the attribute whose name matches exactly.
  - A quoted value is consumed as part of its own attribute, so text inside it is never treated as an attribute name.
- Simplified the `alt` check to use the returned value. The reporting conditions are unchanged:
  - a missing `alt` is reported,
  - a whitespace-only `alt` is reported,
  - `alt` and `alt=""` are still allowed.
- Added invalid test cases:
  - the three cases from the issue, plus a single-quoted variant,
  - `title=" alt='text' "`, where the value contains something that looks like a complete `alt` attribute,
  - `title="an alt" alt=" "`, where the earlier match hid a real whitespace-only `alt`,
  - `alternate="text"` and `alt-text="text"` (see the last section).
- Added valid test cases to make sure a real `alt` is still found next to such attributes, including `data-alt="text" alt="..."`.

#### Related Issues

Fixes #741

1. **`alternate` / `alt-text`.**
   - These are slightly outside the issue title: "alt" is in the attribute *name*, not in a value.
   - The old pattern had no boundary after the name, so it matched them as well. Comparing the full name fixes both at once, so I included the tests here.
2. **`aria-hidden` is out of scope.**
   - `ariaHiddenTruePattern` has the same kind of problem. For example, `` is not reported.
   - I did not change it in this PR. I plan to open a separate issue for it, where the new helper could be reused.

> Disclosure: I'm a participant of the open source contribution program [OSSCA](https://github.com/eslint-ossca).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of image `alt` attributes across different quoting styles.
  * Prevented false positives when “alt” appears in other attribute names or values.
  * Continued reporting images with missing or whitespace-only `alt` text.

* **Tests**
  * Added coverage for valid and invalid image markup involving similar attribute names and values.


kavixo:
This PR is 100% AI-generated. OP's entire account uses Claude for basically every commit. Just look at her other repos: https://github.com/minseonkkim/aqua-world, https://github.com/minseonkkim/sweetbook, etc.

Unfortunately, this isn't an isolated case. This has been an ongoing thing with OSSCA members. You can look through their accounts and see extensive use of Claude throughout their development work. Yet when they open issues/PRs in the ESLint org, they consistently mark them as "I did not use AI to generate this PR."

@lumirlumir, you're an ESLint maintainer and the mentor of these people, so you're clearly aware of this behavior. Your behavior around this has been sketchy, and it's creating serious trust issues for people who rely on ESLint. This needs to be addressed.

If this continues to be ignored, I’m going to escalate this to GitHub staff and ask them to investigate your conduct as an ESLint maintainer, the OSSCA organization, and the private repository in the OSSCA org.

lumirlumir:
@kavixo Thank you for the ping. Could you please name a few PRs or issues where you’ve observed this kind of activity so I can take action? I’ve emphasized the importance of disclosing it honestly, but ultimately, it depends on their honesty and morality. If you could send me the details of any such activity you’ve observed via Discord DM, email, or another channel, I’d appreciate it.

src/rules/require-alt-text.js
```diff
@@ -29,13 +29,26 @@ const imgTagPattern = /<img(?:\s(?:[^>"']|"[^"]*"|'[^']*')*)?\/?>/giu;
 const ariaHiddenTruePattern =
 	/[\s"']aria-hidden\s*=\s*(?:"true"|'true'|true(?=\s|\/?>))/iu;
 
+const attributePattern =
```
DMartens: This regular expression is overengineered:
- `name`: only should match for "alt", why not hardcode this?
- there is no need for three capture groups for double/single/unquoted attributes which also complicates the code below extracting the value

tests/rules/require-alt-text.test.js
```diff
@@ -71,6 +71,11 @@ ruleTester.run("require-alt-text", rule, {
 		'<img before="true"aria-hidden="true"after="true">',
 		"<img before='true'aria-hidden='true'after='true'>",
 
+		// "alt" inside another attribute's name or value
+		'<img src="image.png" title="an alt" alt="Descriptive text">',
+		'<img src="image.png" alt="Descriptive text" title="an alt">',
+		'<img src="image.png" data-alt="text" alt="Descriptive text">',
```
lumirlumir, line 77: Could we also add the following three test cases that were mentioned in the issue?

```md
<img src="b.png" title="an alt">

<img src="c.png" class="no alt">

<img src="d.png" data-caption="Photo alt">
```

src/rules/require-alt-text.js
minseonkkim: Thanks for the review! I've replaced the generic attribute tokenizer with a single regex that matches only the `alt` attribute, and it now uses one capture group for the value as before.

tests/rules/require-alt-text.test.js
minseonkkim: Added the three cases from the issue.

